PR #2794, opened 2021-06-24
Bring back MacOS agent

lrascao:
Some comments seem to point out that it has been working at some point in time.

### DIFF
--- a/pkg/cmd/agent/nodeinfo.go
+++ b/pkg/cmd/agent/nodeinfo.go
@@ -8,7 +8,6 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-	"runtime"
 	"strconv"
 
 	"github.com/go-chi/render"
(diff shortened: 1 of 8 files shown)


mmatczuk:
Cheers for the contribution.

My 2c.
Scylla Agent on Mac makes little sense since Scylla will never work on Mac.
Agent is only run in a docker container in this setup maybe try explicitly setting target linux in makefile where it build the dev binary.

Can you clarify what kind of issue your are trying to solve by this patch.

lrascao:
Sure, this is mostly around exploration of the code base (i'm on MacOS) and seeing how the parts fit together. From a developer standpoint, in my view at least, the less barriers to actually trying out a system the better, i figured a workflow scenario of manager/agent running locally and connecting to a containerized scylla could be made possible, this reduces a bit of dev friction (building/starting docker image takes a bit).
Of course there might be additional barriers that i'm not aware of, in that case feel free to close.

mmatczuk:
I decided to add your patches.
Thanks for the contribution.

PR #2810, opened 2021-06-29
dist: remove legacy and switch to using goreleaser

yaronkaikov:
Following the work on scylladb/scylla-manager-enterprise#2671, we can now use only goreleaser to build scylla-manager


yaronkaikov:
@mmatczuk so i did some changes on top of https://github.com/scylladb/scylla-manager/commit/02edef1e2e29e43adee7ad44eb602c4e64abfedc 

pipeline creation for manager is being done as part of https://github.com/scylladb/scylla-pkg/pull/2151

please see the testing pipeline and let me know if you have any comments about the general idea or anything else 
https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/releng-testing/job/manager-build/

Still working on it

yaronkaikov:
@mmatczuk Scylla-pkg side is ready and about to me merged. Can you please review this PR?

yaronkaikov:
@mmatczuk ping

yaronkaikov:
@mmatczuk @penberg please review

dist/.goreleaser.yaml
```diff
@@ -64,6 +64,7 @@ nfpms:
     builds:
       - server
     package_name: "scylla-manager-server"
+    file_name_template: "{{ .ProjectName }}-server-{{ .Version }}_{{ .Os }}.{{ .Arch }}"
```
penberg, line 67: What is the problem these file name template changes are fixing? It would be nicer to consolidate them in the rpm packaging `overrides` section.

dist/.goreleaser.yaml
```diff
@@ -97,12 +98,15 @@ nfpms:
...
       deb:
+        file_name_template: "{{ .ProjectName }}-server_{{ .Version }}.{{ .Os }}_{{ .Arch }}"
         scripts:
           preinstall: deb/scylla-manager-server.preinst
           postinstall: deb/scylla-manager-server.postinst
           preremove: deb/scylla-manager-server.prerm
           postremove: deb/scylla-manager-server.postrm
       rpm:
+        replacements:
+          amd64: x86_64
```
penberg, line 109: Does `goreleaser` actually generate `.rpm` files with `amd64` in the name? That's sad.

We probably need something for `aarch64` too?

dist/.goreleaser.yaml
```diff
@@ -196,7 +208,6 @@ dockers:
       - client
     image_templates:
       - "scylladb/scylla-manager:{{ .Version }}"
-      - "scylladb/scylla-manager:latest"
```
penberg, line 199: Why is this removed?

dist/.goreleaser.yaml
```diff
@@ -205,7 +216,6 @@ dockers:
       - agent
     image_templates:
       - "scylladb/scylla-manager-agent:{{ .Version }}"
-      - "scylladb/scylla-manager-agent:latest"
```
penberg, line 208: Why is this removed?

dist/.goreleaser.yaml
yaronkaikov: Since we are aiming to use same code in scylla-pkg. i changed the template to match what we have in `Scylla`

dist/.goreleaser.yaml
yaronkaikov: The default is to generate the rpms with amd64. again we need it to be identical to what we have in Scylla

I guess we need to address ARM as well, but i think it could be in another PR , what do you think?

dist/.goreleaser.yaml
yaronkaikov: Latest tag is related only to latest version which we deal during promotion . it's not needed here

dist/.goreleaser.yaml
yaronkaikov: Latest tag is related only to latest version which we deal during promotion . it's not needed here

dist/.goreleaser.yaml
penberg: Ok, but why does this _template_ have to be defined for very package? Can't you add it to the `override` section once for all packages?

dist/.goreleaser.yaml
penberg: Sure, makes sense. Please update the commit description to explain this.

dist/.goreleaser.yaml
penberg: Yeah, better to defer, since we don't even have `arm64` in the `goarch` section.

yaronkaikov: well it's supported based on https://stackoverflow.com/questions/20728767/all-possible-goos-value

dist/.goreleaser.yaml
yaronkaikov: There is a different between rpm and deb packages. this is working (since it's identical to what we have in Scylla):
`scylla-manager-agent_2.5-dev-0.20210722.7b12b648.linux_amd64.deb`
`scylla-manager-agent-2.5-dev-0.20210722.7b12b648_linux.x86_64.rpm`

While if i remove the templates i will get:
`scylla-manager-agent_2.5-dev-0.20210722.7b12b648_linux_x86_64.rpm`
`scylla-manager-agent_2.5-dev-0.20210722.7b12b648_linux_amd64.deb`

So removing the templates will make us change the scripts in scylla-pkg and will add more code and complication
@penberg

dist/.goreleaser.yaml
yaronkaikov: Updated

yaronkaikov:
@penberg - ping :-)

mmatczuk:
Could we rename MANAGER-RELEASE-FILE and MANAGER-VERSION-FILE to `.release` and `.version`

yaronkaikov:
> Could we rename MANAGER-RELEASE-FILE and MANAGER-VERSION-FILE to `.release` and `.version`

i actually removed it , used it only for my debug sessions. but i can add it back as `.release` and `.version`

dist/Makefile
```diff
@@ -1,57 +1,29 @@
...
+$(shell git tag $(VERSION_NAME))
+endif
+
+GORELEASER := goreleaser --rm-dist
+
+.PHONY: release
+release:
+	echo $(RELEASE) > MANAGER-RELEASE-FILE
+	echo $(VERSION) > MANAGER-VERSION-FILE
+	$(GORELEASER) --skip-publish --skip-validate
```
mmatczuk, line 21: What's happening here, a comment would be good. I suggested a rename but maybe just RELEASE and VERSION. Do we intend to keep the files in git?

mmatczuk: If the files are not used let's remove that.

mmatczuk:
Overall LGTM.

PR #2846, opened 2021-08-20
healthcheck: fix deadlock during removal of lost nodes metrics (#2843)

zimnx:
When buffered channel used for collecting metrics was full,
goroutine trying to write to it was keeping read lock.
Reading side of this channel was deleting metrics which requires taking
a write lock on the same mutex. This resulted in deadlock.

Fixes #2843

PR #2847, opened 2021-08-20
[2.4] healthcheck: fix deadlock during removal of lost nodes metrics (#2843)

### LINKED ISSUE #2843: Prometheus endpoint hangs and ends up causing FD starvation
**Scylla Manager version**: 2.4.0-0.20210629.e62d606f
**Issue**: Queries to Scylla Manager metrics endpoint (`curl http://127.0.0.1:56090/metrics`) are hangs forever (i.e: keeps waiting for ~8 minutes and nothing is ever returned). As a side-effect, Scylla Manager may run out of file descriptors when the metrics endpoint is queried frequently.

Restarting the service does not help. For context, this is a large deployment with 12 clusters and ~180 nodes in total.
`lsof` output shows around 20K socket files opened, which seem around too much and may or not be related:

```
grep sock lsof_10783 | wc -l
19892
```

Deployment details (lsof output, `scylla_manager.cluster`, `scylla_manager.repair_run`, `scylla-manager.yaml`, and scylla manager logs) may be found under UUID `04245fd1-4bc6-4e03-9868-05b5aa8a053d`

zimnx:
When buffered channel used for collecting metrics was full,
goroutine trying to write to it was keeping read lock.
Reading side of this channel was deleting metrics which requires taking
a write lock on the same mutex. This resulted in deadlock.

Fixes #2843

### DIFF
--- a/pkg/service/healthcheck/metrics.go
+++ b/pkg/service/healthcheck/metrics.go
@@ -119,8 +119,8 @@ func init() {
 	)
 }
 
-func apply(ch <-chan prometheus.Metric, f func(cluster, dc, host, pt string, v float64)) {
-	for m := range ch {
+func apply(metrics []prometheus.Metric, f func(cluster, dc, host, pt string, v float64)) {
+	for _, m := range metrics {
 		metric := &dto.Metric{}
 		if err := m.Write(metric); err != nil {
 			continue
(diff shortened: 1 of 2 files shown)


zimnx:
Backport of https://github.com/scylladb/scylla-manager/pull/2846

PR #2851, opened 2021-08-23
dist/Makefile:Fix tag creation during build

yaronkaikov:
In #2810 we started building manager using goreleaser tool.

Once merging i saw an error (also managed to repreduce it locally)
```
fatal: No tags can describe '0b9ff39c6cd145048f1a2f32bbfe15dd335c1001'.
Try --always, or create some tags.
error: tag '2.4-dev-0.20210823.0b9ff39c' not found.
goreleaser --rm-dist --skip-publish --skip-validate
   • releasing...
   • loading config file       file=.goreleaser.yaml
   • loading environment variables
   • getting and validating git state
   ⨯ release failed after 0.02s error=git doesn't contain any tags. Either add a tag or use --snapshot
make: *** [Makefile:21: release] Error 1
```

Looks like i missed this during debug with some other changes

Fixing it here

Related to scylladb/scylla-pkg#2308

### DIFF
--- a/dist/Makefile
+++ b/dist/Makefile
@@ -1,28 +1,23 @@
 RELEASE      := 0.$(shell date +%Y%m%d).$(shell git describe --always)
+VERSION	     ?= 2.6-dev
 
-ifeq "$(VERSION)" ""
-$(error Please specify VERSION e.g. VERSION=2.0)
+ifdef $$VERSION
+VERSION := $$VERSION
 endif
-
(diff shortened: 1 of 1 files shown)


penberg:
@yaronkaikov What is the bug you're fixing?

yaronkaikov:
> @yaronkaikov What is the bug you're fixing?

commit description updated.

penberg (approved):
I am fine with this.

PR #2853, opened 2021-08-31
add upgrade content

lauranovich:
Adds the Scylla Manager Upgrade guides to Scylla Manager. 
in going forward, we should limit what content appears in each version, but this move allows all of the upgrade content to live with the rest of the documentation. 

---

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -18,6 +18,7 @@ Scylla Manager
    config/index
    swagger/index
    scylla-monitoring
+   upgrade/index
    troubleshooting
    older-versions
    Slack <https://scylladb-users.slack.com/archives/C01ERVBPWLU>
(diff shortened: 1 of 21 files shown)


mmatczuk:
Since we have it versioned let's add a single page describing the upgrade procedure avoid putting tons of archive pages on users.

mmatczuk (changes requested):
.

tzach:
@mmatczuk IMHO its better to have upgrade "outside" the versioning, so looking at 2.5, for example, I can find upgrades from 2.3 to 2.4 to 2.5 etc
It will make it easier for user to find, and for you to maintain.

mmatczuk:
On a site for version 2.5 there shall be stuff only relevant to that version.

lauranovich:
One question - do all of our users of Scylla Manager upgrade each time? 
OR do they wait and then upgrade meaning they upgrade 2 versions at the same time?
Once we know how our users upgrade, we can decide what to do w/ the older content

mmatczuk:
You just upgrade the package with package manager and it just works.
If it does not work like that we are doing something wrong.

PR #2854, opened 2021-09-01
service/scheduler: make suspend and resume idempotent (#2849)

### LINKED ISSUE #2849: Sctool suspend/resume improvements suggestion.
There's a suggestion that UX on the Manager 2.5 can be improved - when using `sctool suspend/resume`, the commands should not return an error if the cluster is already suspended/resumed. Here's an example:
```
$ sctool resume -c cluster_name; echo $?
0
$ sctool resume -c cluster_name; echo $?
Error: resume: cluster is not suspended
Trace ID: MAzxt1F1T6et_d01XHMJag (grep in scylla-manager logs)

1
```
Reasoning and expectation on commands that trigger a change of state is that they should be idempotent. In addition, there is no variance in the return code `sctool` provides (1), so it's impossible to differentiate between a strong error and an "already done error".

@mmatczuk please take a look. Thanks.

mmatczuk:
Fixes #2849

### DIFF
--- a/pkg/service/scheduler/service.go
+++ b/pkg/service/scheduler/service.go
@@ -640,7 +640,7 @@ func (s *Service) Suspend(ctx context.Context, clusterID uuid.UUID) error {
 
 	// Ignore if already suspended
 	if _, ok := s.suspended[clusterID]; ok {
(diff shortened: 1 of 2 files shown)


mmatczuk:
cc @ShlomiBalalis

PR #2863, opened 2021-09-09
healthcheck simplification

### LINKED ISSUE #2801: healthcheck: dynamic t/o simplification
Sometimes we see healthcheck metric flickering.
While the average part is easy to explain the stddev is not.
We could simplify it by allowing ppl to specify the max timeout over avg.

mmatczuk:
New implementation of dynamicTimeout
- API is changed we no longer return a function to set result, there are two public functions Timeout and Record used by Service and Runners
- The average algorithm is changed from mean to median
- The stddev multiplier is replaced by explicit setting of acceptable time offset to the median value

New config options

```
# Health-check service configuration.
#healthcheck:
# relative_timeout specifies timeout over median ping duration in probes.
# The number of probes kept in memory is specified by the probes parameter.
# There are separate probes for different DCs and ping types
# (CQL, REST, Alternator).
#  relative_timeout: 50ms
# max_timeout specifies maximum ping timeout, zero means no limit.
#  max_timeout: 30s
# Probes specifies how many probes are kept in memory for calculation.
# For different ping types and datacenters there are different probe sets.
#  probes: 200
```

Fixes #2801

### DIFF
--- a/dist/etc/scylla-manager.yaml
+++ b/dist/etc/scylla-manager.yaml
@@ -88,24 +88,19 @@ http: 127.0.0.1:5080
 #  user_cert_file:
 #  user_key_file
 
-# Healthcheck service configuration.
+# Health-check service configuration.
 #healthcheck:
-# Timeout for CQL status checks.
-#  timeout: 250ms
-#  ssl_timeout: 750ms
-#
-# Dynamic timeout calculates timeout based on past measurements.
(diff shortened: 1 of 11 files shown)


mmatczuk:
cc @ShlomiBalalis 

This may cause regression in some integration tests due to configuration changes.

PR #2908, opened 2021-11-02
service/repair: consider memory bloat when calculate repair parallelism

mmatczuk:
Adjust max parallel calculation to [changes in Scylla](https://github.com/scylladb/scylla/commit/4d1c83a4e8130962b1974c754dc1aa03efe65ab5).


mmatczuk:
Fixes #2908

PR #2917, opened 2021-11-04
scylla-manager:add arm64 support

yaronkaikov:
Recently we started building Scylla AMI image based on arm architecture.
Follow that, we had QA artifact tests failing due to the fact that they
are trying to install scylla-manager on arm based machines.

Adding the ability to build scylla-manager on both arm and x86 machines

### DIFF
--- a/dist/.goreleaser.yaml
+++ b/dist/.goreleaser.yaml
@@ -19,6 +19,7 @@ builds:
       - linux
     goarch:
       - amd64
+      - arm64
     env:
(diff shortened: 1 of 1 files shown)


mmatczuk (approved):

yaronkaikov:
@mmatczuk Can you please backport this to 2.6? 

/cc @ShlomiBalalis

mmatczuk:
It is already present in 2.6

PR #2924, opened 2021-11-17
Update index.rst

GavinJE:
Line 34. "Link is being repaired, please be patient" added.

---

Please make sure that:

### DIFF
--- a/docs/source/swagger/index.rst
+++ b/docs/source/swagger/index.rst
@@ -31,4 +31,4 @@ The following procedure instructs how to access a remote Scylla Manager Server U
 
       ssh -L 5080:localhost:5080 <host>
 
-#. Open `<http://localhost:5080/ui/>`_ in your browser.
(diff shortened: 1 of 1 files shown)


GavinJE:
The link in line 34 does not work due to a bug in the code. The code needs to be changed because the http command is not active.

mmatczuk:
What http command do you mean?

GavinJE:
Hi Michal.

When you hover over the link, no http precursor is seen and the link when
clicked does not function.

mmatczuk:
It does work when the ssh tunnel is in place, doesn't it?

GavinJE:
Not sure about that. Sorry.

guy9:
Thanks @GavinJE , @mmatczuk , this is a false positive and can be closed.

GavinJE:
@guy9 maintains this link only works locally and therefore is not an error. Pull request therefore closed.

PR #2937, opened 2021-11-24
Update dockerfile with missing packages

yaronkaikov:
Also unifing dockerfile package location so we can run docker build
command from dist folder

### DIFF
--- a/dist/docker/scylla-manager-agent.dockerfile
+++ b/dist/docker/scylla-manager-agent.dockerfile
@@ -6,7 +6,7 @@ RUN apt-get update && \
     apt-get autoremove && \
     rm -rf /var/lib/apt/lists/*
 
-COPY scylla-manager-agent*.deb /
+COPY release/deb/scylla-manager-agent*.deb /
 RUN dpkg -i scylla-manager-agent*.deb && rm /scylla-manager-agent*.deb
 
 USER scylla-manager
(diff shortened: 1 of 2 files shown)


mmatczuk:
Dockers are being built just fine with gorelaser.

mmatczuk:
`bash-completion` is a recommended package for client maybe install with recommended.

yaronkaikov:
we don't use golreaser for 2.5, only from 2.6

yaronkaikov:
> `bash-completion` is a recommended package for client maybe install with recommended.

How do i add it?

mmatczuk:
https://askubuntu.com/questions/117635/how-to-install-suggested-packages-in-apt-get

yaronkaikov:
> https://askubuntu.com/questions/117635/how-to-install-suggested-packages-in-apt-get

@mmatczuk looks like it's not recommended but it's a dependency 
```
dpkg: dependency problems prevent configuration of scylla-manager-client:
 scylla-manager-client depends on bash-completion; however:
  Package bash-completion is not installed.
```

mmatczuk:
The better.

mmatczuk:
Actually I want to get rid of this dependency. In master we now support bash zsh and fish. We can add the dependency as you described.

mmatczuk:
Why the PR is to 2.5 @yaronkaikov ?

yaronkaikov:
> Why the PR is to 2.5 @yaronkaikov ?

I faced this when releasing 2.5. I need first to run `make release/deb` and after that to build docker images using the dockerfile. 
In master, i just use the goreleaser. Should i do it in master anyway?

yaronkaikov:
@mmatczuk Fixed, now it's based on `master` , we should backport this to `2.5` in case we plan for more releases

PR #2946, opened 2021-11-30
goreleaser:set arch in tar.gz file to match os

yaronkaikov:
When running the command `uname -m` we get the machine arch type. Let's
align the reloc package filename to match it

### DIFF
--- a/dist/Makefile
+++ b/dist/Makefile
@@ -1,5 +1,5 @@
 RELEASE      := 0.$(shell date +%Y%m%d).$(shell git describe --always)
-VERSION	     ?= 2.6-dev
+VERSION	     ?= 3.0-dev
 
 ifdef $$VERSION
 VERSION := $$VERSION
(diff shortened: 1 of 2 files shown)


dist/Makefile
```diff
@@ -1,5 +1,5 @@
 RELEASE      := 0.$(shell date +%Y%m%d).$(shell git describe --always)
-VERSION	     ?= 2.6-dev
+VERSION	     ?= 2.7-dev
```
mmatczuk, line 2: Why do we want that?

yaronkaikov: now that we branched to 2.6 we should change the version in master to next dev release. This should be in master only no need to backport this

mmatczuk: Can we make it something generic? If not 3.0 is next release.

yaronkaikov: for packaging it will simply our process to know that next branch is currently under dev. This is also how we do it in Scylla.
Changed to 3.0-dev

PR #2975, opened 2021-12-15
cmd/agent: require auth-token to start

### LINKED ISSUE #2806: agent: don't start without auth_token
Currently, if no auth_token is specified in the config, manager-agent starts without any authentication and exposes unauthenticated API to the world while logging

```
{"L":"INFO","T":"2021-05-17T07:34:37.880Z","M":"WARNING! Scylla data may be exposed on IP 0.0.0.0, protect it by specifying auth_token in config file","config_file":["/etc/scylla-manager-agent/scylla-manager-agent.yaml","/mnt/scylla-agent-config/scylla-manager-agent.yaml"],"_trace_id":"AOKJu2SmSBqb8ok6kGzTNw"}
```

It is very easy for users to accidentally expose the API to the world. I think the application should never start as unauthenticated and missing auth token should result in a startup failure.

If there is a use case for no auth (like debugging), it should be an explicit flag, but the default should always start secured.

mmatczuk:
Fixes #2806

### DIFF
--- a/pkg/cmd/agent/root.go
+++ b/pkg/cmd/agent/root.go
@@ -60,6 +60,11 @@ var rootCmd = &cobra.Command{
 			logger.Sync() // nolint
 		}()
 
+		// Validate auth-token.
+		if c.AuthToken == "" {
+			return errors.New("configuration file must specify auth-token")
+		}
+
 		// Start server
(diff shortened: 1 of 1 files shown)


mmatczuk:
cc @tnozicka

PR #3016, opened 2022-02-01
sctool: unescape URL encoded cluster name

### LINKED ISSUE #3013: sctool backup command doesn't accept the name of cluster only ID
When you attempt to run the following
sctool backups -c clusterNS/clustername -L <bucket location>

sctool cannot find the cluster. I discussed this in our slack channel and @zimnx confirmed this is a bug with the following:

"Looks like a bug, clusters are registered under namespace/cluster-name and looks like / character is encoded wrong."
```
$ k -n scylla-manager exec -ti deployment.apps/scylla-manager -- sctool task list -c "scylla/simple-cluster"
Error: failed to get resource: load cluster "scylla%2Fsimple-cluster": not found
```

**work-around** 
You can use 
`sctool cluster -list `
to obtain the relevant cluster ID, and then use that in the relevant commands rather than the cluster name

A/C
To allow users to run the above command with the cluster name as well as ID

kevinbarbour:
fixes #3013

### DIFF
--- a/pkg/restapi/cluster.go
+++ b/pkg/restapi/cluster.go
@@ -26,7 +26,11 @@ func (h clusterFilter) clusterCtx(next http.Handler) http.Handler {
 			return
 		}
 
-		clusterID := chi.URLParam(r, "cluster_id")
+		clusterID, err := url.QueryUnescape(chi.URLParam(r, "cluster_id"))
+		if err != nil {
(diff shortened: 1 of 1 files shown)


mmatczuk (approved):

mmatczuk:
Thanks

PR #3025, opened 2022-02-08
service/backup: add retention days

### LINKED ISSUE #2923: backup: add retention days
The current retention policy is expressed in the number of instances.
This is good for controlling your storage but it would be also useful to be able to specify retention in days.

kevinbarbour:
- Allows configuring retention days per backup task
- Purge now handles both removing snapshots over retention policy, and
  snapshots older than retention days
- Adds a "deleted" property to tasks so that retention policy is retained
  after the task is deleted

Previously the behavior of purge was:
- If retention policy exists (> 0) apply it
- If retention policy does not exist delete snapshot older than
  threshold

New behavior is:
- If retention days policy exists (> 0) apply it
- If retention policy exists (> 0) and snapshot was not deleted above apply it
- If neither retention or retention days policy exist delete snapshot
  older than threshold

Fixes #2923

### DIFF
--- a/docs/source/sctool/partials/sctool_backup.yaml
+++ b/docs/source/sctool/partials/sctool_backup.yaml
@@ -98,6 +98,10 @@ options:
   default_value: "7"
   usage: |
     The `number of backups` to store, once this number is reached, the next backup which comes in from this destination will initiate a purge of the oldest backup.
+- name: retention-days
+  usage: |
+    The number of days to retain a backup. All backups older than this will be purged.
+    Set to 0 for no limit (default 0)
 - name: retry-wait
   default_value: 10m
   usage: |
--- a/pkg/schema/migrate/v3.0.0.go
+++ b/pkg/schema/migrate/v3.0.0.go
@@ -9,11 +9,14 @@ import (
 	"github.com/scylladb/gocqlx/v2"
 	"github.com/scylladb/gocqlx/v2/dbutil"
 	"github.com/scylladb/gocqlx/v2/migrate"
+	"github.com/scylladb/gocqlx/v2/qb"
 	"github.com/scylladb/gocqlx/v2/table"
+	"github.com/scylladb/scylla-manager/pkg/util/uuid"
 )
 
 func init() {
 	reg.Add(migrate.CallComment, "rewriteHealthCheck30", rewriteHealthCheck30)
+	reg.Add(migrate.CallComment, "setExistingTasksDeleted", setExistingTasksDeleted)
 }
 
 func rewriteHealthCheck30(ctx context.Context, session gocqlx.Session, ev migrate.CallbackEvent, name string) error {
@@ -91,6 +94,33 @@ func rewriteHealthCheck30(ctx context.Context, session gocqlx.Session, ev migrat
 
 	for _, m := range deleteKeys {
 		if err := dq.BindMap(m).Exec(); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func setExistingTasksDeleted(ctx context.Context, session gocqlx.Session, ev migrate.CallbackEvent, name string) error {
+	q := qb.Select("scheduler_task").Columns("id", "cluster_id", "type").Query(session)
+
+	type Task struct {
+		ClusterID uuid.UUID `db:"cluster_id"`
+		ID        uuid.UUID `db:"id"`
+		Type      string    `db:"type"`
+	}
+
+	var tasks []*Task
+	if err := q.SelectRelease(&tasks); err != nil {
+		return err
+	}
+
+	const setDeletedCql = `UPDATE scheduler_task SET deleted = false WHERE cluster_id = ? AND type = ? AND id = ?`
+	q = session.Query(setDeletedCql, nil)
+	defer q.Release()
+
+	for _, t := range tasks {
+		if err := q.Bind(t.ClusterID, t.Type, t.ID).Exec(); err != nil {
 			return err
 		}
 	}
--- a/pkg/service/backup/model.go
+++ b/pkg/service/backup/model.go
@@ -39,16 +39,17 @@ type ListItem struct {
 
 // Target specifies what should be backed up and where.
 type Target struct {
-	Units            []Unit     `json:"units,omitempty"`
-	DC               []string   `json:"dc,omitempty"`
-	Location         []Location `json:"location"`
-	Retention        int        `json:"retention"`
-	RetentionDays    int        `json:"retention_days"`
-	RateLimit        []DCLimit  `json:"rate_limit,omitempty"`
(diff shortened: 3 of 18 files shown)


docs/source/sctool/partials/sctool_backup.yaml
```diff
@@ -98,6 +98,10 @@ options:
   default_value: "7"
   usage: |
     The `number of backups` to store, once this number is reached, the next backup which comes in from this destination will initiate a purge of the oldest backup.
+- name: retention-days
+  usage: |
+    The number of days to retain a backups. All backups older than this will be purged.
```
mmatczuk, line 103: a backups
```suggestion
    The number of days to retain a backup. All backups older than this will be purged.
```

pkg/restapi/backup.go
```diff
@@ -79,7 +79,7 @@ func (h backupHandler) locationsCtx(next http.Handler) http.Handler {
 }
 
 func (h backupHandler) extractLocations(r *http.Request) ([]backupspec.Location, error) {
-	tasks, err := h.schedSvc.ListTasks(r.Context(), mustClusterIDFromCtx(r), scheduler.ListFilter{TaskType: []scheduler.TaskType{scheduler.BackupTask}})
+	tasks, err := h.schedSvc.ListTasks(r.Context(), mustClusterIDFromCtx(r), scheduler.ListFilter{TaskType: []scheduler.TaskType{scheduler.BackupTask}, Deleted: true})
```
mmatczuk, line 82: We should probably not extract locations from deleted tasks.
A location is valid provided that there is at least one backup using it.

pkg/schema/migrate/v3.0.0.go
```diff
@@ -97,3 +100,30 @@ func rewriteHealthCheck30(ctx context.Context, session gocqlx.Session, ev migrat
 
 	return nil
 }
+
+func setExistingTasksDeleted(ctx context.Context, session gocqlx.Session, ev migrate.CallbackEvent, name string) error {
```
mmatczuk, line 104: Why do we need it?

pkg/service/scheduler/model.go
```diff
@@ -30,6 +30,7 @@ const (
 	RepairTask         TaskType = "repair"
 	SuspendTask        TaskType = "suspend"
 	ValidateBackupTask TaskType = "validate_backup"
+	PurgeBackupsTask   TaskType = "purge_backups"
```
mmatczuk, line 33: Leftover?

pkg/service/backup/purger.go
```diff
@@ -23,31 +24,39 @@ import (
 // staleTags returns collection of snapshot tags for manifests that are "stale".
 // That is:
 // - temporary manifests,
+// - manifests over task days retention days policy,
 // - manifests over task retention policy,
 // - manifests older than threshold if retention policy is unknown.
-func staleTags(manifests []*ManifestInfo, policy map[uuid.UUID]int, threshold time.Time) *strset.Set {
+func staleTags(manifests []*ManifestInfo, policy map[uuid.UUID]Retention, threshold time.Time) *strset.Set {
```
mmatczuk, line 30: I think we replace `policy map[uuid.UUID]Retention, threshold time.Time` with a `func(uuid.UUID) Retention`.

mmatczuk (commented):
Overall it looks good, see comments inline.

pkg/schema/migrate/v3.0.0.go
kevinbarbour: As discussed, keeping this so that we can query for all not-deleted tasks.

pkg/service/scheduler/model.go
kevinbarbour: Yes, my mistake. Removed this and other mentions.

pkg/service/backup/purger.go
kevinbarbour: Added `type RetentionFunc func(uuid.UUID) Retention` which is now a property of the backup target in place of the retention map.

The function checks for policies in the retention map and if none exists returns a default of 30 retention days, which was the old threshold to delete backups with no associated task.

pkg/service/backup/model.go
```diff
@@ -39,21 +39,34 @@ type ListItem struct {
...
+	RetentionDays    int        `json:"retention_days"`
+	RateLimit        []DCLimit  `json:"rate_limit,omitempty"`
+	SnapshotParallel []DCLimit  `json:"snapshot_parallel,omitempty"`
+	UploadParallel   []DCLimit  `json:"upload_parallel,omitempty"`
+	Continue         bool       `json:"continue,omitempty"`
+	PurgeOnly        bool       `json:"purge_only,omitempty"`
+
+	// GetRetention (injected in runtime) returns the policy for all backup
+	// tasks. Returns a default policy if no retention exists.
+	GetRetention RetentionFunc `json:"-"`
```
mmatczuk, line 55: I'd suggest TaskRetention.

pkg/service/backup/service.go
```diff
@@ -83,6 +83,20 @@ func (s *Service) Runner() Runner {
 	return Runner{service: s}
 }
 
+// GetRetentionFrom returns a function that returns the retention policy for a
+// given task ID, with a default of 30 days retention if no policy exists.
+func GetRetentionFrom(retentionMap map[uuid.UUID]Retention) func(uuid.UUID) Retention {
```
mmatczuk, line 88: Should all of this be exported, I mean Retention and all related structs and functions?

pkg/service/backup/model.go
kevinbarbour: Moved this out of the target and into the backup service. Now the retention policies are gathered within the backup service when the purge is executed instead of being injected into each target's properties from the scheduler service.

They also no longer need to be exported as they don't to be injected from the integration tests like before.

mmatczuk (changes requested):
Almost good.

Backup is not allowed to read Scheduler tables.

Let's 
- leave RetentionMap in Target
- have pkg level function that takes the map and returns the policy function

BTW does deleting work with task names i.e. can I add task "A" delete it and add another task named "A".

kevinbarbour:
> Let's
> 
> * leave RetentionMap in Target
> * have pkg level function that takes the map and returns the policy function

Sounds good, made these changes.

> BTW does deleting work with task names i.e. can I add task "A" delete it and add another task named "A".

Ah, good point. The task name doesn't have to be unique in the table, but the application does check and reject a task if another with a matching name exists. Looking into a fix.

kevinbarbour:
Updated the deletion so that when a task is deleted its name is cleared.

However, I did find an existing bug with task deletion. If you delete a task and then try to create a new task with the same name it does not let you, because the resolver map in memory still contains the task and name. After restarting the manager server you can then create a new task with the same name. Will create an issue for this.

PR #3027, opened 2022-02-08
config/agent: use s3 defaults when config is empty

### LINKED ISSUE #3026: config/agent: empty S3 key in config overwrites default
If S3 credentials are not provided in the configuration file the default behavior is to use the environment's configured IAM role to authenticate. Currently if you uncomment the key for the S3 configuration but leave all of the contents empty, instead of using the IAM role it expects the credentials to be configured and will try to connect anonymously.

The behavior can be seen with this config:
```
[...]
s3:
# S3 credentials, it's recommended to use IAM roles if possible, otherwise set
# your AWS Access Key ID and AWS Secret Access Key (password) here.
#  access_key_id:
#  secret_access_key:
[...]
```

kevinbarbour:
Fixes #3026

### DIFF
--- a/pkg/config/agent/agent.go
+++ b/pkg/config/agent/agent.go
@@ -84,7 +84,9 @@ func DefaultConfig() Config {
 // config struct with merged configuration from all provided files.
 func ParseConfigFiles(files []string) (Config, error) {
 	c := DefaultConfig()
-	return c, cfgutil.ParseYAML(&c, files...)
+	err := cfgutil.ParseYAML(&c, DefaultConfig(), files...)
+
+	return c, err
 }
 
 func (c Config) Validate() (errs error) {
--- a/pkg/config/agent/agent_test.go
+++ b/pkg/config/agent/agent_test.go
@@ -60,6 +60,11 @@ func TestParseConfig(t *testing.T) {
 			Input:  []string{"./testdata/prometheus_overwrite.input.yaml"},
 			Golden: "./testdata/prometheus_overwrite.golden.yaml",
 		},
+		{
+			Name:   "structs empty",
+			Input:  []string{"./testdata/structs_empty.input.yaml"},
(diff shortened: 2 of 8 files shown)


pkg/config/agent/agent.go
```diff
@@ -84,7 +84,16 @@ func DefaultConfig() Config {
 // config struct with merged configuration from all provided files.
 func ParseConfigFiles(files []string) (Config, error) {
 	c := DefaultConfig()
-	return c, cfgutil.ParseYAML(&c, files...)
+	err := cfgutil.ParseYAML(&c, files...)
+
+	// If the config YAML has a blank S3 key with nothing defined it will still
+	// overwrite the default config, and not use the IAM role. Fall back to the
+	// default if that is the case.
+	if (rclone.S3Options{}) == c.S3 {
```
mmatczuk, line 92: I guess this is true for all options

kevinbarbour: Yeah - if any of these exist in the YAML with no properties it will overwrite everything.

From what I can tell the other providers don't have an option to use authentication details from the environment though so it wouldn't have the same effect of breaking something that works by default.

pkg/config/agent/agent.go
```diff
@@ -84,7 +85,22 @@ func DefaultConfig() Config {
...
 func ParseConfigFiles(files []string) (Config, error) {
 	c := DefaultConfig()
-	return c, cfgutil.ParseYAML(&c, files...)
+	err := cfgutil.ParseYAML(&c, files...)
+
+	mergeConfig(&c, DefaultConfig())
+	return c, err
+}
+
+func mergeConfig(parsed *Config, def Config) {
```
mmatczuk, line 94: Let's put that cfgutil, and use in ParseYAML directly - we would make a shallow copy of the passed pointer and use that as def.

Issue #3007, opened 2022-01-12
docs: sctool, do not add Inherited options to every command


mmatczuk:
The sctool docs are littered by `Inherited options` that really are what we have in `Global flags and variables`.

I suggest we update the Global flags and variables section with the new data, and remove the Inherited options from the template.

mmatczuk:
@dgarcia360 PTAL

dgarcia360:
I'm  considering [adding a collapsible component](https://github.com/scylladb/sphinx-scylladb-theme/issues/276) to hide "inherited options". What do you think?

mmatczuk:
That could work, you can then remove the Global flags section IMO.

This is a more flexible approach.

PR #3030, opened 2022-02-11
docs: collapse inherited options

dgarcia360:
Closes https://github.com/scylladb/scylla-manager/issues/3007

Preview:

image


mmatczuk:
Why would we have the `[` in ['File path

dgarcia360:
@mmatczuk Good catch! Fixed in https://github.com/scylladb/scylla-manager/pull/3030/commits/f4da8815a1b8d9c3c6e768fb7bc75f21f21e1df7

image

Issue #3039, opened 2022-02-18
scyllaclient: rclone, listing t/o improvement


mmatczuk:
This is the context of https://github.com/scylladb/scylla-enterprise/issues/2083
precisely RcloneListDir call in backup purge init phase.

RcloneListDir should not be used with recursive mode, this should be documented if not enforced.

RcloneListDir in contrast to RcloneListDirIter support retries with t/o increasing.
The time it takes to do a recursive listing is a function of nr of files and directories, each dir is a new AWS call, paging is 1k items but it seems that rclone sends additional head requests to get some details.
The fixed timeout in this situation is like trying to guess how many files there is and it overall poor use of resources.

RcloneListDirIter has a different problem the timeout is set to 12h...

The plan
* Add NoModTime to the call to avoid the head requests
* Switch to RcloneListDirIter
* Fix RcloneListDirIter - change semantics of ListTimeout, set ListTimeout to 5m and evaluate it between batches i.e. if client does not get another batch (flush) of files during ListTimeout the context is canceled, we shall add some unit tests.

mmatczuk:
I'd add NoModTime to 2.6 branch that should radically reduce nr of AWS calls.

PR #3050, opened 2022-02-22
scyllaclient: t/o per batch on RcloneListDirIter

kevinbarbour:
Instead of passing a context into the HTTP request with a 12 hour
timeout, we now set a 5 minute timeout but only enforce it per batch of
files. This is done by waiting for the next JSON decode from the HTTP
request body and selecting between that and a 5 minute timeout each time
we process a new batch, cancelling the context if the timeout is reached.

Had some trouble coming up with a decent way to get this working since we can't really put it in the context, as it needs to be started fresh every time we get something new over the line. Let me know if any feedback on an alternate way to approach this.

Relates to #3039

### DIFF
--- a/pkg/scyllaclient/client_rclone.go
+++ b/pkg/scyllaclient/client_rclone.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/scylladb/scylla-manager/pkg/rclone/rcserver"
@@ -416,8 +417,10 @@ func (c *Client) RcloneListDir(ctx context.Context, host, remotePath string, opt
 // The remotePath is given in the following format "provider:bucket/path".
 // Resulting item path is relative to the remote path.
 func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string, opts *RcloneListDirOpts, f func(item *RcloneListDirItem)) error {
-	ctx = customTimeout(ctx, c.config.ListTimeout)
+	ctx = noTimeout(ctx)
 	ctx = noRetry(ctx)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	// Due to OpenAPI limitations we manually construct and sent the request
 	// object to stream process the response body.
@@ -460,22 +463,45 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
 		}
 	}
 
-	var v RcloneListDirItem
-	for dec.More() {
-		// Detect context cancellation
-		if ctx.Err() != nil {
-			return ctx.Err()
-		}
+	errCh := make(chan error)
+	go func() {
+		var v RcloneListDirItem
+		for dec.More() {
+			// Detect context cancellation
+			if ctx.Err() != nil {
+				errCh <- ctx.Err()
+				return
+			}
 
-		// Read value
-		v = RcloneListDirItem{}
-		if err := dec.Decode(&v); err != nil {
-			return err
+			// Read value
+			v = RcloneListDirItem{}
+			if err := dec.Decode(&v); err != nil {
+				errCh <- err
+				return
+			}
+			f(&v)
+
+			errCh <- nil
 		}
-		f(&v)
+		close(errCh)
+	}()
+
+	timer := time.NewTimer(c.config.ListTimeout)
+	defer timer.Stop()
+	for {
+		select {
+		case err, ok := <-errCh:
+			if !ok {
+				return nil
+			}
+			if err != nil {
+				return err
+			}
+			timer.Reset(c.config.ListTimeout)
+		case <-timer.C:
+			return errors.Errorf("rclone list dir timeout")
+		}
 	}
-
-	return nil
 }
 
 // RcloneCheckPermissions checks if location is available for listing, getting,
--- a/pkg/scyllaclient/client_rclone_test.go
+++ b/pkg/scyllaclient/client_rclone_test.go
@@ -5,7 +5,9 @@ package scyllaclient_test
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"os"
 	"path"
@@ -13,10 +15,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-openapi/strfmt"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/scylladb/scylla-manager/pkg/scyllaclient"
 	"github.com/scylladb/scylla-manager/pkg/scyllaclient/scyllaclienttest"
+	"github.com/scylladb/scylla-manager/swagger/gen/agent/models"
 )
 
 func TestRcloneSplitRemotePath(t *testing.T) {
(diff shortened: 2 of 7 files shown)


pkg/scyllaclient/client_rclone.go
```diff
@@ -8,8 +8,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
```
mmatczuk, line 11: We do not use this package.

pkg/scyllaclient/client_rclone.go
```diff
@@ -460,22 +463,42 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
...
 
+	more := make(chan bool)
+	errCh := make(chan error)
+
 	var v RcloneListDirItem
-	for dec.More() {
-		// Detect context cancellation
-		if ctx.Err() != nil {
-			return ctx.Err()
+	go func() {
```
mmatczuk, line 470: It looks like we cannot do without a go routine, but we shall clean it up.

The reading go routine should take a token form channel and return result to channel, on time out the token channel is closed. In other words one channel in one channel out.

mmatczuk:
There should be a commit where this is actually used in manifest listing in backup.

pkg/scyllaclient/client_rclone.go
```diff
@@ -460,22 +463,42 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
 		}
 	}
 
+	more := make(chan bool)
```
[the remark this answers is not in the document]
kevinbarbour: The boolean value is actually being used. If `dec.More()` is true it passes true to the channel so that the timer restarts. If it is false it passes false through the channel so that the loop below can finish execution and return from the function.

pkg/scyllaclient/client_rclone.go
kevinbarbour: Sounds good will clean up. Could you explain though what you mean by "token form channel"?

pkg/scyllaclient/client_rclone.go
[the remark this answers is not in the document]
mmatczuk: I see, I guess it could be refactored to a done channel that you would close as there is nothing more to do.

pkg/scyllaclient/client_rclone.go
mmatczuk: Let me elaborate a little bit...

I think that we could have a single output channel from the go routine.
It technically can be the error channel:
* nil error - OK
* not nil error - NOK
* closed - Done

Also, there need to be a way to inform the go routine to stop.
This is only relevant when it is locked on More() refilling the data...

When it times out we should close the body making the More() return false.

mmatczuk: Ergo the input channel may not exist.

kevinbarbour: Thanks - implemented like this and it works great and is cleaner.

I did not close the body explicitly when it times out since that is deferred earlier on and we are returning with an error when there is a timeout.

pkg/scyllaclient/client_rclone.go
```diff
@@ -416,8 +417,9 @@ func (c *Client) RcloneListDir(ctx context.Context, host, remotePath string, opt
 // The remotePath is given in the following format "provider:bucket/path".
 // Resulting item path is relative to the remote path.
 func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string, opts *RcloneListDirOpts, f func(item *RcloneListDirItem)) error {
-	ctx = customTimeout(ctx, c.config.ListTimeout)
 	ctx = noRetry(ctx)
+	ctx, cancel := context.WithCancel(ctx)
```
mmatczuk, line 421: I guess we need to explicitly disable the timeout.

pkg/scyllaclient/client_rclone.go
```diff
@@ -460,22 +462,43 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
...
+			f(&v)
+
+			ch <- nil
 		}
-
-		// Read value
-		v = RcloneListDirItem{}
-		if err := dec.Decode(&v); err != nil {
-			return err
+		close(ch)
```
mmatczuk, line 484: Let's add that in defer and use return instead of break

pkg/scyllaclient/client_rclone.go
```diff
@@ -460,22 +462,43 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
...
+			// Detect context cancellation
+			if ctx.Err() != nil {
+				ch <- ctx.Err()
+				break
+			}
+
+			// Read value
+			v = RcloneListDirItem{}
+			if err := dec.Decode(&v); err != nil {
+				ch <- err
```
mmatczuk, line 478: Should we return here too?

pkg/scyllaclient/client_rclone.go
```diff
@@ -460,22 +462,43 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
...
+
+			ch <- nil
 		}
-
-		// Read value
-		v = RcloneListDirItem{}
-		if err := dec.Decode(&v); err != nil {
-			return err
+		close(ch)
+	}(errCh)
```
mmatczuk, line 485: We use ctx from the outer scope why pass errCh as ch?
It makes things less readable with no clear goal.

pkg/scyllaclient/client_rclone.go
```diff
@@ -460,22 +462,43 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
...
+			}
+
+			// Read value
+			v = RcloneListDirItem{}
+			if err := dec.Decode(&v); err != nil {
+				ch <- err
+			}
+			f(&v)
+
+			ch <- nil
```
mmatczuk, line 482: It would be even nicer to have just `ch <- err` sometimes the err is nil

pkg/scyllaclient/client_rclone.go
```diff
@@ -460,22 +462,43 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
...
+
+	timer := time.NewTimer(c.config.ListTimeout)
+	for {
+		select {
+		case err, ok := <-errCh:
+			if !ok {
+				return nil
+			}
+			if err != nil {
+				return errors.Wrap(err, "list rclone dir iter")
```
mmatczuk, line 495: The "list rclone dir iter" message is for caller to add

pkg/scyllaclient/client_rclone.go
```diff
@@ -460,22 +462,43 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
...
 		}
-
-		// Read value
-		v = RcloneListDirItem{}
-		if err := dec.Decode(&v); err != nil {
-			return err
+		close(ch)
+	}(errCh)
+
+	timer := time.NewTimer(c.config.ListTimeout)
```
mmatczuk, line 487: You can be very nice and stop it on exit

pkg/scyllaclient/client_rclone.go
mmatczuk: Oh the closing must only happen in the happy path.
On error it should not be closed.

pkg/scyllaclient/client_rclone.go
```diff
@@ -460,22 +462,43 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
...
+	errCh := make(chan error)
+	go func(ch chan error) {
+		var v RcloneListDirItem
+		for dec.More() {
+			// Detect context cancellation
+			if ctx.Err() != nil {
+				ch <- ctx.Err()
+				break
+			}
+
```
mmatczuk, line 474: Could we get rid of that?

kevinbarbour: Yeah shouldn't be needed, if it's cancelled or we time out `More()` should come back false.

pkg/scyllaclient/client_rclone.go
kevinbarbour: Not sure what you mean. If this context comes from a caller with a different timeout attached shouldn't we still respect that?

mmatczuk: Sure. Check [this](https://github.com/scylladb/scylla-manager/blob/master/pkg/scyllaclient/timeout.go) atm there is no need to send something with no timeout for the whole operation.

Try changing server timeout parameter in the test.

mmatczuk:
As a general comment maybe we should move the tests to client_rclone.go next to other RcloneListDir tests.

pkg/scyllaclient/client_rclone.go
kevinbarbour: Ah, now I understand. As it is now there doesn't appear to be a way to disable it. We could either give it a very long value for the custom timeout within this Iter function the way it did the 12h before, or can modify the timeout function so it considers a duration of 0 to be no timeout.

mmatczuk: Let's add a separate function noTimeout, it's more explicit and it's easier to track it's users.

pkg/scyllaclient/client_rclone.go
kevinbarbour: Left this as-is because in both cases where err gets set we now return immediately.

pkg/service/backup/list.go
```diff
@@ -50,7 +50,11 @@ func listManifests(ctx context.Context, client *scyllaclient.Client, host string
 		FilesOnly: true,
 		Recurse:   true,
 	}
-	items, err := client.RcloneListDir(ctx, host, location.RemotePath(baseDir), &opts)
+	var items []*scyllaclient.RcloneListDirItem
```
mmatczuk, line 53: Let's extend that to the loop below.

pkg/scyllaclient/context.go
```diff
@@ -46,6 +47,13 @@ func noRetry(ctx context.Context) context.Context {
 	return context.WithValue(ctx, ctxNoRetry, true)
 }
 
+// noTimeout disables timeouts. This should only be used by
+// RcloneListDirIter as it handles its own timeouts per batch and can run much
+// longer than is typically desired.
```
mmatczuk, line 52:
```suggestion
// noTimeout disables timeouts - if in doubt do not use it.
// This should only be used by functions that handle timeouts internally.
```

pkg/scyllaclient/client_rclone_test.go
```diff
@@ -494,3 +498,66 @@ func TestRclonePut(t *testing.T) {
 		t.Fatalf("put/a = %s, expected %s", string(content), "hello")
 	}
 }
+
+func TestRcloneTimeouts(t *testing.T) {
```
mmatczuk, line 502:
```suggestion
func TestRcloneListDirTimeouts(t *testing.T) {
```

pkg/scyllaclient/client_rclone_test.go
```diff
@@ -494,3 +498,66 @@ func TestRclonePut(t *testing.T) {
...
+		}
+
+		w.Write(b)
+		w.Write([]byte("]}"))
+	})
+
+	host, port, closeServer := scyllaclienttest.MakeServer(t, s)
+	defer closeServer()
+
+	t.Run("ListDir", func(t *testing.T) {
```
mmatczuk, line 535:
```suggestion
	t.Run("default", func(t *testing.T) {
```

pkg/scyllaclient/client_rclone_test.go
```diff
@@ -494,3 +498,66 @@ func TestRclonePut(t *testing.T) {
...
+		})
+
+		l, err := client.RcloneListDir(context.Background(), host, "s3:foo", nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+		t.Logf("got %d items", len(l))
+	})
+
+	t.Run("ListDirIter", func(t *testing.T) {
```
mmatczuk, line 547:
```suggestion
	t.Run("eter", func(t *testing.T) {
```

mmatczuk (approved):

PR #3051, opened 2022-02-23
rclone: use server mod time by default

### LINKED ISSUE #3046: rclone: s3, listing with mod time shall not require n HEAD requests
Using ModTime i.e. listing with `NoModTime` (or `NoMimeType`) set to `false` (default) causes HEAD request being sent for each file. This is typical n+1 problem instance single request to get a list of file and n HEAD requests. It add a HUGE performance and cost overhead.

One can test this with this simple test

```go
func TestRcloneS3ListIntegration(t *testing.T) {
	client, closeServer := scyllaclienttest.NewFakeRcloneServer(t)
	defer closeServer()

	ctx := context.Background()
	opts := scyllaclient.RcloneListDirOpts{
		Recurse:    true,
		FilesOnly:  true,
		NoModTime:  true,
		NoMimeType: true,
	}
	x, err := client.RcloneListDir(ctx, scyllaclienttest.TestHost, "s3:my-foo/backup/meta/cluster/5c3667c9-e8da-4a75-8541-a48b7c2b3db4", &opts)
	if err != nil {
		t.Fatal(err)
	}
	fmt.Printf("%+v, %d\n", x[0], len(x))
}
```

Make sure to add `c.Dump = fs.DumpRequests` to rclone config

One can observe that [API_ListObjectsV2](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html) **has** the `<LastModified>2013-09-17T18:07:53.000Z</LastModified>`.

Validate is the only use case for that but very important as the data volume is huge.

kevinbarbour:
This prevents rclone from performing a HEAD request for each file when
listing objects.

fixes #3046

### DIFF
--- a/pkg/config/agent/testdata/auth_token_overwrite.golden.yaml
+++ b/pkg/config/agent/testdata/auth_token_overwrite.golden.yaml
@@ -79,7 +79,7 @@ rclone:
   stats_file_name_length: 45
   ask_password: true
   password_command: []
-  use_server_mod_time: false
+  use_server_mod_time: true
   max_transfer: -1
   max_duration: 0s
   cutoff_mode: 0
--- a/pkg/config/agent/testdata/basic.golden.yaml
+++ b/pkg/config/agent/testdata/basic.golden.yaml
@@ -79,7 +79,7 @@ rclone:
   stats_file_name_length: 45
   ask_password: true
   password_command: []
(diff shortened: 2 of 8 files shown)


mmatczuk (approved):

mmatczuk:
❤️

PR #3074, opened 2022-03-14
CI: add release workflow for auto-generate changelog and release

benipeled:
The release will trigger by a tag-push and should look like:

image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,32 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'scylla-manager-**'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set go version
+        run: echo "GOVERSION=$(cat .go-version)" >> $GITHUB_ENV
+
+      - name: Set up go
+        uses: actions/setup-go@v2
+        with:
+          go-version: "${{env.GOVERSION}}"
+
+      - name: Release
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
+          workdir: dist
+          args: release
+        env:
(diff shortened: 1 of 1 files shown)


yaronkaikov (approved):
Nice!!!

benipeled:
@mmatczuk please review.

mmatczuk:
These days when you manually create release the change log is auto generated too.

There is also changelog in release that we make generated by go releaser maybe it would be better to use that.
The important bit here is not including all commits just the relevant ones.

mmatczuk:
Why we decided to use `metcalfc/changelog-generator`?

benipeled:
> There is also changelog in release that we make generated by go releaser maybe it would be better to use that.

I'm not familiar with this option but I guess you're talking about https://goreleaser.com/customization/release/#github which seems simple to implement but in any case, require a github-action workflow with something like `goreleaser release`,

> The important bit here is not including all commits just the relevant ones.

If you're talking about the final release-notes posted on the site - it anyway requires manual editing,
With this automation, the github-release page can be the base for creating the final release-notes - WDYT?

> Why we decided to use metcalfc/changelog-generator?

Just because of the simplicity it provides

mmatczuk:
This is not the final solution IMO

Maybe the gorelaser changelog file can be committed together with version bump and tagging?

It's difficult to see what it would have produce for versions we already have.
We have some exclude rules [here](https://github.com/scylladb/scylla-manager/blob/master/dist/.goreleaser.yaml#L4) and it's easy to manage and validate them.

benipeled:
@mmatczuk, following our discussion, I'll adjust the pipeline as following

1. Use `goreleaser` as the release CI tool
2. Add a github-actions workflow that will use `goreleaser-action` following a new-tag event.

The only "problem" with this implementation is that goreleaser enforces semantic versioning [0] which means we'll have to change the tag-convention `scylla-manager-x.y.z` -> `vx.y.z`  - do you know something that depends on this convention?

[0] https://goreleaser.com/limitations/semver/

PR #3091, opened 2022-04-06
managerclient: don't limit status column

### LINKED ISSUE #3065: Task history shows status of "..."
Manager build: http://downloads.scylladb.com/manager/relocatable/unstable/branch-3.0/2022-03-05T21:57:56Z/scylla-manager_3.0-dev-0.20220305.e0ef361ca85.linux_x86_64.tar.gz

When trying to see the status of of previous runs of a task with the `task history` command, the status column shows an empty three dot status instead of the actual one:
```
+--------------------------------------+------------------------+----------+--------+
| ID                                   | Start time             | Duration | Status |
+--------------------------------------+------------------------+----------+--------+
| fc7234d9-98ff-11ec-aef2-f4ee08c9cc47 | 06 Mar 22 15:36:16 IST | 2s       | …      |
| faef6767-98ff-11ec-ac9c-f4ee08c9cc47 | 06 Mar 22 15:36:13 IST | 2s       | …      |
+--------------------------------------+------------------------+----------+--------+
```

kevinbarbour:
As far as I can tell this limit really didn't bring any benefit to begin
with. Every status is almost as short as the table header "Status"
anyway and the other columns in the table are much longer and more
likely to cause the lines to wrap. There is also only a 1-2 width window
where this could actually be beneficial, and in other cases it leads to
every status showing as "…" which can cause confusion.

fixes #3065

### DIFF
--- a/pkg/managerclient/model.go
+++ b/pkg/managerclient/model.go
@@ -477,7 +477,6 @@ type TaskRunSlice []*TaskRun
 // Render renders TaskRunSlice in a tabular format.
 func (tr TaskRunSlice) Render(w io.Writer) error {
 	t := table.New("ID", "Start time", "Duration", "Status")
-	t.LimitColumnLength(3)
 	for _, r := range tr {
 		s := r.Status
 		if r.Cause != "" {


mmatczuk:
It was never intended to limit that column. A column was removed and the index moved.

PR #3109, opened 2022-04-12
gocql hardening

mmatczuk:


mmatczuk:
Unfortunately in a situation where we have only one host the value of retry policy is zero.
In case of timeout or a connection error the connection is restarted, host is marked as down and eliminated from policy even if the pool exists.

kevinbarbour (approved):

PR #3120, opened 2022-04-14
scyllaclient: expose config to config file

### LINKED ISSUE #3112: Export scylla client config in the config file
Currently to change the values one needs to recompile - impossible to use in a production setting.

kevinbarbour:
I created an additional struct containing only the user-configurable
parts of the client config, and embedded them within the config. This is
because fields like Hosts and Port are populated from the cluster and
they would just be overriden if exposed in the configuration.

fixes #3112

### DIFF
--- a/pkg/config/server/testdata/scylla-manager.yaml
+++ b/pkg/config/server/testdata/scylla-manager.yaml
@@ -46,3 +46,18 @@ repair:
   long_polling_timeout_seconds: 5
   age_max: 12h
   graceful_stop_timeout: 60s
+
+agent_client:
+  timeout: 45s
+  max_timeout: 5h
+  list_timeout: 7m
+  backoff:
+    wait_min: 5s
+    wait_max: 20s
+    max_retries: 12
+    multiplier: 8
+    jitter: 0.6
+  interactive_backoff:
+    wait_min: 2s
+    max_retries: 4
+  pool_decay_duration: 1h
--- a/dist/etc/scylla-manager.yaml
+++ b/dist/etc/scylla-manager.yaml
@@ -133,3 +133,25 @@ http: 127.0.0.1:5080
 # Distribution of data among cores (shards) within a node.
 # Copy value from Scylla configuration file.
 #  murmur3_partitioner_ignore_msb_bits: 12
+
+# Connection configuration to Scylla Agent.
+#  agent_client:
+#
+# Timeouts configuration.
+#    timeout: 30s
+#    max_timeout: 1h
+#    list_timeout: 5m
+#
+# Connection backoff configuration.
+#    backoff:
+#      wait_min: 1s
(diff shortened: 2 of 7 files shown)


pkg/config/server/testdata/scylla-manager.yaml
```diff
@@ -46,3 +46,19 @@ repair:
   long_polling_timeout_seconds: 5
   age_max: 12h
   graceful_stop_timeout: 60s
+
+client:
```
mmatczuk, line 50: `client` is too generic, maybe `agent` since it controls how manager server talks to agent?

mmatczuk, line 50: `client` is too generic, maybe `agent` since it controls how manager server talks to agent?

pkg/config/server/testdata/scylla-manager.yaml
```diff
@@ -46,3 +46,19 @@ repair:
...
+  scheme: http
```
mmatczuk, line 51: Scheme is https there is no way agent can run http.

mmatczuk, line 51: Scheme is https there is no way agent can run http.

pkg/config/server/testdata/scylla-manager.yaml
```diff
@@ -46,3 +46,19 @@ repair:
...
+  timeout: 45s
```
mmatczuk, line 52: Isn't it 30s?

mmatczuk, line 52: Isn't it 30s?

kevinbarbour: This test is for setting a modified config, not the default values. The values in here are all arbitrary chosen for a test.

pkg/config/server/testdata/scylla-manager.yaml
kevinbarbour: This is also an arbitrary value to make sure we can set config values. If it can only be https though then there's no reason to expose it in config at all. Will remove it as a configurable option.

mmatczuk:
We need to document the defaults in the config file template in `dist`.

mmatczuk:
We need to document the defaults in the config file template in `dist`.

mmatczuk (changes requested):
.

kevinbarbour:
@mmatczuk 

* Renamed client config to agent
* Added with defaults in config file template
* Removed scheme from configurable values

mmatczuk:
It's all good but the naming is not consistent.
I suggest we use

* `agent_client` - in yaml
* `TimeoutConfig` - everywhere else

kevinbarbour:
Sounds good to me, will update the naming.

PR #3126, opened 2022-04-21
service/backup: split manifest load and deletion

### LINKED ISSUE #3113: backup: purge limit parallelism of manifests loading
What you can see below is starting a backup purge process (with --purge-only).
It loads and processes manifests, reaches a plateau, almost OOMs, and moves on to deletes.

image

The `PurgeSnapshotTags` func needs to be split to:
- reading phase (1 or X hosts in parallel depending on available memory)
- delete phase (full parallelism)

The `worker::Purge` would be like:

```
for host {
load manifests and calculate files to delete
[async] go delete the files
}
[async] join deletes
```

kevinbarbour:
Loading all manifests in parallel can be memory-intensive. This splits
the loading of manifests and the actual file deletion in the purge step.
This way by default manifests are loaded only one at a time, but a
configurable parallelism can be provided if memory is not as constrained
and they can be loaded in parallel again.

fixes #3113

### DIFF
--- a/pkg/command/backup/cmd.go
+++ b/pkg/command/backup/cmd.go
@@ -37,6 +37,7 @@ type command struct {
 	dryRun           bool
 	showTables       bool
 	purgeOnly        bool
+	purgeParallel    int
 }
 
 func NewCommand(client *managerclient.Client) *cobra.Command {
@@ -82,6 +83,7 @@ func (cmd *command) init() {
 	w.Keyspace(&cmd.keyspace)
 	w.Unwrap().IntVar(&cmd.retention, "retention", 7, "")
 	w.Unwrap().IntVar(&cmd.retentionDays, "retention-days", 0, "")
+	w.Unwrap().IntVar(&cmd.purgeParallel, "purge-parallel", 1, "")
 	w.Unwrap().StringSliceVar(&cmd.rateLimit, "rate-limit", nil, "")
 	w.Unwrap().StringSliceVar(&cmd.snapshotParallel, "snapshot-parallel", nil, "")
 	w.Unwrap().StringSliceVar(&cmd.uploadParallel, "upload-parallel", nil, "")
@@ -152,6 +154,10 @@ func (cmd *command) run(args []string) error {
 	}
 	if cmd.Flag("purge-only").Changed {
 		props["purge_only"] = cmd.purgeOnly
+		ok = true
+	}
+	if cmd.Flag("purge-parallel").Changed {
+		props["purge-parallel"] = cmd.purgeParallel
 		ok = true
 	}
 
--- a/pkg/command/backup/res.yaml
+++ b/pkg/command/backup/res.yaml
@@ -34,3 +34,8 @@ show-tables: |
 
 purge-only: |
   Run the backup cleanup only.
+
+purge-parallel: |
+  The number of hosts to read manifests from in parallel.
+  The reading of manifests can be a memory-intensive process so this limit can be raised if sufficient memory is available to the host.
+  Set to 0 for no limit.
--- a/pkg/restapi/backup.go
+++ b/pkg/restapi/backup.go
@@ -6,10 +6,12 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"strconv"
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/render"
 	"github.com/pkg/errors"
+	"github.com/scylladb/scylla-manager/v3/pkg/service"
 	"github.com/scylladb/scylla-manager/v3/pkg/service/backup"
 	"github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
 	"github.com/scylladb/scylla-manager/v3/pkg/service/scheduler"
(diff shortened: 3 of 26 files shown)


pkg/service/backup/config.go
```diff
@@ -14,13 +14,15 @@ import (
 type Config struct {
 	DiskSpaceFreeMinPercent   int           `yaml:"disk_space_free_min_percent"`
 	LongPollingTimeoutSeconds int           `yaml:"long_polling_timeout_seconds"`
+	ManifestLoadParallelism   int           `yaml:"manifest_load_parallelism"`
```
mmatczuk, line 17: This should be configurable per task.

pkg/service/backup/service.go
```diff
@@ -1067,28 +1068,51 @@ func (s *Service) DeleteSnapshot(ctx context.Context, clusterID uuid.UUID, locat
...
+			n, err := p.PurgeFiles(ctx, manifests, strset.New(snapshotTags...), files)
+			deletedManifests.Add(int32(n))
+			if err != nil {
+				s.logger.Error(ctx, "Purging snapshot data failed on host", "host", h.IP, "error", err)
+			} else {
+				s.logger.Info(ctx, "Done purging snapshot data on host", "host", h.IP)
+			}
+			purgeErr = multierr.Append(purgeErr, err)
+
+			wg.Done()
```
mmatczuk, line 1103: It's safer to do that under defer at the top go the func.

pkg/service/backup/service.go
```diff
@@ -1067,28 +1068,51 @@ func (s *Service) DeleteSnapshot(ctx context.Context, clusterID uuid.UUID, locat
...
+		wg.Add(1)
+		go func() {
+			n, err := p.PurgeFiles(ctx, manifests, strset.New(snapshotTags...), files)
+			deletedManifests.Add(int32(n))
+			if err != nil {
+				s.logger.Error(ctx, "Purging snapshot data failed on host", "host", h.IP, "error", err)
+			} else {
+				s.logger.Info(ctx, "Done purging snapshot data on host", "host", h.IP)
+			}
+			purgeErr = multierr.Append(purgeErr, err)
```
mmatczuk, line 1101: Is that thread safe?

mmatczuk (changes requested):
This code adds a possible go routine leak when listing of manifests fails.
We need to improve error handling.

Also, it looks like multi err is [not thread-safe](https://github.com/uber-go/multierr/blob/6fede5ccb33ebb0dbc001cfb3220566db7614acb/error.go#L225).

pkg/command/backup/res.yaml
```diff
@@ -34,3 +34,8 @@ show-tables: |
 
 purge-only: |
   Run the backup cleanup only.
+
+manifest-parallelism: |
+  The number of hosts to read manifests from in parallel.
+  The reading of manifests can be a memory-intensive process so this limit can be raised if sufficient memory is available to the host.
+  Set to 0 for no limit (default 1).
```
mmatczuk, line 41: (default 1) is added automatically

mmatczuk:
What you need here is https://pkg.go.dev/golang.org/x/sync/errgroup

mmatczuk:
I merged the commit fixing mocks and rebased.

pkg/command/backup/cmd.go
```diff
@@ -25,18 +25,19 @@ type command struct {
...
+	keyspace            []string
+	retention           int
+	retentionDays       int
+	rateLimit           []string
+	snapshotParallel    []string
+	uploadParallel      []string
+	dryRun              bool
+	showTables          bool
+	purgeOnly           bool
+	manifestParallelism int
```
mmatczuk, line 40: Naming: note snapshotParallel and uploadParallel, we should use common naming scheme.

swagger/scylla-manager.json
```diff
@@ -1601,6 +1601,12 @@
             },
             "name": "locations",
             "in": "query"
+          },
+          {
+            "type": "integer",
+            "name": "manifest_parallelism",
```
mmatczuk, line 1607: ???

pkg/restapi/backup.go
```diff
@@ -174,18 +176,40 @@ func (h backupHandler) listFiles(w http.ResponseWriter, r *http.Request) {
 	render.Respond(w, r, v)
 }
 
+func (h backupHandler) parallelism(r *http.Request) (int64, error) {
```
mmatczuk, line 179: Name should be more specific.

pkg/service/backup/service.go
```diff
@@ -1067,25 +1070,45 @@ func (s *Service) DeleteSnapshot(ctx context.Context, clusterID uuid.UUID, locat
 		return errors.Wrap(err, "resolve hosts")
 	}
 
-	deletedManifests := atomic.NewInt32(0)
-	if err := hostsInParallel(hosts, parallel.NoLimit, func(h hostInfo) error {
+	var (
+		errGroup         *errgroup.Group
```
mmatczuk, line 1074: The var entry is not needed.

mmatczuk:
With the other improvements in this area I think we should be better with adding a `purge-parallel` flag.
That is to say that it would control both loading manifests and deletes.
Should be 0 by default.

mmatczuk:
I do not like the implementation of the changed workflow.

mmatczuk:
Closing this in favor of #3171

Issue #2945, opened 2021-11-30
repair: setting small table threshold to 0 does not work on empty tables


mmatczuk:
If table size is 0 and `small-table-threshold` is set to 0 the table is still detected as a small table.
Setting small-table-threshold to 0 shall disable the feature completely.

PR #3127, opened 2022-04-21
Allow disabling Small Table Threshold optimitzation

fee-mendes:
This commit corrects the logic in repair/service.go to ensure that only tables below a given unit (and not equal) are ignored during the smallTable optimization check. This simple one-liner now allows one to effectively disable the optimization altogether by setting it as 0B, for example.


mmatczuk:
This is fixed in - https://github.com/scylladb/scylla-manager/issues/2945

PR #3131, opened 2022-04-25
service/backup: iteratively load manifest index

### LINKED ISSUE #3116: backup: use gob instead of json for manifest storage format
This needs to be analyzed first but using gob can potentially reduce CPU and memory usage due to much easier parsing.
Let's take a big manifest and try to encode it in json and gob in a benchmark to see the differences.
There is also the size component gob should be smaller.

kevinbarbour:
fixes #3116

### DIFF
--- a/pkg/util/jsonutil/seek.go
+++ b/pkg/util/jsonutil/seek.go
@@ -1,0 +1,93 @@
+// Copyright (C) 2017 ScyllaDB
+
+package jsonutil
+
+import (
+	"encoding/json"
+	"errors"
+	"io"
+	"strings"
+)
+
+// ErrKeyNotFound is an error if a key does not exist.
+var ErrKeyNotFound = errors.New("key not found")
+
+// Decoder is a wrapper around a json.Decoder.
+type Decoder struct {
+	*json.Decoder
+}
+
+// NewDecoder creates a new JSON decoder.
+func NewDecoder(r io.Reader) Decoder {
+	return Decoder{json.NewDecoder(r)}
+}
+
+// Seek seeks forward-only to a path in the JSON.
+// This should always be used when the underlying Reader is at the beginning of the root JSON object.
+// The reader must be reset to the start if seeking for multiple JSON objects.
+func (d Decoder) Seek(path string) error {
+	var (
+		err   error
+		level int
+
+		splitPath = strings.Split(path, "/")
+		depth     = len(splitPath) - 1
+	)
+
+	for t, err := d.Token(); err == nil; t, err = d.Token() {
+		s, ok := t.(string)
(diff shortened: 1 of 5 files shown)


pkg/util/jsonutil/seek.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// NewDecoder creates a new JSON decoder.
+func NewDecoder(r io.Reader) *Decoder {
+	return &Decoder{json.NewDecoder(r)}
+}
+
+// Seek seeks forward-only to a key in the JSON.
+func (d *Decoder) Seek(key string) error {
+	for t, err := d.Token(); err == nil; t, err = d.Token() {
```
mmatczuk, line 26: Nice for loop. How about handling the error returned from d.Token?

pkg/util/jsonutil/seek_test.go
```diff
@@ -0,0 +1,74 @@
...
+}
+`
+
+type Person struct {
+	Name string
+	Dogs []string
+	Cats []string
+}
+
+func TestSeekAndDecode(t *testing.T) {
```
mmatczuk, line 38: Seek is too simplistic to be called seek tbh.
It would benefit from struct awareness. 
I'd suggest it accepts super simplified jsonpath stings i.e. `store/book/author` so that if one passes `foo` it's clear you mean the top level foo.

mmatczuk:
Overall it looks OK but seek needs some hardening.

pkg/util/jsonutil/seek.go
```diff
@@ -0,0 +1,92 @@
...
+// ErrKeyNotFound is an error if a key does not exist.
+var ErrKeyNotFound = errors.New("key not found")
+
+// Decoder is a wrapper around a json.Decoder.
+type Decoder struct {
+	*json.Decoder
+}
+
+// NewDecoder creates a new JSON decoder.
+func NewDecoder(r io.Reader) *Decoder {
```
mmatczuk, line 21: Return just Decoder it's only holds a pointer

pkg/util/jsonutil/seek.go
```diff
@@ -0,0 +1,92 @@
...
+	var (
+		err   error
+		level int
+
+		splitPath = strings.Split(path, "/")
+		depth     = len(splitPath) - 1
+	)
+
+	for t, err := d.Token(); err == nil; t, err = d.Token() {
+		if s, ok := t.(string); ok { //nolint: nestif
```
mmatczuk, line 40: Maybe handle that first with continue?

mmatczuk: Or use fmt.Sprint?

PR #3145, opened 2022-05-02
Full range repair

### LINKED ISSUE #3128: repair: join ranges in small table optimization
When dealing with small tables we would send all ranges for given replicas in one Scylla job.
A special case is when data is fully replicated data `RF=nr. nodes`.
In such a case it enough to repair a full range, example: `nodetool repair -st -9209318497534407381  -et 9190878967713862152 system_auth roles`.

A) in case of small tables maybe it's always better to just do a full repair like in the example above.
B) in case of fully replicated data it is always better to send the full range

kevinbarbour:
service/repair: join ranges when fully replicated small table

Run repairs on full ranges when tables are below small table threshold
and fully replicated.

fixes #3128

### DIFF
--- a/pkg/service/cluster/service_integration_test.go
+++ b/pkg/service/cluster/service_integration_test.go
@@ -30,7 +30,6 @@ func TestServiceStorageIntegration(t *testing.T) {
 	session := CreateSession(t)
 
 	secretsStore := store.NewTableStore(session, table.Secrets)
-	// TimeoutConfig: scyllaclient.DefaultTimeoutConfig(),
 
 	s, err := cluster.NewService(session, metrics.NewClusterMetrics(), secretsStore, scyllaclient.DefaultTimeoutConfig(), log.NewDevelopment())
 	if err != nil {
--- a/pkg/service/repair/range.go
+++ b/pkg/service/repair/range.go
@@ -14,12 +14,13 @@ import (
 )
 
 type tableTokenRange struct {
-	Keyspace   string
-	Table      string
-	Pos        int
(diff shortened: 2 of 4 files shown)


mmatczuk:
Consider this code we use to init the generator.
Could we change the newTableTokenRangeBuilder to detect the fully replicated table and just add a single entry to gen?

```
		// Get ring
		ring, err := client.DescribeRing(ctx, u.Keyspace)
		if err != nil {
			return errors.Wrapf(err, "keyspace %s: get ring description", u.Keyspace)
		}

		// Transform ring to tableTokenRanges
		b := newTableTokenRangeBuilder(target, ring.HostDC)
		b.Add(ring.Tokens)

		// Calculate worker count
		if v := b.MaxParallelRepairs(); v > maxParallel {
			maxParallel = v
		}

		// Add token ranges to generator
		gen.Add(ctx, b.Build(u))
```

kevinbarbour:
We can but we would need also need to change when/how we do the small table optimisation because at that point we haven't yet marked the small tables in the generator. We only want to add a single entry if the table is small right? Not for every fully-replicated table?

mmatczuk:
I'd prefer a post processing in gen approach.
After running STO that marks tables as small in `gen` you would run a function that:
* calculates hash of all nodes `h := replicaHash(g.Hosts().List())`
* iterates over ranges for that hash and removes the small tables and replaces them with full range TTR

Make sure you do that before gen.Init so that progress is setup correctly.

mmatczuk (approved):

PR #3171, opened 2022-06-01
service/backup: use jsoniter to stream manifest indexes

### LINKED ISSUE #3169: Progress missing IP info during backup purge
Introduced by 7c67bc8a43dd1a01718d4aae60d3c3698c9c5ebc

In an effort to reduce memory usage we are only loading the Indexes iteratively from manifests during a purge. Unfortunately the progress details use some of this other metadata (specifically IP address) so this causes that to break. Need to change this so we no longer skip that data, but still load it iteratively.

kevinbarbour:
When adding streaming of manifest indexes to the purge process, we
stopped loading in the metadata for manifest content which is used in
recording progress.

This replaces the seek and decode functionality added to jsonutil with
the [json-iterator package](https://github.com/json-iterator/go). This allows us to easily decode the needed JSON
contents but still stream the indexes and retain the memory footprint
improvements.

Revert "util/jsonutil: add decoder wrapper with seeker"

This reverts commit 205788b2513cc47e869b1ead3c79f217332c0194.

fixes #3169


mmatczuk:
Can you separate vendor changes to a separate commit?

mmatczuk:
This work touches an important aspect that `func (m *ManifestContent) Read(r io.Reader)` still exists and is used in the listing. I have a proposal that uses 2 pass reading to separate metadata reading from iteration over table files.

`ManifestContent.Read`
- creates a temp file
- tees compressed payload to the file
- has a finalizer that removes the file
- reads the metadata

`ManifestContent.Index []FilesMeta `json:"index"`
- is replaced with 
  - Index() []FilesMeta function that returns FilesMeta without files
  - ForEachIndexIter

This can use the existing implementation and also addresses all ManifestReading scenarios.

WDYT @kevinbarbour ?

kevinbarbour:
@mmatczuk I like it, except I'm not sure about the reliance on the finalizer to remove the files. It seems like a way to potentially end up with a lot of orphaned files if the manager is restarted/aborted before GC has run, or a lot of space taken up in intermediate time between GC runs.

The rest of the implementation sounds good though - I will do some testing with it cleaning up files with the finalizer to see if it leads to any issues.

mmatczuk (approved):

PR #3175, opened 2022-06-03
2.4 backport check location fixes

mmatczuk:


mmatczuk:
The failure is not relevant to the change.

PR #3196, opened 2022-07-12
generic scheduler

### LINKED ISSUE #3117: scheduler: use generics
Both [Key and Properties](https://github.com/scylladb/scylla-manager/blob/660630e9125e5c712f336e04bd17d6b2faf41d7d/pkg/scheduler/scheduler.go#L16) were designed to be replaced by a generic type.

Michal-Leszczynski:
Replaced scheduler `type Key = uuid.UUID` with generic type.
Replaced scheduler `type Properties = json.RawMessage` with `any interface`.

Fixes #3117


mmatczuk:
Also change go version in [.go-version](https://github.com/scylladb/scylla-manager/blob/master/.go-version) file.

mmatczuk:
@Michal-Leszczynski can you organize the commits better.

Michal-Leszczynski:
> TestValidateIntegration
> TestPurgeTemporaryManifestsIntegration

Those two can randomly fail on my pc when running all integration tests (they pass when I run them separately).

> TestServiceRepairIntegration / repair temporary network outage

This test always fails both on this branch and on master on my pc.

PR #3226, opened 2022-09-06
Universal Restore

### LINKED ISSUE #3170: backup: add universal restore
Design doc https://docs.google.com/document/d/1ugf6XCs28c4pwKsw9Lbu_EFb6ZSgMpJ_-Usf-Q9XAP4/edit

Michal-Leszczynski:
This PR implements:
- restore table contents
- restore schema
- restore dry run
- restore progress

Current implementation displays information about the need to run a follow-up action (repair/restart) after restore completion in `sctool tasks` and `sctool progress` commands.

Fixes #3170


karol-kokoszka (commented):
Review is still in progress....
I paste the comments I already have just to .. not loose them.

karol-kokoszka:
@Michal-Leszczynski please add the information that PR closes universal restore issue. You can do it by simply writing `Fixes #3170` somewhere in the PR's description.

pkg/metrics/restore.go
```diff
@@ -0,0 +1,21 @@
...
+	l := prometheus.Labels{
+		"cluster":  clusterID.String(),
+		"keyspace": keyspace,
+		"table":    table,
+		"host":     host,
+	}
+	m.filesSizeBytes.With(l).Add(float64(size))
+	m.filesUploadedBytes.With(l).Add(float64(uploaded))
+	m.filesSkippedBytes.With(l).Add(float64(skipped))
+	m.filesFailedBytes.With(l).Add(float64(failed))
```
karol-kokoszka, line 20: These are backup metrics with the description e.g. `Number of bytes uploaded to backup location.`
There must be separated set of metrics for restore.

karol-kokoszka (commented):
another set of comments
still in progress

pkg/config/server/server.go
```diff
@@ -89,7 +89,12 @@ func DefaultConfig() Config {
 // config struct with merged configuration from all provided files.
 func ParseConfigFiles(files []string) (Config, error) {
 	c := DefaultConfig()
-	return c, cfgutil.ParseYAML(&c, DefaultConfig(), files...)
+	err := cfgutil.ParseYAML(&c, DefaultConfig(), files...)
+	if err == nil && c.Backup.LocalDC == "" {
+		c.Backup.LocalDC = c.Database.LocalDC
```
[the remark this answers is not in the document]
Michal-Leszczynski: I wanted to add `localDC` field (that already exists in database config) to backup config. 
I figured out that it might cause an inconvenience for the user to specify `localDC` in both backup and database `yaml` files, so if possible, I wanted to copy `localDC` from database to backup config.

Maybe I should just take away `'yaml:"local_dc"'` tag from backup config and copy it blindly from database config?

pkg/service/backup/restore_worker.go
```diff
@@ -0,0 +1,39 @@
...
+	Shards     uint
+	OngoingJob *RestoreRunProgress
+}
+
+// bundle represents SSTables with the same ID.
+type bundle []string
+
+// restoreWorker extends implementation of backup worker.
+type restoreWorker struct {
+	worker
```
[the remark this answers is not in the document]
Michal-Leszczynski: After checking, I only need 3 relatively small methods from worker that requires two worker fields at best. What if I could extract them to regular functions and pass those fields as arguments? 
The only problem is that I would have to pass logger as an argument and I'm not sure if it's pretty. Maybe passing logger covered in function or interface would be better?

pkg/service/backup/restore_worker_schema.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := w.SetGraceSecondsAndCompaction(ctx, keyspace, table, maxGGS, tmpComp); err != nil {
+		return err
+	}
+	// Reset gc grace seconds and compaction
+	defer w.SetGraceSecondsAndCompaction(ctx, keyspace, table, ggs, comp) //nolint: errcheck
+
+	return f()
+}
+
+func (w *restoreWorker) RecordTableVersion(ctx context.Context, keyspace, table string) (string, error) {
```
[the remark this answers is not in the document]
Michal-Leszczynski: Kevin started this name convention and I kept on going with it, but I will change that for it to be simpler.

pkg/service/backup/restore_worker_data.go
```diff
@@ -0,0 +1,544 @@
...
+				return errors.Wrap(err, "call agent's restore")
+			}
+
+			w.Logger.Info(ctx, "Restored batch",
+				"host", h.Host,
+				"batch", batch,
+			)
+
+			// Close pool and free hosts awaiting on it if all SSTables have been successfully restored.
+			if ctr.RestoreTables(len(pr.SSTableID)) {
```
karol-kokoszka, line 220: nit: `RestoreTables` is not descriptive enough.
It decreases the counter and checks if counter == 0.
Maybe the previous approach without this method, but the check maybe explicitly here was cleaner.

etc/scylla-manager/scylla-manager.yaml
```diff
@@ -0,0 +1,157 @@
+# Scylla Manager config YAML
+
```
karol-kokoszka, line 2: I think this file is not needed ?
`/etc/scylla-manager/scylla-manager.yaml` is the configuration of scylla-manager server working on host.
Is it some leftover ?

pkg/service/backup/restore.go
```diff
@@ -72,8 +72,97 @@ func (s *Service) GetRestoreTarget(ctx context.Context, clusterID uuid.UUID, pro
...
+		if err != nil {
+			return errors.Wrap(err, "initialize units")
+		}
+	}
+
+	run.Stage = StageRestoreData
+	w.InsertRun(ctx, run)
+
+	if err := w.restoreData(ctx, run, target); err != nil {
+		return errors.Wrapf(err, "restore data")
```
karol-kokoszka, line 159: In what state the ScyllaDB is left if an error during restore appears ?
Is it something that can be reversed ?
Just wondering whether there should be some more descriptive information on stdout when an error appeard. I mean what to do to fix the state.

swagger/gen/scylla-manager/client/operations/get_cluster_cluster_id_tasks_restore_target_parameters.go
```diff
@@ -0,0 +1,153 @@
+// Code generated by go-swagger; DO NOT EDIT.
```
karol-kokoszka, line 1: commit reladed.
Could be merged with the previous commit.

pkg/command/restore/res.yaml
```diff
@@ -0,0 +1,39 @@
...
+  Increasing batch size increases job granularity.
+
+parallel: |
+  The maximum number of Scylla restore jobs that can be run at the same time (on different SSTables).
+  Each node can take part in at most one restore at any given moment.
+
+restore-schema: |
+  Specifies restore type (alternative to '--restore-tables' flag).
+  Restore will recreate schema by targeting 'system_schema.*' tables only ('--keyspace' flag mustn't be set).
+  It requires the cluster to be in initial state (e.g. for each node: stop scylla, remove contents of
```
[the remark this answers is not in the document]
karol-kokoszka: I think it would be good to reference the initial state (link to documentation for example) or describe what is the initial state.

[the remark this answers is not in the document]
Michal-Leszczynski: After running some restore schema tests I came to the conclusion that the requirement should be that none of the restored keyspaces is present in the destination cluster (they can be dropped and restored). Of course this is only empiric observation, so what should be written in the documentation?

pkg/service/backup/restore_worker.go
```diff
@@ -37,3 +42,106 @@ type restoreWorker struct {
...
+				"run_progress", *pr,
+				"error", err,
+			)
+		}
+	})
+}
+
+// GetRun returns run with specified cluster, task and run ID.
+// If run ID is not specified, it returns latest run with specified cluster and task ID.
+func (w *restoreWorker) GetRun(ctx context.Context, clusterID, taskID, runID uuid.UUID) (*RestoreRun, error) {
```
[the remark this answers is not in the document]
Michal-Leszczynski: When I was browsing through repository, I found that other packages (like repair and backup) export this method.
Should we export it here for the sake of consistency?

pkg/cmd/agent/restore.go
```diff
@@ -0,0 +1,173 @@
...
+	if err := h.callLoadAndStream(ctx, params.Keyspace, params.Table); err != nil {
+		h.respondWithInternalError(ctx, w, r, errors.Wrap(err, "call load and stream"))
+
+		return
+	}
+
+	render.Status(r, http.StatusOK)
+	render.Respond(w, r, "success")
+
+	h.logger.Info(ctx, "Restore ended")
```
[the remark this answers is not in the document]
Michal-Leszczynski: Yes, that's basically all that agent has to do after the data has been already downloaded to it.

pkg/cmd/agent/restore.go
```diff
@@ -0,0 +1,173 @@
...
+		h.respondWithInternalError(ctx, w, r, errors.Wrap(err, "get upload directory absolute path"))
+
+		return
+	}
+
+	h.logger.Info(ctx, "Upload data directory",
+		"dir", absUploadDir,
+	)
+
+	//TODO - how to do this ownership in production? Should the manager be running as scylla user?
```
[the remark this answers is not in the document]
Michal-Leszczynski: Scylla needs to be the owner of the downloaded files in order to use them in load and stream.

pkg/service/backup/restore.go
Michal-Leszczynski: In case of error, restore should behave in the same way as in case of context cancellation (so a safe state that has some already restored data and can be continued - at least that's the goal). Problem occurs if error is returned from resetting gc_grace_seconds and compaction on the table, because then table will stay in this disabled state. For this scenario I can add more descriptive error message which will include original values of gc_grace_seconds and compaction and ask user to reset them manually.

But that's from scylla-manager perspective.

I don't know how rclone handles errors (does it actually cleans already downloaded data or does it leave it there?).
The same goes for error during load and stream procedure (does it clean upload directory, does it clean already uploaded data to Scylla?)

What do you think about that?

pkg/rclone/rcserver/rc.go
```diff
@@ -616,6 +660,20 @@ func init() {
...
 	})
+
+	rc.Add(rc.Call{
+		Path:         "sync/copypaths",
+		AuthRequired: true,
+		Fn:           wrap(rcCopyPaths(), remoteToLocal()),
+		Title:        "Copy paths from source directory to destination",
+		Help: `This takes the following parameters:
+
+- srcFs - a remote name string eg "drive:" for the source
```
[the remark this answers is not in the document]
Michal-Leszczynski: From my pov `s3` is more informative than drive.

> - fs - a remote name string eg "`s3`:path/to/file"

Some other descriptions already have `s3` instead of `drive` in them, so I will `drive` to `s3` in the whole file.

pkg/cmd/agent/restore.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		h.logger = &l
+	}
+
+	h.logger.Info(ctx, "Restoring files")
+
+	params := new(models.RestoreParams)
+	dec := json.NewDecoder(r.Body)
+	err := dec.Decode(params)
```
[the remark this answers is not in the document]
Michal-Leszczynski: There is no validation included right now.
Params are: ` keyspace, table, version string, files []string`

I could add some regex validation.
Table version in `system_schema.tables` is kept with  **'-'** characters, but it has to be stripped of them in here.
Should I add it?

pkg/command/restore/res.yaml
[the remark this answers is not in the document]
karol-kokoszka: Yes, please.

pkg/service/backup/restore_worker.go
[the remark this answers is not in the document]
karol-kokoszka: I saw it's refactored already. Looks much better!
What is worth to do, is to split Metrics as well.
The ones that are changed by common `worker` struct, may stay in `BackupMetrics` = `Metrics`.
Other ones should be part of process specific Metrics struct.
Another option is to just change name `BackupMetrics` to `Metrics`.

pkg/service/backup/backupspec/manifest.go
```diff
@@ -219,28 +221,44 @@ func (m *ManifestContentWithIndex) ForEachIndexIter(cb func(fm FilesMeta)) error
 		return err
 	}
 
+	filter, err := ksfilter.NewFilter(keyspace)
```
karol-kokoszka, line 224: defer f.Close() (couple of lines above) doesn't log the potential error. 
I know it's not introduced in your PR, but lacking at least some info message when the file couldn't be closed makes tracing sometimes weird production bugs easier.
Appreciate if you can add it.

pkg/service/backup/restore_worker_schema.go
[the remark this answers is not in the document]
karol-kokoszka: np, thanks for changing it

pkg/cmd/agent/restore.go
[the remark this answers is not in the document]
karol-kokoszka: Yes, please add some basic validation.
What do you think about just logging with ERROR if some of the required parameters are missing and returning HTTP 400 (Bad Request) ?

pkg/cmd/agent/restore.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, f := range params.Files {
+		h.logger.Info(ctx, "Restore file",
+			"file", f,
+		)
+
+		err = os.Chown(
+			path.Join(absUploadDir, f),
+			107,
+			109,
+		)
```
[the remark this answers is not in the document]
karol-kokoszka: What about using using chmod and setting read (or write) permissions to others with https://pkg.go.dev/os#Chmod ?
I think it's much easier than trying to find the uid and gid of scylla user in system and changing the owner.
The only concern about chmod with +r is the security. I'm not an expert there, but I think if it's in the scylla owned directory then no other user will see the file anyway.

pkg/cmd/agent/restore.go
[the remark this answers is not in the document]
karol-kokoszka: See the previous comment on chmod

pkg/service/backup/restore.go
karol-kokoszka: > Problem occurs if error is returned from resetting gc_grace_seconds and compaction on the table, because then table will stay in this disabled state. For this scenario I can add more descriptive error message which will include original values of gc_grace_seconds and compaction and ask user to reset them manually.

Yes, please.

> I don't know how rclone handles errors (does it actually cleans already downloaded data or does it leave it there?).
The same goes for error during load and stream procedure (does it clean upload directory, does it clean already uploaded data to Scylla?)

To be asked on the planning.

pkg/cmd/agent/restore.go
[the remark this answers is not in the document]
Michal-Leszczynski: Unfortunately even with `os.Chomd("file", 0777)` the same error is returned - Scylla checks for file ownership (probably because it wants to delete the file after issuing load and stream).

pkg/config/server/server.go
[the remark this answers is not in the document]
Michal-Leszczynski: As discussed during the meeting, localDC is no longer needed for restore purposes, so I am dropping this commit.

pkg/cmd/agent/restore.go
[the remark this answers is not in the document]
Michal-Leszczynski: Since the ownership problem occurs only with Scylla 4.6.1 (or older) that does not have to be supported by the restore, we shouldn't change ownership of the downloaded files. Having said that, we should bump current Scylla version used for testing.
I tried to bump Scylla version to 5.x, but currently it's not possible because of #3235, so for now I will bump it to 4.6.7.

pkg/config/server/server.go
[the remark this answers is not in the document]
karol-kokoszka: I guess it's not commit yet. I'll resolve the conversation when it's done.

pkg/cmd/agent/restore.go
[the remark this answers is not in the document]
Michal-Leszczynski: Since agent is only supposed to call load and stream I deleted agent restore endpoint and now I use Scylla's API directly.

karol-kokoszka (approved):
:+1: Great work !

Michal-Leszczynski:
Changes introduced in commits that need to be squashed:
- improved restore parameters validation and logging
- In case location's dc is not specified, restore uses all nodes (to remain compatible with SimpleStrategy)
- added links for cluster restart and repair docs to restore command description
- added restore validation so that it is not scheduled in case of initial error
- fixed resume procedure (rclone jobs cannot be resumed but they can be restarted)
- added tests for resume, agent down, agent restart

pkg/service/backup/restore_worker_data.go
```diff
@@ -25,22 +25,30 @@ func (w *restoreWorker) restoreData(ctx context.Context, run *RestoreRun, target
...
+			defer w.Logger.Info(ctx, "Restoring manifest finished", "manifest", miwc.ManifestInfo)
+
+			w.miwc = miwc
 			run.ManifestPath = miwc.Path()
 
 			return miwc.ForEachIndexIterWithError(target.Keyspace, w.restoreFiles(ctx, run, target))
 		}
 
 		if err := w.forEachRestoredManifest(ctx, w.location, manifestHandler); err != nil {
+			w.Logger.Info(ctx, "Restoring location finished", "location", w.location)
```
karol-kokoszka, line 43: Why to log this Info message on error ?
It may be misleading. Error will be logged at the top of invocation tree anyway.

pkg/service/backup/restore_worker_data.go
```diff
@@ -25,22 +25,30 @@ func (w *restoreWorker) restoreData(ctx context.Context, run *RestoreRun, target
...
 			return err
 		}
+
+		w.Logger.Info(ctx, "Restoring location finished", "location", w.location)
```
karol-kokoszka, line 47: Maybe `Restoring location completed` would be better ?
Completed means fulfilled, finished means stopped with no matter what result.

https://ejoy-english.com/blog/complete-vs-finish-similar-but-different/#:~:text='To%20complete%20something'%20means%20to,something%20means%20to%20end%20it.

pkg/service/backup/restore_worker_data.go
```diff
@@ -0,0 +1,667 @@
...
+	// All restore hosts are trying to get IDs for batch from the pool.
+	// Pool is closed after the whole table has been restored.
+
+	// Take at most batchSize IDs
+	for i := 0; i < batchSize; i++ {
+		select {
+		case id, ok := <-w.bundleIDPool:
+			if !ok {
+				return takenIDs
+			}
```
karol-kokoszka, line 517: I'm trying to understand the logic here.
`!ok` will be true only if `w.bundleIDPool` channel is closed.
Channel is being closed only when there is no more tables to restore.

What do you want to achieve here ? What this check suppose to do ?

pkg/service/backup/restore_worker_data.go
```diff
@@ -0,0 +1,667 @@
...
+			takenIDs = append(takenIDs, id)
+			if ctx.Err() != nil {
```
karol-kokoszka, line 519: Why ?

Do you want to handle the situation when the context is cancelled ?
If so, then take a look on this snippet https://go.dev/play/p/rGEzKDEkaXr

It is enough to add `case <- ctx.Done()` as the first check of select.
When the context is cancelled, then the ctx.Done() channel is closed what directly means that `case <-ctx.Done()` is always chosen from select if it's at the first place.

pkg/service/backup/restore_worker_data.go
```diff
@@ -0,0 +1,667 @@
...
+				return takenIDs
+			}
+		default:
```
karol-kokoszka, line 522: In my understanding, this `default:` is chosen when there is nothing available in ` <-w.bundleIDPool` channel.

Shouldn't this `default:` clause just return `takenIDs` ?

pkg/service/backup/restore_worker_data.go
```diff
@@ -0,0 +1,667 @@
...
+			// Don't wait for more IDs if the pool is empty
+			// and host already has something to restore.
+			if len(takenIDs) > 0 {
+				return takenIDs
+			}
+			// Here host hasn't taken any IDs and pool is empty,
+			// so it waits for the whole table to be restored or
+			// for IDs that might return to the pool in case of error on the other hosts.
```
karol-kokoszka, line 530: You wrote that 
``` or for IDs that might return to the pool in case of error on the other hosts```

Is it a must to pickup from the pool IDs that went back to the pool in case of an error ?
I guess not. Restore may end with error and we accept it (as long as it can be fixed with another run).

Let's not do it. It complicates the code much. Besides, we still don't guarantee that all IDs that went back to the pool will be processed again.

File: pkg/service/backup/restore_model.go [hunk 23]
```diff
@@ -108,6 +108,10 @@ func (rp *RestoreRunProgress) setRestoreCompletedAt() {
 	rp.RestoreCompletedAt = &t
 }
 
+func validateTime(t *time.Time) bool {
```
karol-kokoszka, line 111: `validateTime` against what ?
I see it just checks if time is set.
I propose to change the method signature to include this information.
E.g. **validateTimeIsSet**

pkg/service/backup/restore_worker_schema.go
```diff
@@ -13,6 +13,28 @@ import (
 	"go.uber.org/multierr"
 )
 
+func (w *restoreWorker) ValidateTableExistsAndEmpty(ctx context.Context, keyspace, table string) error {
```
karol-kokoszka: nit: ` ValidateTableExistsAndEmpty` -> ` ValidateTableExistsAndIsEmpty`

pkg/scyllaclient/client_ping.go
```diff
@@ -32,7 +33,7 @@ func (c *Client) GetLiveNodesWithLocationAccess(ctx context.Context, nodes NodeS
 	err = parallel.Run(len(liveNodes), parallel.NoLimit, func(i int) error {
 		n := liveNodes[i]
 
-		if err := c.RcloneCheckPermissions(ctx, n.Addr, remotePath); err == nil {
+		if err = c.RcloneCheckPermissions(ctx, n.Addr, remotePath); err == nil {
```
karol-kokoszka, line 36: Why this change ? `err:=` defines scope of current `err` variable to the method where it's created.

pkg/service/backup/service_backup_restore_integration_test.go
```diff
@@ -420,29 +420,29 @@ func TestRestoreTablesNodeDownIntegration(t *testing.T) {
...
+	//			Path:     testBucket,
+	//		},
+	//	},
+	//	Keyspace:      []string{"system_schema"},
+	//	BatchSize:     testBatchSize,
+	//	Parallel:      testParallel,
+	//	RestoreSchema: true,
+	//}
+	//
+	//restoreWithNodeDown(t, target, testKeyspace, testBackupSize)
```
karol-kokoszka, line 445: Why the test is commented out ? Not needed or work in progress ?

pkg/service/backup/service_backup_restore_integration_test.go
```diff
@@ -510,29 +510,29 @@ func TestRestoreTablesRestartAgentsIntegration(t *testing.T) {
...
+	//			Path:     testBucket,
+	//		},
+	//	},
+	//	Keyspace:      []string{"system_schema"},
+	//	BatchSize:     testBatchSize,
+	//	Parallel:      testParallel,
+	//	RestoreSchema: true,
+	//}
+	//
+	//restoreWithAgentRestart(t, target, testKeyspace, testBackupSize)
```
karol-kokoszka, line 535: why commented out ?

karol-kokoszka:
@Michal-Leszczynski Couple of minor comments. The ones related to `restore_worker_data` are something what we could discuss bit deeper.
Kudos for adding more integration-tests, but please make sure they are **passing** (See `Sanity check/Integration tests` status)

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: Even when some host does not have any SSTable ID assigned for its batch (currently channel is empty and he hasn't taken any IDs yet), he still waits on the channel with SSTable IDs, because there is a chance (because of error on some host) that IDs will return to the pool. If this happens, this host will be able to take returned IDs and proceed with restore. That's why I check if channel is closed, because, as you mentioned, it means that all SSTables have been successfully restored and only after that I let this host exit this loop.

Michal-Leszczynski: Although I could replace takenIDs with nil, since in this case takenIDs is always nil.

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: > When the context is cancelled, then the ctx.Done() channel is closed what directly means that case <-ctx.Done() is always chosen from select if it's at the first place.

Are you sure that selects work like that in go? Take a look on this snippet https://go.dev/play/p/FtFclYmptAx
Here all channels have values inside of them, but select isn't always choosing the first case - it does that randomly.

Michal-Leszczynski: In my code I wanted to make sure that returned takenIDs are always accurate - even in case of context cancellation. That's why even if I take sth out of the channel when context is cancelled, I still append it to the takenIDs so that it does not get lost in the process.

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: This function is supposed to be a waiting place for host that couldn't take any IDs from the pool - but are still valid hosts that might be useful for restore procedure if some IDs return to the pool (because of an error on other host).
Later in the code, the return value of this function is checked and if it returned 0 IDs, then host ends it's restore loop thinking that there is no more data to be restored, so that's why it shouldn't return nil if this host might be useful in the near future.

pkg/service/backup/service_backup_restore_integration_test.go
Michal-Leszczynski: It works fine but I had some problems with restore integration tests running on github - I assume that was because of the amount of space and time required for those tests. I changed the way in which backups to be restored are generated and now it passes the tests.
Commenting out this test was an attempt to see if reducing the amount of restore tests would help in any way.

pkg/service/backup/service_backup_restore_integration_test.go
Michal-Leszczynski: Same as above.

pkg/scyllaclient/client_ping.go
Michal-Leszczynski: I changed that because after the if I am still using this error in `nodeErr[i] = err`, but i guess that I should just assign the result of this operation to error outside of if statement, so it looks cleaner.

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: But isn't it useful to have this kind of fallback implemented (especially that it's already there)? As far as I'm concerned, restore can be a really expensive operation (downloading data from backup location and using load and stream mechanism), so maybe it makes sense to try to execute it even if one node is going through some unexpected errors?

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: I wrote that in this way so that it's consistent with logging information about table/manifest restore finished. In those two cases, I am using defer without any error checking, so it logs the information regardless of whether the restore finished with success or error (that's also the reason why I decided to use word "finished", so that it does not implicate success). 
I could add some error checking to those defers which would log either restore location/manifest/table completed/failed
and apply the same logic here. What do you think about that?

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: Answered to that in above comment.

pkg/service/backup/restore_worker_data.go
karol-kokoszka: Thanks for the explanation, it's fine.

pkg/service/backup/restore_worker_data.go
karol-kokoszka: Ok

pkg/service/backup/restore_worker_data.go
karol-kokoszka: It's ok as it is, as per our video-sync.

pkg/service/backup/restore_worker_data.go
```diff
@@ -0,0 +1,732 @@
...
+	pr.Failed = job.Failed
+
+	w.metrics.UpdateFilesProgress(w.ClusterID, pr.ManifestPath, pr.Keyspace, pr.Table,
+		deltaDownloaded, deltaSkipped, deltaFailed)
+
+	w.insertRunProgress(ctx, pr)
+}
+
+func (w *restoreWorker) restoreSSTables(ctx context.Context, host, keyspace, tables string) error {
+	// TODO: how to set the following variables?
```
karol-kokoszka, line 611: I understand you want to use retry..backoff mechanism, right ?
The question is related to how to achieve the possibility of setting / overriding these default values you set here ?

If so, then there is some work done on `do_not_run_backup_or_repair_when_load_is_high` branch.
I would say it's ok, to commit it as is. (Remove TODO, and apply the fix when the branch I mentioned is merged)

pkg/scyllaclient/client_scylla.go
```diff
@@ -823,3 +823,57 @@ func (c *Client) TableDiskSizeReport(ctx context.Context, hostKeyspaceTables Hos
 
 	return report, err
 }
+
+// LoadSSTables that are already downloaded to host's table upload directory.
+func (c *Client) LoadSSTables(ctx context.Context, host, keyspace, table string, loadAndStream, primaryReplicaOnly bool) error {
+	const restoreTimeout = 15 * time.Minute // TODO: should we even set that? To what value?
```
karol-kokoszka, line 829: Good point.
What you do here is just setting timeout on API call. It protect the request from hanging forever.
According to https://github.com/scylladb/scylla-manager/blob/58f469cdb00b24d0c30c3c509f9a1cc33dc54042/pkg/service/backup/restore_worker_data.go#L637
this API call is async (or it behaves this way).

I would say there is completely no need of setting this value to 15 minutes.
I would even say value of this timeout is not so important. 

I propose to lower the value to 1m, 15 minutes is not harmful still tough.

Michal-Leszczynski: Calling load_and_stream when no load_and_stream is running is sync - response will not be sent until load_and_stream finishes.
Calling load_and_stream when other load_and_stream is running is async - we immediately get error response that says, that load_and_stream is already running.
That's why I am hesitating on this timeout value, but as you said, for now it should be left for QA team to decide what to do with the timeout.

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: Same as comment below.

pkg/scyllaclient/client_scylla.go
karol-kokoszka: OK, as it's confirmed that load&stream scylla API call is sync.

PR #3231, opened 2022-09-12
testing: makefile, docker-compose, changed UID to CURRENT_UID

Michal-Leszczynski:
UID is a readonly system variable, so it can't be assigned and passed to docker-compose file


karol-kokoszka (approved):
:+1:

PR #3232, opened 2022-09-13
service/backup: forEachManifest no longer loads Index into memory

Michal-Leszczynski:
Iterating over Index should be done using ForEachIndexIter or ForEachIndexIterFiles.

Commit 4dc88bebb68c549a010ff3e66fa7891fa0c39346 introduced a new way of processing Index file that consumes less memory. Normal iteration over Index has been replaced with either ForEachIndexIter or ForEachIndexFiles (especially in all uses of ForEachManifest method).

Unfortunately, ForEachManifest hasn't been changed so it still loads the whole index to memory, only for ForEachIndexIter to load it again (I guess someone forgot to delete this part).


karol-kokoszka (approved):
:+1:

PR #3233, opened 2022-09-14
Ml/fix backup list filter bug

Michal-Leszczynski:
Commit 4dc88be introducead a bug in which 'backup list' command ignored '-K' flag.
This PR adds additional tests for 'backup list' and fixes this issue.


karol-kokoszka (approved):
:+1:

PR #3238, opened 2022-09-29
util: Add support for backoff dependency version 4.1.3

YvanDaSilva:
Update the "backoff" dependency.

Version 4.1.3 brings fixes for context cancelling.
https://github.com/cenkalti/backoff/commit/1357787b952893cff77d8349aaf320c3fd327c2a

However, this "patch" version change does bring an API breakage as .Permanent is now returning a go std error instead of a Permanent.Error from the backoff module.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/aws/aws-sdk-go v1.35.17
-	github.com/cenkalti/backoff/v4 v4.0.0
+	github.com/cenkalti/backoff/v4 v4.1.3
 	github.com/cespare/xxhash/v2 v2.1.1
 	github.com/go-chi/chi/v5 v5.0.0
 	github.com/go-chi/render v1.0.0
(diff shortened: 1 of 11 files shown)


karol-kokoszka:
@YvanDaSilva could you please remind why you needed this branch ?
Does backoff 4.1.3 contains any breaking change in comparison to the 4.0.0 that manager is using right now ?
I just wonder if it could be just merged to master.

YvanDaSilva:
@karol-kokoszka backoff 4.1.3 contains an API breaking change.

The API is now returning a go error instead of a customer backoff one (Permanent.Error).

This branch is what the siren / bo are actually targeting ( github.com/scylladb/scylla-manager/v3 v3.0.0-20220929135250-b593dfbd3239 ) 

If you merge to master, we will be able to target your latest release instead.
If you do so, leave the branch alive until we change it on our side.

karol-kokoszka:
Closing the PR. Siren is not referencing this commit.

PR #3245, opened 2022-10-17
Do not run backup or repair when load is high

### LINKED ISSUE #3181: Do not run backup or repair when load is high
The idea is to prevent running backup on occasional load spike. It's a lightweight version of #3180. 
#Related is https://github.com/scylladb/scylla-enterprise/issues/2268.

The implementation idea:
- Add configuration options in the configuration file that would allow to specify max load when starting backup and repair
- On start check relevant node metrics and validate them against the the threshold
- If nok return an error, scheduler would backoff and retry

This could be also helpful with https://github.com/scylladb/scylla-enterprise/issues/2196 @AdamStawarz

karol-kokoszka:
Closes #3181 

To be discussed:
- if prometheus is down, or if someone messed up with `scylla_reactor_utilization` metrics then neither repair nor backup will be performed.

UPDATE:
If prometheus is down, check is not going to be executed and backup/repair run as is.

---

Please make sure that:

### DIFF
--- a/pkg/service/backup/config.go
+++ b/pkg/service/backup/config.go
@@ -6,8 +6,9 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/scylladb/scylla-manager/v3/pkg/service"
 	"go.uber.org/multierr"
+
+	"github.com/scylladb/scylla-manager/v3/pkg/service"
 )
 
 // Config specifies the backup service configuration.
@@ -15,6 +16,7 @@ type Config struct {
 	DiskSpaceFreeMinPercent   int           `yaml:"disk_space_free_min_percent"`
 	LongPollingTimeoutSeconds int           `yaml:"long_polling_timeout_seconds"`
 	AgeMax                    time.Duration `yaml:"age_max"`
+	MaxLoadThreshold          float64       `yaml:"max_load_threshold"`
 }
 
 func DefaultConfig() Config {
@@ -22,6 +24,7 @@ func DefaultConfig() Config {
 		DiskSpaceFreeMinPercent:   10,
 		LongPollingTimeoutSeconds: 10,
 		AgeMax:                    12 * time.Hour,
+		MaxLoadThreshold:          100,
 	}
 }
 
@@ -36,6 +39,9 @@ func (c *Config) Validate() error {
 	}
 	if c.AgeMax < 0 {
 		err = multierr.Append(err, errors.New("invalid age_max, must be >= 0"))
+	}
+	if c.MaxLoadThreshold < 0 {
+		err = multierr.Append(err, errors.New("invalid max_load_threshold, must be >= 0"))
 	}
 
 	return err
--- a/pkg/service/backup/service.go
+++ b/pkg/service/backup/service.go
@@ -780,7 +780,12 @@ func (s *Service) Backup(ctx context.Context, clusterID, taskID, runID uuid.UUID
 	}
 	stageFunc := map[Stage]func() error{
 		StageClusterCheck: func() error {
-			return w.CheckCluster(ctx, hi, s.backoffConfiguration[StageClusterCheck])
+			err := w.CheckCluster(ctx, hi, s.backoffConfiguration[StageClusterCheck])
+			if err != nil && errors.Is(err, service.ErrNodeLoadMetricUnavailable) {
+				w.Logger.Error(ctx, "Couldn't check cluster's current load, proceeding with the backup")
+				return nil
+			}
+			return err
 		},
 		StageAwaitSchema: func() error {
 			clusterSession, err := s.clusterSession(ctx, clusterID)
--- a/pkg/service/backup/service.go
+++ b/pkg/service/backup/service.go
@@ -779,6 +779,9 @@ func (s *Service) Backup(ctx context.Context, clusterID, taskID, runID uuid.UUID
 		return nil
 	}
 	stageFunc := map[Stage]func() error{
+		StageClusterCheck: func() error {
+			return w.CheckCluster(ctx, hi, s.backoffConfiguration[StageClusterCheck])
+		},
 		StageAwaitSchema: func() error {
 			clusterSession, err := s.clusterSession(ctx, clusterID)
 			if err != nil {
--- a/pkg/service/backup/service.go
+++ b/pkg/service/backup/service.go
@@ -16,6 +16,8 @@ import (
 	"github.com/scylladb/go-set/strset"
 	"github.com/scylladb/gocqlx/v2"
 	"github.com/scylladb/gocqlx/v2/qb"
+	"go.uber.org/atomic"
+
 	"github.com/scylladb/scylla-manager/v3/pkg/metrics"
 	"github.com/scylladb/scylla-manager/v3/pkg/schema/table"
 	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
@@ -24,9 +26,9 @@ import (
 	"github.com/scylladb/scylla-manager/v3/pkg/util/inexlist/dcfilter"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/inexlist/ksfilter"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/parallel"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/retry"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/timeutc"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
-	"go.uber.org/atomic"
 )
 
 const defaultRateLimit = 100 // 100MiB
@@ -47,6 +49,8 @@ type Service struct {
 	scyllaClient   scyllaclient.ProviderFunc
 	clusterSession SessionFunc
 	logger         log.Logger
+
+	backoffConfiguration map[Stage]retry.Backoff
 }
 
 func NewService(session gocqlx.Session, config Config, metrics metrics.BackupMetrics, clusterName ClusterNameFunc, scyllaClient scyllaclient.ProviderFunc,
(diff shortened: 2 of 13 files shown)


pkg/service/backup/service.go
```diff
@@ -76,9 +80,22 @@ func NewService(session gocqlx.Session, config Config, metrics metrics.BackupMet
 		scyllaClient:   scyllaClient,
 		clusterSession: clusterSession,
 		logger:         logger,
+
+		backoffConfiguration: DefaultStagesBackoffSetup(),
 	}, nil
 }
 
+func NewTestService(session gocqlx.Session, config Config, metrics metrics.BackupMetrics, clusterName ClusterNameFunc, scyllaClient scyllaclient.ProviderFunc,
```
mrVanboy, line 88: Future us and other maintainers will appreciate docs to distinguish between `NewService` and `NewTestService` ;)

pkg/service/backup/worker_clustercheck.go
```diff
@@ -0,0 +1,36 @@
...
+		for _, h := range hosts {
+			load, err := w.Client.CurrentLoad(ctx, h.IP)
+			if err != nil {
+				return retry.Permanent(errors.Wrap(err, "node's load metric"))
+			}
+			if load >= w.Config.MaxLoadThreshold {
+				return errors.Errorf("load %s on %s node is above configured threshold equal to %s",
+					strconv.FormatFloat(load, 'f', 2, 64),
+					h.IP,
+					strconv.FormatFloat(w.Config.MaxLoadThreshold, 'f', 2, 64))
```
mrVanboy, line 30: Could I have one nitpick? Why not utilize float conversion as a part of `Errotf`?

```suggestion
				return errors.Errorf("load %.2f on %s node is above configured threshold equal to %.2f",	load, h.IP, w.Config.MaxLoadThreshold)
```

mmatczuk: I believe `%.2f` should do the trick

karol-kokoszka: @mrVanboy of course you could and thanks for the comment. Will do.
@mmatczuk you're alive ! :) How are the things going ? Wanna go back :) ?

pkg/service/backup/config.go
```diff
@@ -37,6 +40,9 @@ func (c *Config) Validate() error {
 	if c.AgeMax < 0 {
 		err = multierr.Append(err, errors.New("invalid age_max, must be >= 0"))
 	}
+	if c.MaxLoadThreshold < 0 {
```
mrVanboy, line 43: Can the `MaxLoadThreshold` be `101` and higher? 

I would like to be sure that this validation is in sync with the condition we have in `worker_clustercheck.go`:
```go

func (w *worker) CheckCluster(ctx context.Context, hosts []hostInfo, clusterCheckBackoff retry.Backoff) error {
    ...
	if w.Config.MaxLoadThreshold == 100 {
		return nil
	}
    ...
}
```

Where we have exact equality with `100` hardcoded.

Also, do we want to consider `0` as a valid value in the config for disabling the repairs? If yes, I'll appreciate documenting such a fact in the config's field.

pkg/service/backup/worker_clustercheck.go
karol-kokoszka: Done

pkg/service/backup/service.go
karol-kokoszka: Done

pkg/service/backup/worker_clustercheck.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/scylladb/scylla-manager/v3/pkg/util/retry"
+)
+
+func (w *worker) CheckCluster(ctx context.Context, hosts []hostInfo, clusterCheckBackoff retry.Backoff) error {
```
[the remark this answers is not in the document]
karol-kokoszka: Done. 
On the other hand, this method could be private.
It's not used outside of the package at all.

pkg/service/backup/config.go
karol-kokoszka: Good point.
Let me change the check.

karol-kokoszka: About documenting 0 value.
Is this value for load even possible ? I think it's rather obvious that 0% disables the backup/repair.

Michal-Leszczynski: I would say that setting some field to 0 in config file should disable this option and not take the value 0 literally, especially when 0 value does not really make any sense.

For example, if backup's config `AgeMax` field is set to 0, it means that any backup run can be resumed, no matter how old it is, when if `AgeMax` field was taken literally, it should mean that no backup run can be resumed.
The same goes for `DiskSpaceFreeMinPercent` field, but in this case 0 taken literally and as disabling option has the same effect.

So I would propose that setting `MaxLoadThreshold` to 0 and 100 should have the same effect and just allow repair to run with any load.

Michal-Leszczynski: Otherwise, I would suggest that setting `MaxLoadThreshold` field to 0 in config should result in validation error - something is probably really wrong if someone is trying to schedule repair that always fails.

Michal-Leszczynski: Also, I would add some validation in terms of the upper bound of `MaxLoadThreshold`, because setting it to 200 also does not make any sense and I could indicate some form of an error.

pkg/service/backup/worker_clustercheck.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	notify := func(err error, wait time.Duration) {
+		w.Logger.Info(ctx, "Current load of one of the nodes is above the configured threshold...", "error", err, "wait", wait)
+	}
+
+	return retry.WithNotify(ctx, func() error {
+		for _, h := range hosts {
+			load, err := w.Client.CurrentLoad(ctx, h.IP)
+			if err != nil {
+				return retry.Permanent(errors.Wrap(err, "node's load metric"))
```
Michal-Leszczynski: So if we can't extract metrics of only one node, the validation of the others should be skipped? Do we want such behavior? 

Moreover, this approach can return different results for different orders of nodes in the slice.
Let's assume that the slice contains both node that exceeds load param and node from which we cannot extract load metrics. Then, depending on which node is examined first, we return different errors.
This in combination with the fact, that decision about proceeding with the backup depends on the type of returned error, can create some issues.

I would suggest that if we can't extract metrics of a given node, we just skip it's validation, but we proceed with the validation of the remaining nodes and if any node exceeds load limit, we do not proceed with the backup.

pkg/service/backup/worker_clustercheck.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/scylladb/scylla-manager/v3/pkg/util/retry"
+)
+
+// CheckCluster checks the current load on every element of hosts and returns the first one exceeding the configured
```
Michal-Leszczynski, line 15: > every element of hosts and returns the first one exceeding the configured maximum threshold

This comment sounds to me as if this function was supposed to return an actual node / load, not just validation error.
Maybe change the comment a little bit so that it indicates validation in a more direct way.

pkg/util/retry/backoff.go
```diff
@@ -53,3 +53,21 @@ func (b BackoffFunc) Reset() {}
 func (b BackoffFunc) Clone() Backoff {
 	return b
 }
+
+func DefaultExponentialBackoff() Backoff {
+	const (
+		waitMin        = 15 * time.Second // nolint: revive
+		waitMax        = 1 * time.Minute
+		maxElapsedTime = 15 * time.Minute
```
Michal-Leszczynski, line 61: Should we perhaps add some way for user to configure the time we wait for nodes to go below load threshold?
This could be useful, since some clusters might have a pattern of spikes that usually last for let's say about an hour.

Moreover, AFAIK Manager's scheduler also reschedules tasks that returned an error with exponential backoff, so should we take that into consideration when setting default value here?.

Michal-Leszczynski (changes requested):

pkg/service/backup/config.go
```diff
@@ -6,22 +6,25 @@ import (
...
+
+	"github.com/scylladb/scylla-manager/v3/pkg/service"
 )
 
 // Config specifies the backup service configuration.
 type Config struct {
 	DiskSpaceFreeMinPercent   int           `yaml:"disk_space_free_min_percent"`
 	LongPollingTimeoutSeconds int           `yaml:"long_polling_timeout_seconds"`
 	AgeMax                    time.Duration `yaml:"age_max"`
+	MaxLoadThreshold          float64       `yaml:"max_load_threshold"`
```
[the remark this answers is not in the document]
mrVanboy: I wasn't sure here too. Maybe there is a requirement to fine-tune this value and be able to differentiate between `69` and `69,69`.

> maybe simple int would be enough to store this value

could we rather consider `uint`? Or even `uint8`. It's just safer and prevents manually introduced bugs.

pkg/service/backup/worker_clustercheck.go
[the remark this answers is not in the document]
mrVanboy: Thanks! Yea, you are right, we can make it non-exported. Non-critical tbh.

Just to be clear: I still prefer, even if it's private, to stay friendly to other maintainers too, and not force them to inspect the source code of the function to understand what's going on. Especially as it has logic that is not intuitive from the name/signature of the method itself.

pkg/service/backup/worker_clustercheck.go
mrVanboy: Agree, it's on me actually, I suggested that comment 😅

pkg/service/backup/config.go
karol-kokoszka: > I would say that setting some field to 0 in config file should disable this option and not take the value 0 literally, especially when 0 value does not really make any sense.

That's true, but frankly speaking setting MaxLoadThreshold to a bit higher value, e.g. **1** have almost the same effect.
Besides, it's float, you can set 0.0001 etc..
It's still > 0, but the effect is that there will be almost no backups/repairs taken.
In my understanding, setting some config param to 0 disables the feature only if we provide binary value to choose from. This is not the case here.

The value defines **maximal load threashold on nodes** that is OK to start repair/backup tasks. It's up to the end user to set this value he wants.

Default is 100 = no check.

Let me add explanation around this config field to the official scylla manager documentation, so that's it's clear what this value means.
As an addition, I'll add warning to config validation that will LOG WARNING when the value is low (let's say < 10).

I'm OK with adding validation error on value > 100 tough.

pkg/service/repair/service.go
```diff
@@ -429,6 +453,30 @@ func (s *Service) Repair(ctx context.Context, clusterID, taskID, runID uuid.UUID
...
+
+	return retry.WithNotify(ctx, func() error {
+		for _, h := range hosts {
+			load, err := client.CurrentLoad(ctx, h)
+			if err != nil {
+				return retry.Permanent(errors.Wrap(err, "node's load metric"))
+			}
+			if load >= s.config.MaxLoadThreshold {
+				return errors.Errorf("load %.2f on %s node is above configured threshold equal to %.2f",
+					load, h, s.config.MaxLoadThreshold)
```
avikivity, line 475: What if the load never goes down?

avikivity: I think we should. If the cluster is underprovisioned, it's better to sacrifice performance than to sacrifice data correctness or disaster recovery. It's easy to add more nodes, not so easy to recover from corruption or data loss.

PR #3247, opened 2022-10-26
Missing flag descriptions

Michal-Leszczynski:
Commit 65e2d34d rewrote sctool commands and misspelled 'fail-fast', which resulted in lack of description for this flag.
On the other hand, description for 'retention-days' has never been provided.


karol-kokoszka (approved):
:+1:

PR #3262, opened 2022-11-08
CI: skip publish docker during release creation

yaronkaikov:
Following 79107c2e351458339e6e91129f4f002f0909bd81 , we don't really need to publish docker images since it's been done by our Jenkins job. Adding `--skip-publish` so the release in GitHub will be created


karol-kokoszka (approved):
👍

PR #3272, opened 2022-11-29
service/backup: extend scylla snapshot api call timeout to 30m

karol-kokoszka:
Part of https://github.com/scylladb/scylladb/issues/11938

---

Please make sure that:


Michal-Leszczynski (approved):

Michal-Leszczynski:
LGTM:)

Issue #3278, opened 2022-12-22
Missing docker images for ARM


sebver:
When we were trying to run scylla on ARM we noticed there were no docker images published for linux/arm64:

https://hub.docker.com/r/scylladb/scylla-manager/tags
https://hub.docker.com/r/scylladb/scylla-manager-agent/tags

Since they are published for scylla itself, it would be nice if they could be published for the agent/manager as well.

karol-kokoszka:
@yaronkaikov Is it possible to publish linux/arm64 docker images as well ?

yaronkaikov:
@karol-kokoszka In general we can. I assume we want one docker image which supports multi-arch (Same as we do for Scylla). based on the current code in Scylla-manager I can see we don't build arm-based docker, so first, we should add it.

DhirajBhakta:
This PR seems to have been reverted. Any plans to add ARM support soon?

Kavinraja-Ganesan:
@yaronkaikov -- Any reason for this PR to be reverted 😕 ? I was looking for arm64 scylla-manager/agent images.

Kavinraja-G:
@yaronkaikov -- Any reason for this PR to be reverted 😕 ? I was looking for arm64 scylla-manager/agent images.

yaronkaikov:
@Kavinraja-G Yes, we faced some issues with arm based docker image, I will re-post it soon

tzach:
@karol-kokoszka should we reopen this issue?

karol-kokoszka:
@tzach we are done from the scylla-manager perspective as per my understanding. There is some work to be done on the scylla-pkg (?) side.
 @yaronkaikov Please correct me if I'm wrong. Can you link the issue that must be completed to have ARM images available ?

yaronkaikov:
I think we should re-open it (again) since I am about to change the way we build the manager to include the docker images and manifest.

yaronkaikov:
@karol-kokoszka https://github.com/scylladb/scylla-pkg/issues/3239 , this is the issue from pkg side, only when both will be done we can say we have a proper docker image for ARM

PR #3281, opened 2023-01-02
.goreleaser: add support for arm based docker image

yaronkaikov:
Adding arm support to our Scylla-manager docker image

This will only build additional docker image, the publishing part will be done as part of https://github.com/scylladb/scylla-pkg/issues/3239

Closes: https://github.com/scylladb/scylla-manager/issues/3278

### DIFF
--- a/dist/.goreleaser.yaml
+++ b/dist/.goreleaser.yaml
@@ -233,19 +233,66 @@ nfpms:
 
 dockers:
   - ids:
-      - server
-      - client
+    - server
+    - client
+    use: docker
+    goos: linux
+    goarch: amd64
     image_templates:
-      - "scylladb/scylla-manager:{{ .Version }}"
-    dockerfile: docker/scylla-manager.dockerfile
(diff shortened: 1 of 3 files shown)


dist/.goreleaser.yaml
```diff
@@ -233,19 +233,66 @@ nfpms:
...
     image_templates:
-      - "scylladb/scylla-manager:{{ .Version }}"
-    dockerfile: docker/scylla-manager.dockerfile
+      - "scylladb/scylla-manager:{{ .Version }}.amd64"
+    dockerfile:
+      docker/scylla-manager.dockerfile
     extra_files:
       - docker/scylla-manager.yaml
+    build_flag_templates:
+    - "--build-arg=BASE_IMAGE=ubuntu:20.04"
```
mykaul, line 248: 20? not 22?

yaronkaikov: make sense to move it to 22.04 :+1:

karol-kokoszka (approved):

PR #3287, opened 2023-01-16
Revert ".goreleaser: add support for arm based docker image"

karol-kokoszka:
Reverts scylladb/scylla-manager#3281


yaronkaikov (approved):
Thanks, i will sent v2 soon

Issue #3202, opened 2022-07-13
golangci: reenable go-critic when it works with generics


mmatczuk:
The [issue]( https://github.com/go-critic/go-critic/issues/1224) in the linter makes it impossible to run `make .check-lint`. When the issue is gone go-critic needs to be reenabled.

mmatczuk:
```
ERRO [runner] Panic: gocritic: package "scheduler" (isInitialPkg: true, needAnalyzeSource: true): assertion failed: goroutine 49423 [running]:
runtime/debug.Stack()
        /opt/homebrew/Cellar/go/1.18.3/libexec/src/runtime/debug/stack.go:24 +0x68
github.com/golangci/golangci-lint/pkg/golinters/goanalysis.(*action).analyzeSafe.func1()
        /Users/michal/go/pkg/mod/github.com/golangci/golangci-lint@v1.46.2/pkg/golinters/goanalysis/runner_action.go:101 +0x104
panic({0x10529dc60, 0x10544a750})
        /opt/homebrew/Cellar/go/1.18.3/libexec/src/runtime/panic.go:838 +0x204
go/types.assert(...)
        /opt/homebrew/Cellar/go/1.18.3/libexec/src/go/types/errors.go:21
go/types.(*StdSizes).Alignof(0x140001b52d0, {0x105452500?, 0x1402ea0c750})
        /opt/homebrew/Cellar/go/1.18.3/libexec/src/go/types/sizes.go:72 +0x278
go/types.(*StdSizes).Offsetsof(0x105452398?, {0x1402eb7b890, 0x5, 0x0?})
        /opt/homebrew/Cellar/go/1.18.3/libexec/src/go/types/sizes.go:101 +0xcc
go/types.(*StdSizes).Sizeof(0x140001b52d0, {0x105452398?, 0x1402ea57650})
        /opt/homebrew/Cellar/go/1.18.3/libexec/src/go/types/sizes.go:154 +0x1a8
github.com/go-critic/go-critic/checkers.(*rangeValCopyChecker).VisitStmt(0x140165da880, {0x1054558c8?, 0x1402e0db270?})
        /Users/michal/go/pkg/mod/github.com/go-critic/go-critic@v0.6.3/checkers/rangeValCopy_checker.go:68 +0xb0
github.com/go-critic/go-critic/checkers/internal/astwalk.(*stmtWalker).WalkFile.func1({0x105451fb0?, 0x1402e0db270?})
[...]

PR #3292, opened 2023-01-25
Go and golangci version update

Michal-Leszczynski:
Updated go and golangci linter (and also some other required dependencies) versions.
Unfortunately, I wasn't able to fix issue #3202 and as I understand, we still have to wait until `go-critic` is fully fixed.
I disabled deprecated linters:
- deadcode
- varcheck
- structcheck

And two new ones:
- interfacebloat
- nosnakecas

I don't think that they are that useful in SM repository, but they create some issues.


karol-kokoszka (approved):
:+1:

Issue #3289, opened 2023-01-16
sctool progress fails when checking the progress of a restore task: `invalid memory address or nil pointer dereference`


ShlomiBalalis:
Client version: 3.1-dev-0.20230111.71ee737dafe-SNAPSHOT-71ee737dafe
Server version: 3.1-dev-0.20230111.71ee737dafe-SNAPSHOT-71ee737dafe

In this scenario I run a backup task to completion and trunate the table that was backed up. Afterwards we add a new node to the cluster and run a restore, like so:
```
sctool restore -c 9122dc14-225b-4839-9941-0f88ff1dfe9e --restore-tables --location s3:backup-bucket --snapshot-tag sm_20230116125214UTC

restore/c31ed335-d2c8-4b85-b186-4a3d575b2d57
```
But when I try to examine the progress of the task, the command fails:
```
sctool progress restore/c31ed335-d2c8-4b85-b186-4a3d575b2d57 -c 9122dc14-225b-4839-9941-0f88ff1dfe9e
Run:		e2a9fcff-959c-11ed-a013-f4ee08c9cc47
Status:		Error: template: restore_progress:3:12: executing "restore_progress" at <status>: error calling status: runtime error: invalid memory address or nil pointer dereference
```

The issue seems to be consistent in this scenario and in this scenario alone.

Logs:
[scylla-manager.log](https://github.com/scylladb/scylla-manager/files/10427506/scylla-manager.log)
Agent logs:
* [manager_agent_1.log](https://github.com/scylladb/scylla-manager/files/10427521/manager_agent_1.log)
* [manager_agent_2.log](https://github.com/scylladb/scylla-manager/files/10427528/manager_agent_2.log)
* [manager_agent_3.log](https://github.com/scylladb/scylla-manager/files/10427531/manager_agent_3.log) (The node that was added)

Michal-Leszczynski:
I believe that the following thing happened:
- SM got restore target (preparation before restore and before inserting restore run into DB)
- SM was queried for restore progress
- SM should return 404 error because it couldn't find restore run in DB, but in this scenario SM prefers to return empty progress information instead of error and that's the place where I forgot to insert empty restore progress instead of `nil`
- SM crashed because of this bug

So It's not about really about adding new node to the cluster, but rather about asking for restore progress before restore run could be inserted into DB.
Anyways, it's a bug and I already created fix for it (#3293) but I still need to verify if this for sure fixes that issue.

[...]

PR #3293, opened 2023-01-26
Restore progress display fix

Michal-Leszczynski:
SM prefers to return empty progress instead of error 404 when task run hasn't been found. 
Restore (and validateBackup as well I believe) was lacking this approach and that's why #3289 happened.

Still need to check if this PR for sure fixes this issue.


karol-kokoszka (approved):
:+1:

PR #3303, opened 2023-02-09
Fixed repair example documentation

### LINKED ISSUE #3229: Issue in page Examples
I would like to report an issue in page https://manager.docs.scylladb.com/branch-3.0/repair/examples

### Problem
```
Procedure

To run an immediate repair on the prod-cluster cluster, saving the repair in my-repairs, run the following command replacing the -c cluster flag with your cluster’s cluster name or ID and replace the -L flag with your repair’s location:

sctool repair -c prod-cluster -L 's3:my-repairs'
```
Repair cannot be saved and it does not accept location argument. 

###  Suggest a fix
```
Procedure

To run an immediate repair on the cluster, run the following command replacing the -c flag with your cluster’s name or ID.

sctool repair -c prod-cluster
```

Michal-Leszczynski:
There is a small bug in repair example documentation originating from copying some examples from backup procedure.

Fixes #3229

### DIFF
--- a/docs/source/repair/examples.rst
+++ b/docs/source/repair/examples.rst
@@ -108,12 +108,17 @@ Additional parameters can be used. Refer to :ref:`repair parameters <sctool-repa
 
 **Procedure**
 
-To run an immediate repair on the prod-cluster cluster, saving the repair in my-repairs, run the following command
-replacing the ``-c`` cluster flag with your cluster's cluster name or ID and replace the ``-L`` flag with your repair's location:
+To run an immediate repair on the cluster, run the following command specifying the name or ID of your cluster with the ``-c`` flag:
 
 .. code-block:: none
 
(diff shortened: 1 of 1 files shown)


annastuchlik:
Could you make one more update? Your fix looks good, but the wording was strange from the beginning - it shouldn't be about replacing the flat, but about passing a specific cluster name with that flag. I suggest changing the wording and adding an example:



To run an immediate repair on the cluster, run the following command specifying the name of your cluster with the ``-c`` flag:

```
.. code-block:: none

   sctool repair -c <cluster_name>
```
   
Example:

```
.. code-block:: none

   sctool repair -c prod-cluster
```

Michal-Leszczynski:
done! I also changed `name of your cluster` to `name or ID of your cluster`.

karol-kokoszka (approved):
👍

Issue #3305, opened 2023-02-16
Move Docker base images to Ubuntu 22 (from Ubuntu 20)


mykaul:
Move to 22 LTS, currently all images are based on 20.

karol-kokoszka:
@mykaul it's merged to master. We don't have it on the 3.1 branch. Should I include it into 3.1 ?

PR #3306, opened 2023-02-16
docker: move to ubuntu22.04 based image

yaronkaikov:
Move docker images to be based on ubuntu:22.04

Closes: https://github.com/scylladb/scylla-manager/issues/3305


karol-kokoszka (approved):
:+1:

PR #3310, opened 2023-02-22
Restore all choosen backed-up tables 

Michal-Leszczynski:
Restoring tables contents should restore all (modulo `-K` flag) backed-up tables - also the system ones like `system_auth` etc. We should relay on backup mechanism not to back up tables that shouldn't ever be restored.


Michal-Leszczynski:
This issue was created by me at the beginning of writing universal restore procedure, because I believed that system tables shouldn't be restored in general. It turns out, that indeed `system` tables shouldn't be restored, but tables from e.g. `system_auth` should be restored.

Michal-Leszczynski:
Should I squash those two commits into one?

karol-kokoszka (approved):
:+1: @Michal-Leszczynski yes, please squash into one

PR #3321, opened 2023-03-05
Add sstable versioning

### LINKED ISSUE #3288: [SPIKE] Check possibilities of storing multiple SSTable versions in backup
We need to find a way of storing multiple versions of single SSTable in backup location.

Option with enabled object versioning on cloud storage.
- verify whether S3 object versioning and GCS object versioning meets our requirements
- validate if Rclone fork we use, supports object versioning on S3 and GCS
- validate if it's possible to choose expected version of object when downloading backup with `download-files` (following restore procedure from ansible script) - we should rely on object version creation timestamp + snapshot tag that is the timestamp
- check the costs of enabling versioning on S3/GCS buckets

Option with artificially added snapshot tag to the object name in cloud storage.
- verify that we can easily update metadate file and clean it when manager downloads it to perform restore
- verify that RClone can rename file when moving it to cloud storage
- validate that the change is compatible with already existing backups

@Michal-Leszczynski please add anything what you think is relevant to verify/validate here.

Expected output from this task is to have verified/validated plan of adding support for multiple SSTable versions in backup location that will play well with the restore procedure.

Michal-Leszczynski:
This PR introduces versioning of SSTables in Scylla Manager.
The three main points which are influenced by that are backup, purge and restore procedure.

Fixes #3288


pkg/scyllaclient/client_rclone_agent_integration_test.go
```diff
@@ -58,7 +58,7 @@ func TestRcloneSkippingFilesAgentIntegration(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	id, err := client.RcloneCopyDir(ctx, testHost, remotePath(""), "data:tmp/copy")
+	id, err := client.RcloneCopyDir(ctx, testHost, remotePath(""), "data:tmp/copy", "")
```
karol-kokoszka, line 61: I don't see any integration test validating usage of "suffix" option.
Please add. 
It's not enough to test it just with the backup / restore.

pkg/service/backup/purger.go
```diff
@@ -42,16 +44,25 @@ func staleTags(manifests []*ManifestInfo, retentionMap RetentionMap) *strset.Set
...
 		}
 
 		if taskPolicy.Retention > 0 && taskTags.Size()-taskPolicy.Retention > 0 {
 			l := taskTags.List()
 			sort.Strings(l)
-			tags.Add(l[:len(l)-taskPolicy.Retention]...)
+			cut := len(l) - taskPolicy.Retention
+			tags.Add(l[:cut]...)
+
+			t, _ := SnapshotTagTime(l[cut]) // nolint: errcheck
```
karol-kokoszka, line 58: Please check the error and just log it.

pkg/service/backup/purger.go
```diff
@@ -110,9 +126,47 @@ func (p purger) PurgeSnapshotTags(ctx context.Context, manifests []*ManifestInfo
...
+		nodeDir := RemoteSSTableBaseDir(anyM.ClusterID, anyM.DC, anyM.NodeID)
+		opts := &scyllaclient.RcloneListDirOpts{
+			FilesOnly:     true,
+			Recurse:       true,
+			VersionedOnly: true,
+		}
+
+		cb := func(item *scyllaclient.RcloneListDirItem) {
+			snapshotSuffix := path.Ext(item.Name)[1:]
+			t, err := SnapshotTagTime(snapshotSuffix)
```
karol-kokoszka, line 140: I don't like it, because there is a hidden assumption made in the code.
I know that the file extension is the snapshot_tag, but I know it because I'm going through the PR and I saw the part where you defined in `rc.go`
```
const snapshotTagSuffixRegex = `{**.sm_*UTC}`
```
Please wrap the versioning related code into `pkg/util/versioning` package or anything else what will define it's single responsibility.
`pkg/util/versioning` should expose everything related to your versioning implementation including`snapshotTagSuffixRegex` const from `rc.go`

pkg/service/backup/purger_test.go
```diff
@@ -51,7 +51,7 @@ func TestStaleTags(t *testing.T) {
 	x.Temporary = true
 	manifests = append(manifests, x)
 
-	tags := staleTags(manifests, RetentionMap{task0: {0, 3}, task1: {0, 2}})
+	tags, _ := staleTags(manifests, RetentionMap{task0: {0, 3}, task1: {0, 2}})
```
karol-kokoszka, line 54: You have a great opportunity to test here if `staleTags` returns the oldest tag the correct way. Please don't ignore the `oldest` value from `staleTags` return but include it into the assertion.

pkg/service/backup/restore_worker_data.go
```diff
@@ -443,43 +495,104 @@ func (w *restoreWorker) newRunProgress(ctx context.Context, run *RestoreRun, tar
...
+	// The assumption is that the existence of versioned files is low and that they
+	// are rather small, so we can do it in a synchronous way.
+	// Moving files can be done in full parallel because of rclone ability to limit transfers.
+	err := parallel.Run(len(versionedBatch), parallel.NoLimit, func(i int) error {
+		file := versionedBatch[i]
+
+		dst := path.Join(dstDir, file.Name)
+		src := path.Join(srcDir, strings.Join([]string{file.Name, file.Version}, "."))
+
+		if err := w.Client.RcloneMoveFile(ctx, host, dst, src); err != nil {
```
karol-kokoszka, line 567: Why move files ? We don't want to change the bucket insides. We just want to restore using correct version of SSTable.

karol-kokoszka (changes requested):
Why to move files when restoring from versioned backup files ? 
We don't want to change the data in backup location at all.

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: It indeed should be CopyFile, my bad!

pkg/service/backup/purger.go
Michal-Leszczynski: I added it to `service/backup/backupspec` because it seems to better fit there.

pkg/service/backup/purger_test.go
Michal-Leszczynski: I also extended `staleTags` for mixed retention policies.

karol-kokoszka:
@Michal-Leszczynski I added you as a reviewer. Here https://github.com/scylladb/scylla-enterprise/issues/2534#issuecomment-1470381254 you have an explanation of what do we want to achieve.

karol-kokoszka:
@Michal-Leszczynski pls try to avoid force-pushing after making PR fixes. It's hard to find where the change is made.

karol-kokoszka (approved):
:+1:

PR #3322, opened 2023-03-08
Hardening restore

### LINKED ISSUE #3296: Set proper restore timeout
While developing restore procedure some timeouts were set to constant values:
https://github.com/scylladb/scylla-manager/blob/aa670a6aed10796ef92ce8708bae41e03238ce31/pkg/service/backup/restore_worker_data.go#L610-L614

https://github.com/scylladb/scylla-manager/blob/aa670a6aed10796ef92ce8708bae41e03238ce31/pkg/scyllaclient/client_scylla.go#L829-L830

Mentioned places in the code correspond to `load&stream` procedure. SM has no way of tracking `load&stream` progress, it just hangs on connection until 200 response is returned (that's how `load&stream` is designed right now).
Because of that, I added this 1 hour timeout so that SM does not hang on this connection indefinitely.

My question is, should we even provide any timeout for hanging on this connection?
If so, we can estimate this timeout based on the size of streamed SSTables, but that requires some information about the speed of `load&stream`.

Just to clarify, SM is using async rclone API and long polling  in order to download SSTables from backup location into node's upload directory and because of that no timeouts are used nor necessary here.

### LINKED ISSUE #3315: Updating a restore task should not require the 'snapshot-tag` parameter
Updating a paused restore task fails if you don't include the `snapshot-tag` parameter, even when it's irrelevant:
```
sctool restore update restore/c4c99c51-7feb-4689-9e18-a1366ec75b70 -c 993bea8d-7b86-4820-b322-eb1f54911950 --batch-size 1

Error: required flag(s) "snapshot-tag" not set
```
It really shouldn't be required for the update command.

Michal-Leszczynski:
This PR contains many small changes to restore procedure that should make it more robust.
List of changes:
- forbid updating core restore flags (it creates unsafe scenarios as restore procedure depends on iterating over restored files in specific order and changing e.g. -L flag may disrupt that)
- ensure listing manifests in a deterministic way (reason same as above)
- try to look for resumed rclone job before recreating it (although rclone jobs cannot be resumed, this might be useful in scenarios like SM crash)
- remove timeout from calling load and stream with proper resume support and logging (it's impossible to set right timeout so it's better not to do it at all)

Fixes #3315
Fixes #3296

Edit: moved changes regarding compaction and usage of primary_replica_only to separate PR


karol-kokoszka (approved):
:+1:

Michal-Leszczynski:
Removed changes regarding compaction and usage of primary_replica_only, as they will be targeted in another PR.
Moreover, added 1 "if" in the last commit in order to avoid wasting 10 seconds when encountering an actual error during load and stream.

PR #3300, opened 2023-02-08
Restore procedure docs description

Michal-Leszczynski:
This PR adds documentation for universal restore procedure.

Fixes #3295

PR #3331, opened 2023-03-23
Universal restore metrics update

karol-kokoszka:
This PR:
- removes filesSizeBytes metric
- removes filesDownloadedBytes metric
- removes filesSkippedBytes metric
- removes filesFailedBytes metric
- adds batchSize metric that show how restore batches are propagated between nodes
- adds remainingBytes metric that shows how much bytes from backup dir are still not restored
- adds state metrics that shows the node state during the repair (idle / downloading / load&stream / error )
- changes default parallelism to 0 (infinity)
- updates testing/prometheus.yaml with IP of the manager running on local machine

---

Please make sure that:


pkg/service/backup/restore_worker_data.go
```diff
@@ -472,6 +477,18 @@ func (w *restoreWorker) newRunProgress(ctx context.Context, run *RestoreRun, tar
...
 
+func (w *restoreWorker) updateBatchSizeMetric(ctx context.Context, clusterID uuid.UUID, batch []string, host, srcDir string) {
+	var batchSize int64
+	for _, file := range batch {
+		fi, err := w.Client.RcloneFileInfo(ctx, host, srcDir+"/"+file)
+		if err != nil {
+			w.Logger.Error(ctx, "Failed to check file size", "err", err)
+		}
+		batchSize += fi.Size
+	}
```
Michal-Leszczynski, line 524: Maybe it's better to fetch this information for all files using `RcloneListDirIter` at the beginning of restoring given table?

Michal-Leszczynski: Moreover, fetching it here will be in conflict with restoring versioned SSTables, because `batch` contains only the regular file names without their specific versions.

pkg/service/backup/restore_worker_data.go
```diff
@@ -435,7 +435,7 @@ func (w *restoreWorker) newRunProgress(ctx context.Context, run *RestoreRun, tar
 		return nil, errors.Wrap(err, "validate free disk space")
 	}
 
-	takenIDs := w.chooseIDsForBatch(ctx, h.Shards, target.BatchSize)
+	takenIDs := w.chooseIDsForBatch(ctx, target.BatchSize)
```
Michal-Leszczynski, line 475: Was this change consulted with someone with more understanding on the importance of shard cnt in the process of load and stream?

pkg/metrics/backup.go
```diff
@@ -186,3 +192,25 @@ func (rm RestoreM) UpdateBatchSize(clusterID uuid.UUID, host string, size int64)
...
 	rm.batchSize.With(l).Add(float64(size))
 }
+
+// SetRemainingBytes sets restore "remaining_bytes" metric.
+func (rm RestoreM) SetRemainingBytes(clusterID uuid.UUID, location backupspec.Location, snapshotTag string, keyspaces []string, remainingBytes int64) {
+	l := prometheus.Labels{
+		"cluster":      clusterID.String(),
+		"location":     location.String(),
+		"snapshot_tag": snapshotTag,
+		"keyspaces":    strings.Join(keyspaces, ","),
```
Michal-Leszczynski: Do we want to change labels on metrics?
Why `manifest_path` has been replaced with `location`? 
Is `snapshot_tag` really needed here?
Why slice of keyspaces and not keyspace + table?

pkg/service/backup/restore_worker_data.go
```diff
@@ -57,6 +58,28 @@ func (w *restoreWorker) restoreData(ctx context.Context, run *RestoreRun, target
...
+					func(fm FilesMeta) error {
+						restorePerLocationDataSize += fm.Size
+						return nil
+					})
+			})
+		if err != nil {
+			w.Logger.Info(ctx, "Couldn't count restore data size", "location", w.location)
+			continue
+		}
+		w.metrics.SetRemainingBytes(run.ClusterID, location, target.SnapshotTag, target.Keyspace, restorePerLocationDataSize)
```
Michal-Leszczynski, line 80: `target.Keyspace` is the glob pattern specified by the `-K` flag, not the list of all restored keyspaces.
Maybe `Units` from `RestoreRun` are more useful in this situation?
But still I don't understand the need to label those metrics by keyspace slice.

pkg/service/backup/restore_worker_data.go
```diff
@@ -247,7 +270,9 @@ func (w *restoreWorker) workFunc(ctx context.Context, run *RestoreRun, target Re
 			pr.setRestoreCompletedAt()
 			w.insertRunProgress(ctx, pr)
 
-			w.metrics.UpdateRestoreProgress(w.ClusterID, pr.ManifestPath, pr.Keyspace, pr.Table, pr.Downloaded+pr.Skipped)
+			restoredBytes := pr.Downloaded + pr.Skipped
+			w.metrics.UpdateRestoreProgress(w.ClusterID, pr.ManifestPath, pr.Keyspace, pr.Table, restoredBytes)
+			w.metrics.DecreaseRemainingBytes(w.ClusterID, w.location, w.SnapshotTag, target.Keyspace, restoredBytes)
```
Michal-Leszczynski: What about when batch returns to the pool when given node encounters an error?
I guess the metrics should be increased in this case.

pkg/metrics/backup.go
```diff
@@ -214,3 +217,28 @@ func (rm RestoreM) DecreaseRemainingBytes(clusterID uuid.UUID, location backupsp
...
+	// RestoreStateError means that node ended up with error
+	RestoreStateError
+)
+
+// SetRestoreState sets restore "state" metric.
+func (rm RestoreM) SetRestoreState(clusterID uuid.UUID, location backupspec.Location, snapshotTag, host string, state RestoreState) {
+	l := prometheus.Labels{
+		"cluster":      clusterID.String(),
+		"location":     location.String(),
+		"snapshot_tag": snapshotTag,
```
Michal-Leszczynski, line 214: Are those labels useful here?

pkg/service/backup/restore_worker_data.go
```diff
@@ -210,10 +211,20 @@ func (w *restoreWorker) workFunc(ctx context.Context, run *RestoreRun, target Re
...
+				w.metrics.SetRestoreState(run.ClusterID, w.location, target.SnapshotTag, h.Host, metrics.RestoreStateError)
+				return
+			}
+			w.metrics.SetRestoreState(run.ClusterID, w.location, target.SnapshotTag, h.Host, metrics.RestoreStateIdle)
+		}()
+
 		for {
+			w.metrics.SetRestoreState(run.ClusterID, w.location, target.SnapshotTag, h.Host, metrics.RestoreStateDownloading)
+
 			pr, err := w.prepareRunProgress(ctx, run, target, h, dstDir, srcDir)
```
Michal-Leszczynski, line 228: The downloading state should be set only after `prepareRunProgress`, because host might wait in `chooseIDsForBatch` before downloading remaining batches. This also means that the `idle` state should be set at the end of the for loop.

pkg/service/backup/restore_worker_data.go
karol-kokoszka: This change decreases the batch size actually.
`h.Shards` were used as a multiplier for a default batch size.
But actually you're right let me check it with scylla core team.

pkg/service/backup/restore_worker_data.go
karol-kokoszka: > target.Keyspace is the glob pattern specified by the -K flag, not the list of all restored keyspaces.

Yes, that's true.
See this comment https://github.com/scylladb/scylla-manager/pull/3331#discussion_r1146788445

pkg/metrics/backup.go
karol-kokoszka: See this comment https://github.com/scylladb/scylla-manager/pull/3331#discussion_r1146788445

pkg/service/backup/restore_worker_data.go
karol-kokoszka: I don't really follow.
Remaining bytes are decreased when we see that load&stream ended up with the success.
https://github.com/scylladb/scylla-manager/blob/01994811ae1c6736bafc4edf203106b095a93659/pkg/service/backup/restore_worker_data.go#L283-L300

Is there something happening in parallel that I'm not aware of ?

pkg/service/backup/restore_worker_data.go
karol-kokoszka: Makes sense.
Thanks !

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: > The number of shards on the node where you perform load and stream do not have impact on how we process the data.

Does this mean that the number of shards does not have significant influence on load and stream speed?
I'm wondering whether load and stream most expensive operation is the actual file transfer between the nodes, or maybe the process of deciding, to which node should the files be sent (I'm considering the `primary_replica_only` option). If the second case is more expensive, does the number of shards influence its speed?

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: It looks like I attached wrong lines of code, my bad.
I was concerned about the `batch_size` metric and its behavior, when some node (after assembling the batch) encountered an error and returned its batch to the pool (so that the other nodes might try to restore it).
Should we decrease the `batch_size` metric in this case?

pkg/service/backup/restore_worker_data.go
karol-kokoszka: ^^ @asias please take a look on last Michal's question.

pkg/service/backup/restore_worker_data.go
karol-kokoszka: Let's do it, after you merge SSTable versioning.

pkg/service/backup/restore_worker_data.go
karol-kokoszka: I still don't follow.
Metric is not decreased if batch is returned to the pool.
It decreases it only if there is no error on downloading and no error on load&streaming when node tries to restore the batch.

pkg/service/backup/restore_worker_data.go
asias: A node with 32 shards is definitely faster than a node with 2 shards because it has 16x more cpu powers.  But load and stream on 2 shards node is as efficient as on 32 shards. I guess you want to check if the later is true. 

Load and stream does not send files over network. It reads data (mutation fragments) from sstables and figure out which nodes own the data and send the mutation fragments to the owning node. 

The most expensive operations are the actual reading from disk and send over network. The deciding part is relatively cheap.

karol-kokoszka: @Michal-Leszczynski this comment is very relevant here as well https://github.com/scylladb/scylla-manager/pull/3300/files#r1150523568

Michal-Leszczynski: So maybe we want to replace shard count in work distribution process with file size? Well, not exactly files size as (for my understanding) it's safer to upload all SSTables with the same ID at once, so rather the size of the SSTable bundle? 

On the other hand, removing shard count and setting batch size to 1 would also force hosts to divide bundles in a more fair way. Right now hosts just take the bundles from bundle pool, restore them and then check if there are any more bundles left to restore, so if we treat all nodes equally, this should get a good work distribution.

So I agree with removing shard count form the restore procedure.

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: SSTable versioning is now merged!

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski:
```go
	batch := w.batchFromIDs(pr.SSTableID)
	w.updateBatchSizeMetric(ctx, pr.ClusterID, batch, pr.Host, srcDir)

	w.cleanUploadDir(ctx, pr.Host, dstDir, batch)

	jobID, err := w.startDownload(ctx, pr.Host, dstDir, srcDir, batch)
	if err != nil {
		w.deleteRunProgress(ctx, pr)
		w.returnBatchToPool(pr.SSTableID)
		return err
	}
	pr.AgentJobID = jobID
	w.insertRunProgress(ctx, pr)
	return nil
}
```

This seems to be the only place where the `batch_size` metric is increased. My question is, shouldn't we decrease it inside `w.returnBatchToPool(pr.SSTableID)` function? Or is it fine that the sum over `batch_size` might be greater than the actual size of all batches (in case of error)?

pkg/service/backup/restore_worker_data.go
karol-kokoszka: > On the other hand, removing shard count and setting batch size to 1 would also force hosts to divide bundles in a more fair way. Right now hosts just take the bundles from bundle pool, restore them and then check if there are any more bundles left to restore, so if we treat all nodes equally, this should get a good work distribution.

I think deciding if we should set batch_size to 1 instead of 2 is something what we can fine tune later (with the next minor releases).
Let's stay with default batch_size = 2, and let's remove this shards_count dependency at all.

pkg/service/backup/restore_worker_data.go
karol-kokoszka: now I've got it,
Yes, batch_size should be decreased.

Thanks !

pkg/service/backup/restore_worker_data.go
Michal-Leszczynski: So it would be nice to remove it from `restoreHosts` an `initHosts` as well!

pkg/service/backup/restore_worker_data.go
```diff
@@ -614,9 +614,9 @@ func (w *restoreWorker) startDownload(ctx context.Context, host, dstDir, srcDir
 }
 
 // chooseIDsForBatch returns slice of IDs of SSTables that the batch consists of.
-func (w *restoreWorker) chooseIDsForBatch(ctx context.Context, shards uint, size int) []string {
+func (w *restoreWorker) chooseIDsForBatch(ctx context.Context, size int) []string {
 	var (
-		batchSize = size * int(shards)
+		batchSize = size
```
Michal-Leszczynski, line 661: After removing shards count modifier, there's no need to create a new variable here.

karol-kokoszka:
I'm closing this PR in favor of https://github.com/scylladb/scylla-manager/pull/3349

PR #3337, opened 2023-03-30
restore: set default parallelism to 0 (infinity)

karol-kokoszka:
Restore default parallelism to 0 (inf).

---

Please make sure that:


karol-kokoszka:
I need to merge it to allow QA to test restore procedure properly.

PR #3338, opened 2023-03-30
restore: do not include no shards into batch_size

karol-kokoszka:
Batches are too big.
Number of shards is completely irrelevant for restore.

It's a cherry pick from https://github.com/scylladb/scylla-manager/pull/3331

---

Please make sure that:


karol-kokoszka:
I need to merge it to unblock QA.

Michal-Leszczynski (changes requested):
Let's remove shard count completely. It is still present in `initHosts` and `restoreHosts`.

karol-kokoszka:
> Let's remove shard count completely. It is still present in initHosts and restoreHosts.

Definitely makes sense.
PR is updated.

Michal-Leszczynski (approved):

PR #3339, opened 2023-03-30
Separate restoring schema and tables

Michal-Leszczynski:
This PR separates workflow of restoring schema and tables. The main differences are:
- restoring tables requires altering ggs which is forbidden on schema tables
- restoring schema duplicates data across all nodes in order to simulate repair
- restoring schema cannot be resumed and monitoring of its progress is rather weak
- restoring schema does not show data duplication on metrics and progress

Even though it might look like a regression, restoring schema should have much bigger success ratio with those changes.

Sorry for such a big commit, but those changes couldn't really be divided because of all the refactor that needed to be done at once.

### DIFF
--- a/pkg/service/backup/service_backup_restore_integration_test.go
+++ b/pkg/service/backup/service_backup_restore_integration_test.go
@@ -402,6 +402,12 @@ func smokeRestore(t *testing.T, target RestoreTarget, keyspace string, loadCnt, 
 
 	target.SnapshotTag = srcH.simpleBackup(target.Location[0], keyspace)
 
+	pr, err := srcH.service.GetProgress(context.Background(), srcH.clusterID, srcH.taskID, srcH.runID)
+	if err != nil {
+		srcH.t.Fatal(err)
+	}
+	Printf("And: backup progress: %+#v\n", pr)
+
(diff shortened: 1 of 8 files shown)


pkg/service/backup/service_backup_restore_integration_test.go
```diff
@@ -736,6 +742,34 @@ func TestRestoreTablesVersionedIntegration(t *testing.T) {
 	restoreWithVersions(t, target, testKeyspace, testLoadCnt, testLoadSize, corruptCnt)
 }
 
+func TestRestoreSchemaVersionedIntegration(t *testing.T) {
```
karol-kokoszka, line 718: Love it ! Thanks !

karol-kokoszka (approved):
Just one comment that is not important at all.
In `service_backup_restore_integration_test.go` you create `toCorrupt` slice that keeps files that you want to change ofc.
The point is that when you complete this slice, it's possible that `*.crc32` files are not included.
These file are not important at all in terms of restore procedure (as per my understanding), we should only care about sstables.
Maybe it's worth to make sure that only `*.db` are being corrupted.

Michal-Leszczynski:
Added commit in which I increased the amount of versions created in backup location for testing versioned restore. Moreover, now only `*.db` files will be included as versions.

PR #3352, opened 2023-04-07
backup: merge backup/paths to backupspec/paths

karol-kokoszka:
Merges backup/paths.go with backup/backupspec/paths that are serving the same purpose

---

Please make sure that:


Michal-Leszczynski:
Why have you merged commit which failed Integration Tests because the code didn't compile?!?

karol-kokoszka:
Wanted to have it faster on master...
I had to do the follow up anyway.

Michal-Leszczynski:
That's ok, but let's try to wait for the checks to end before merging future PRs

karol-kokoszka:
sure, don't worry, no one should use the latest master
stable/released scylla-manager packages is something what customer should use

PR #3353, opened 2023-04-07
Split tables worker

karol-kokoszka:
The main purpose of this PR is to split restore_worker_tables into two pieces:
- restore worker tables which works in context of the whole restore
- restore worker index which works in context of one manifest index

---

Please make sure that:


karol-kokoszka:
@Michal-Leszczynski I'm merging it, please leave comments if you find something weird in the code. I'll be notified via email about this fact, so it won't be lost.

pkg/service/backup/restore_worker_tables.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/scylladb/go-set/strset"
+	"github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
+)
+
+type tablesWorker struct {
+	restoreWorkerTools
+
+	hosts   []restoreHost // Restore units created for currently restored location
```
Michal-Leszczynski: After separating `tablesWorker` and `indexWorker` there is no actual need to keep `hosts` as a field in `tablesWorker`, because we only pass it straight to the `indexWorker` right after its initialization.
This would also fit into the theme of separating those two structures.

pkg/service/backup/restore_worker_tables.go
```diff
@@ -0,0 +1,158 @@
+package backup
```
Michal-Leszczynski, line 1: This file is lacking a license!

PR #3356, opened 2023-04-11
rclone: compare files basing on checksum

### LINKED ISSUE #3355: rclone: file skip should base on checksum explicitly
We need to rely on checksum instead of size and modification time when rclone compares files.
Otherwise, it's possible that the same files basing on content will be marked as not equal.
Besides, as we enabled versioning for SSTables in 3.1 and master, it will cause creation of another copy of the same file.

Here are more details about checksum flag https://rclone.org/docs/#c-checksum

karol-kokoszka:
Closes #3355 

---

Please make sure that:

### DIFF
--- a/pkg/scyllaclient/client_rclone.go
+++ b/pkg/scyllaclient/client_rclone.go
@@ -598,6 +598,7 @@ func (c *Client) RcloneCheckPermissions(ctx context.Context, host, remotePath st
 }
 
 // RclonePut uploads file with provided content under remotePath.
+// WARNING: This API call doesn't compare checksums. It relies on sizes only. This call cannot be used for moving sstables.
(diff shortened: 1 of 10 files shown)


pkg/scyllaclient/client_rclone.go
```diff
@@ -598,6 +598,7 @@ func (c *Client) RcloneCheckPermissions(ctx context.Context, host, remotePath st
 }
 
 // RclonePut uploads file with provided content under remotePath.
+// WARNING: This API call doesn't compare checksums. It relies on sizes only. This call cannot be used for moving sstables.
```
Michal-Leszczynski, line 601: Why is that?
I'm not saying that's not true, but I am just curious.

PR #3360, opened 2023-04-14
Tests for restoring scylla tables

Michal-Leszczynski:
This PR fixes bug in restore cdc tables blacklisting.
Moreover, it adds test for restoring Scylla tables `system_auth.* system_traces.* system_distributed.service_levels`.


Michal-Leszczynski:
The tests weren't passing because the restore destination cluster had 29 rows in `system_auth.role_permissions` and restores source cluster had only 5 of those rows (so it's strange, because usually it's the other way around). I could only reproduce it locally when running the whole test suit, but I couldn't see it when running a single test that was supposed to fail. I don't know how those additional rows got there, but after I truncated Scylla tables before populating them with test data, the tests passed on github actions.

karol-kokoszka (approved):
:+1:

PR #3361, opened 2023-04-16
Remind about resetting tombstone_gc after restore

Michal-Leszczynski:
This PR reminds user to also reset tombstone_gc mode (alongside repair) after successful restore.


Michal-Leszczynski:
The other place where user is reminded to run restore follow-up action is at displaying progress.
On the other hand, this message does not require any update, because it says:
`NOTE: all completed restore tasks require specific follow-up action - see docs for more information`.

karol-kokoszka (approved):

PR #3371, opened 2023-04-20
restore: skip mv and si restoration

### LINKED ISSUE #3368: MV and SI should be skip from the restore procedure
Restoration of materialize view and secondary indexes doesn't work with Manager 3.1.
Command fails with information 
```
Error: get resource: create restore units: validate table exists: not found
```
It's because manager checks `system_schema.tables` table against mv and si existence.
Materialize views and secondary indexes are defined in `system_schmea.view` instead.

Manager 3.1 must skip the materialize view and secondary index restoration.
Instead, end user must drop and recreate the materialize view and secondary index after he/she restored all tables.

karol-kokoszka:
Fixes #3368 

---

Please make sure that:

### DIFF
--- a/pkg/service/backup/restore.go
+++ b/pkg/service/backup/restore.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-
+	"github.com/scylladb/gocqlx/v2/qb"
 	. "github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
 )
@@ -63,6 +63,16 @@ func (s *Service) GetRestoreTarget(ctx context.Context, clusterID uuid.UUID, pro
 
 		for _, ks := range doNotRestore {
 			t.Keyspace = append(t.Keyspace, "!"+ks)
+		}
+
+		// Filter out all materialized views and secondary indexes. They are not a part of restore procedure at the moment.
+		// See https://docs.scylladb.com/stable/operating-scylla/procedures/backup-restore/restore.html#repeat-the-following-steps-for-each-node-in-the-cluster.
+		views, err := s.listAllViews(ctx, clusterID)
+		if err != nil {
+			return t, errors.Wrapf(err, "list all views of cluster %s", clusterID.String())
+		}
+		for _, viewName := range views {
+			t.Keyspace = append(t.Keyspace, "!"+viewName)
 		}
 	}
 
@@ -243,4 +253,28 @@ func (s *Service) forEachRestoredManifest(clusterID uuid.UUID, target RestoreTar
 		}
 		return s.forEachManifest(ctx, clusterID, []Location{location}, filter, f)
 	}
+}
+
+// listAllViews is the utility function that queries system_schema.views table to get list of all views created on the cluster.
+// system_schema.views contains view definitions for materialized views and for secondary indexes.
+func (s *Service) listAllViews(ctx context.Context, clusterID uuid.UUID) (views []string, err error) {
+	// Get cluster session
+	clusterSession, err := s.clusterSession(ctx, clusterID)
+	if err != nil {
+		return nil, errors.Wrap(err, "get CQL cluster session")
+	}
+
+	iter := qb.Select("system_schema.views").
+		Columns("keyspace_name", "view_name").
+		Query(clusterSession).Iter()
+	defer func() {
+		err = iter.Close()
+	}()
+
+	var keyspace, view string
+	for iter.Scan(&keyspace, &view) {
+		views = append(views, keyspace+"."+view)
+	}
+
+	return
 }
--- a/pkg/service/backup/service_backup_restore_integration_test.go
+++ b/pkg/service/backup/service_backup_restore_integration_test.go
@@ -335,21 +335,40 @@ func TestRestoreTablesSmokeIntegration(t *testing.T) {
 		testParallel  = 3
 	)
 
-	target := RestoreTarget{
-		Location: []Location{
-			{
-				DC:       "dc1",
-				Provider: S3,
-				Path:     testBucket,
-			},
+	location := []Location{
+		{
+			DC:       "dc1",
+			Provider: S3,
+			Path:     testBucket,
 		},
-		Keyspace:      []string{testKeyspace},
-		BatchSize:     testBatchSize,
-		Parallel:      testParallel,
-		RestoreTables: true,
 	}
+	locStr, err := json.Marshal(location)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tablesTargetProps := fmt.Sprintf(`
+{
+  "location": %s,
+  "keyspace": [
+    "%s"
+  ],
+  "snapshot_tag": "sm_20060102150405UTC",
+  "batch_size": %d,
+  "parallel": %d,
+  "restore_tables": true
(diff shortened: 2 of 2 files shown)


Michal-Leszczynski:
I don't really like the fact that we have to "if-out" MV and SI in different places of the code (creating restore units, iterating over filesMeta). I would say that this type of blacklisting is the responsibility of `RestoreTarget.Keyspace` field, so maybe we should consider blacklisting MV and SI in the same way? This would require us to iterate over all rows in `system_schema.views` in the `GetRestoreTarget`, but that shouldn't be much of a problem.

What do you think about that approach?

Michal-Leszczynski:
@karol-kokoszka ^^

karol-kokoszka:
Yes, it make sense to include that into RestoreTarget.
I just made a change in my PR locally, but seems it's not enough to add the view to doNotRestore.

Michal-Leszczynski:
What do you mean by that?
If this is about failing `restoreTablesSmokeIntegration`, then this test should be adjusted, because we skip the `getRestoreTarget` there, but it cannot be skipped in the normal workflow.

karol-kokoszka:
@Michal-Leszczynski You are right !
It was enough to update GetRestoreTarget.
Test is updated.

pkg/service/backup/restore.go
```diff
@@ -64,6 +66,14 @@ func (s *Service) GetRestoreTarget(ctx context.Context, clusterID uuid.UUID, pro
...
 			t.Keyspace = append(t.Keyspace, "!"+ks)
 		}
+
+		views, err := s.listAllViews(ctx, clusterID)
+		if err != nil {
+			return t, errors.Wrapf(err, "list all views of cluster %s", clusterID.String())
+		}
+		for _, viewName := range views {
+			t.Keyspace = append(t.Keyspace, "!"+viewName)
+		}
```
Michal-Leszczynski, line 76: Maybe add a comment here why do not support restoration of MV and SI.

pkg/service/backup/restore.go
```diff
@@ -244,3 +254,29 @@ func (s *Service) forEachRestoredManifest(clusterID uuid.UUID, target RestoreTar
 		return s.forEachManifest(ctx, clusterID, []Location{location}, filter, f)
 	}
 }
+
+// listAllViews is the utility function that queries system_schema.views table to get list of all views created on the cluster.
+func (s *Service) listAllViews(ctx context.Context, clusterID uuid.UUID) (views []string, err error) {
```
Michal-Leszczynski, line 259: Maybe it's worth to mention that the `system_schema.views` also contains indexes?
It wasn't clear for me why we only query this table and not the `system_schema.indexes`.

pkg/service/backup/restore.go
```diff
@@ -244,3 +254,29 @@ func (s *Service) forEachRestoredManifest(clusterID uuid.UUID, target RestoreTar
...
+			err = multierr.Append(err, err2)
+		}
+	}()
+
+	var keyspace, view string
+	for iter.Scan(&keyspace, &view) {
+		views = append(views, keyspace+"."+view)
+	}
+
+	return
```
Michal-Leszczynski, line 281: What is the need for this defer and error appending?
If we get to the point of defer, err always equals nil.
Wouldn't it be nicer to get rid of named return and write:
```go
	var keyspace, view string
	for iter.Scan(&keyspace, &view) {
		views = append(views, keyspace+"."+view)
	}

	return iter.Close()
```

pkg/service/backup/service_backup_restore_integration_test.go
```diff
@@ -335,21 +335,40 @@ func TestRestoreTablesSmokeIntegration(t *testing.T) {
...
+{
+  "location": %s,
+  "keyspace": [
+    "%s"
+  ],
+  "snapshot_tag": "sm_20060102150405UTC",
+  "batch_size": %d,
+  "parallel": %d,
+  "restore_tables": true
+}`, locStr, testKeyspace, testBatchSize, testParallel)
```
Michal-Leszczynski, line 359: I really like the idea to store restore parameters as json, so that we can properly use `GetRestoreTarget` in the tests!

But it could be the case for more tests than just the `testRestoreTablesSmokeIntegration`! Moreover, it does not feel good that we have to hardcode some random snapshot_tag just for the `GetRestoreTarget` to pass and than swap it with an actual one. 
For those reasons, maybe we could write a function, which could take restore task params and convert them to a json accepted by the `GetRestoreTarget`. Maybe this function could accept `RestoreTarget` for simplicity? Then, this function should be used in all restore tests which would also make them more robust.

pkg/service/backup/service_backup_restore_integration_test.go
```diff
@@ -362,40 +381,55 @@ func TestRestoreSchemaSmokeIntegration(t *testing.T) {
...
 		srcClientCfg = scyllaclient.TestConfig(ManagedSecondClusterHosts(), AgentAuthToken())
 		mgrSession   = CreateScyllaManagerDBSession(t)
-		dstSession   = CreateSessionAndDropAllKeyspaces(t, ManagedClusterHosts())
-		srcSession   = CreateSessionAndDropAllKeyspaces(t, ManagedSecondClusterHosts())
-		dstH         = newRestoreTestHelper(t, mgrSession, cfg, target.Location[0], nil)
-		srcH         = newRestoreTestHelper(t, mgrSession, cfg, target.Location[0], &srcClientCfg)
+		dstH         = newRestoreTestHelper(t, mgrSession, cfg, location, nil)
+		srcH         = newRestoreTestHelper(t, mgrSession, cfg, location, &srcClientCfg)
+		dstSession   gocqlx.Session
+		srcSession   gocqlx.Session
```
Michal-Leszczynski, line 416: Maybe it's worth to leave the basic `restoreTablesSmokeIntegration` test and write a brand new one for this case? This would perhaps allow us to avoid the ifology in `smokeRestore` function in favor of creating another one.

Michal-Leszczynski (changes requested):

pkg/service/backup/restore.go
karol-kokoszka: Index actually is a view that shows the table ordered according to the index definition.
But I can add comment about that anyway.

pkg/service/backup/service_backup_restore_integration_test.go
karol-kokoszka: Sure, I can remove snapshot_tag from the payload definition.
But I'm not gonna refactor other tests now.

Please propose follow up PR if you see it as something really helpfull and needed.

pkg/service/backup/service_backup_restore_integration_test.go
karol-kokoszka: I don't think it make sense to leave previous restoreTablesSmokeIntegration.
It did almost the same what the current one does.
The difference is that the current one restores the schema the "universal restore" way + current one creates materialized view + secondary index in backup.

pkg/service/backup/service_backup_restore_integration_test.go
karol-kokoszka: UPDATE: I cannot remove snapshot_tag.
This field cannot be empty https://github.com/scylladb/scylla-manager/blob/803d2919ccbca134cfe9fc3437cd88a642abd9a1/pkg/service/backup/restore_model.go#L23

Test fails with "not a Scylla Manager snapshot tag, expected format is sm_20060102150405UTC" on GetRestoreTarget() call if the field is empty.

PR #3378, opened 2023-04-25
Refactor restore tests

Michal-Leszczynski:
This PR removes unnecessary clutter from restore tests.


karol-kokoszka (approved):
nice ! :+1:

Issue #3381, opened 2023-04-28
Separate restore and backup pkg


Michal-Leszczynski:
Currently the code for backup and restore lives in the same package (backup). They share some small amount of code, so it was convenient to write it that way, but it has its own drawbacks (e.g. structures specific for restore had to be named like `RestoreTarget`, `RestoreUnits` instead of `Target`, `Units` like in other pkgs).
Moreover, right now integration tests are divided by pkg, so we always run backup and restore tests together.
We could benefit from splitting them (it would decrease the time needed for gh actions to finish, ref #3372).

I suggest that we should separate those pkgs.

Michal-Leszczynski:
@karol-kokoszka What do you think about that?

karol-kokoszka:
@Michal-Leszczynski Brilliant ! 
Let's do that. But outside of 3.1 scope and outside of 3.1.1 scope.
Let me tag it with 3.2.

PR #3383, opened 2023-05-04
Run integration tests in parallel in github actions

### LINKED ISSUE #3372: Split sanity-check/integration-tests per pkg-integration-test
Right now, integration tests per package are executed one by one.
It can be parallelized and integration tests check should be split per package.
See https://github.com/scylladb/scylla-manager/blob/803d2919ccbca134cfe9fc3437cd88a642abd9a1/Makefile#L89-L100

Please split Sanity-check into two:
- Sanity-checks (that contains make lint + make unit-test)
- Integration-tests (that contains make pkg_integration_tests X number of these package tests)

Michal-Leszczynski:
This PR refactors environment set up, so that it's cleaner and reusable.
It also makes github actions run each pkg test in a separate job in parallel. 
In order to further improve performance:
- backup and restore tests have been separated into different jobs (even though they live in the same pkg #3381)
- restore tables tests and restore schema tests have been further divided into different jobs (because both of them take huge amount of time)
- tests for smaller packages have been merged into one job (we don't need to set up env for each of them separately)

Fixes #3372

### DIFF
--- a/.github/actions/test-setup/action.yml
+++ b/.github/actions/test-setup/action.yml
@@ -1,0 +1,32 @@
+name: "Set up testing dependencies"
+description: "Set up GO, package dependencies and install-dependencies"
+inputs:
+  start-dev-env:
+    description: "Should this action run 'make start-dev-env'"
+    required: false
+    default: 'true'
+runs:
+  using: "composite"
+  steps:
+    - name: Set up Go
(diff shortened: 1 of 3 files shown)


karol-kokoszka (approved):
:+1:

PR #3384, opened 2023-05-05
Add properties display to sctool tasks

### LINKED ISSUE #3373: Scylla Manager Tasks visibility - aggregate information
In version < 3.0 after "sctool task list", we were able to see all task list with information's such as Keyspace for the task was created. From version > 3.0 we have to check each task one by one to see additional parameters eg for repairs like:
intensity/paraller/keyspace --> it's very time consuming in case when we have a lot repair tasks specified. 

Maybe from one command like "sctool info repair" we can aggregate all information for all repair tasks

Michal-Leszczynski:
This PR adds an additional flag `--show-properties` into `sctool tasks` command, which adds addiitional `Properties` column to task table display. 

Fixes #3373

### DIFF
--- a/docs/source/sctool/partials/sctool_tasks.yaml
+++ b/docs/source/sctool/partials/sctool_tasks.yaml
@@ -23,6 +23,10 @@ options:
 - name: show-ids
   default_value: "false"
   usage: "Always display task UUID, do not show task names. \n"
+- name: show-properties
+  default_value: "false"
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
My only concern is that adding a potentially  wide column might damage readability in some cases.

karol-kokoszka (approved):
:+1:

PR #3390, opened 2023-05-13
Improve sctool tasks/info readability

### LINKED ISSUE #3333: Bring back output from `sctool task list` in 3.1.x
This is the ghost issue.
Real issue is here https://github.com/scylladb/scylla-enterprise/issues/2780#issuecomment-1486372351

Michal-Leszczynski:
This PR changes 4 main things:
- show full next activation date (not just delta) in `sctool tasks`
- show next activation date in `sctool info` even for non-standard cron (e.g. `--interval '7d'`, `--cron '@every 2h'`)
- log errors in `parallel.Run` instead of appending them (return only 1 error)
- parse XML backed errors before returning them to SM

Fixes #3333

### DIFF
--- a/pkg/command/info/cmd.go
+++ b/pkg/command/info/cmd.go
@@ -58,13 +58,16 @@ func (cmd *command) run(args []string) error {
 
 	w := cmd.OutOrStdout()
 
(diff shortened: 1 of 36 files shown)


karol-kokoszka (approved):
:+1:

PR #3391, opened 2023-05-18
dist/Makefile: switch to `--clean` instead of `--rm-dist`

yaronkaikov:
During local build i noticed the following warning
```
• DEPRECATED: --rm-dist was deprecated in favor of --clean, check https://goreleaser.com/deprecations#-rm-dist for more details
• DEPRECATED: `archives.rlcp` will be the default soon, check https://goreleaser.com/deprecations#archivesrlcp for more info
• DEPRECATED: `archives.replacements` should not be used anymore, check https://goreleaser.com/deprecations#archivesreplacements for more info
• DEPRECATED: `nfpms.replacements` should not be used anymore, check https://goreleaser.com/deprecations#nfpmsreplacements for more info
• DEPRECATED: `nfpms.replacements` should not be used anymore, check https://goreleaser.com/deprecations#nfpmsreplacements for more info
• DEPRECATED: `nfpms.replacements` should not be used anymore, check https://goreleaser.com/deprecations#nfpmsreplacements for more info
```

when running `make snapshot` (in master), the default template is `{{ .Version }}-SNAPSHOT-{{.ShortCommit}}` , the `.ShortCommit` is redundant (already part of `.Version`), so removing it

PR #3392, opened 2023-05-18
dist/.goreleaser: fix deb packages prefixes

yaronkaikov:
Following the changed made in #3391, the manager build failed since it was not aligned with Scylla-pkg.

Reverting those changes only for relevant packages

### DIFF
--- a/dist/.goreleaser.yaml
+++ b/dist/.goreleaser.yaml
@@ -66,7 +66,7 @@ builds:
 
 archives:
   - name_template: >-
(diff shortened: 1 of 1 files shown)


yaronkaikov:
Verified with https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/releng-testing/job/manager-build/576/

karol-kokoszka (approved):
:+1:

PR #3406, opened 2023-06-04
Restore: Test user permissions

Michal-Leszczynski:
This PR refactors restore tests so that restore is executed with CQL credentials that have only necessary permissions.


karol-kokoszka (approved):
:+1: and sorry for the late approve, there are some conflicts now as the compose file has changed.

karol-kokoszka (approved):

PR #3407, opened 2023-06-04
Docs: Update restore tables prerequisites

### LINKED ISSUE #3402: No documentation about required permissions for the Manager CQL role
We're creating the manager user in order to get the CQL schema backup in plaintext:

>Optionally you can provide CQL credentials to the cluster with `--username` and `--password` flags. This enables [CQL query based health check](https://manager.docs.scylladb.com/stable/health-check.html#cql-query-health-check) compared to [credentials agnostic health check](https://manager.docs.scylladb.com/stable/health-check.html#credentials-agnostic-health-check) if you do not specify the credentials. This also enables CQL schema backup in text format, which isn’t performed if credentials aren’t provided

This permission set is currently empty, but it does suffice to ensure that our schema is backed up:

```
cassandra@cqlsh> LIST ALL OF manager;

 role | username | resource | permission
------+----------+----------+------------

(0 rows)
```

However, when attempting a restore through `sctool restore --restore-tables`, we are seeing the following error, on a blank, new cluster. The `--restore-schema` was successful:

```
sctool -c au-test progress restore/restore-tables
Run:            b9de1057-fea1-11ed-9a88-027a9a47800a
Status:         ERROR
Cause:          restore data: disable gc_grace_seconds: User manager has no ALTER permission on <table TABLENAME> or any of its parents
```

There seems to be a lack of documentation around what exact permissions the CQL role used to perform the `sctool restore` requires. Could this please be clarified?

Michal-Leszczynski:
The documentation of SM `sctool restore --restore-tables` was lacking information about required permissions.
Fixes #3402

### DIFF
--- a/docs/source/restore/restore-tables.rst
+++ b/docs/source/restore/restore-tables.rst
@@ -9,7 +9,7 @@ To restore the content of the tables (rows), use the :ref:`sctool restore <sctoo
 Prerequisites
 =============
 
-* Scylla Manager with CQL credentials to restore the destination cluster.
+* Scylla Manager requires CQL credentials with `permission to alter <https://docs.scylladb.com/stable/operating-scylla/security/authorization.html#permissions>`_ restored tables.
 
 * Restoring the content of the tables assumes that the correct schema is already present in the destination cluster.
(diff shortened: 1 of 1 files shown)


annastuchlik (commented):
LGTM

PR #3408, opened 2023-06-04
Repair: test single repair on single host

Michal-Leszczynski:
This PR adds repair test which checks if there is at most one repair executed on each host.
This test does not pass in current repair implementation, but it is a good point of reference for repair procedure refactor.


Michal-Leszczynski:
It's worth to mention that this test passes if we change limits in repair controller to allow at most one concurrent repair per host - but isn't the behavior of current repair implementation.

karol-kokoszka (approved):
:+1:

Michal-Leszczynski:
The code is now rebased on master!

Issue #3397, opened 2023-05-24
Add percents of completion metric to the restore


karol-kokoszka:
Right now we have we three metrics to observe restore behavior.
We can see:
- what is the amount of remaining bytes yet to be restore,
- what is node participating in the restore doing (downloading, load&streami, idle)
- how does manager distributes batches of data between the nodes participating in the restore

We want to add another general metric to show percents of completion of the full restore.

PR #3409, opened 2023-06-06
Add metric to report progress of the restore process

charconstpointer:
closes https://github.com/scylladb/scylla-manager/issues/3397

This PR introduces a new `progress` metric that reports progress of the restore process for a given `backup.RestoreTarget` target

Used 100MB backup to test this metric locally, here's the outcome.
Screenshot 2023-07-04 at 09 55 23

### DIFF
--- a/pkg/metrics/backup.go
+++ b/pkg/metrics/backup.go
@@ -39,7 +39,8 @@ func NewBackupMetrics() BackupMetrics {
 			batchSize: gr("Cumulative size of the batches of files taken by the host to restore the data.", "batch_size", "cluster", "host"),
 			remainingBytes: gr("Remaining bytes of backup to be restored yet.", "remaining_bytes",
 				"cluster", "snapshot_tag", "location", "dc", "node", "keyspace", "table"),
-			state: gr("Defines current state of the restore process (idle/download/load/error).", "state", "cluster", "location", "snapshot_tag", "host"),
+			state:    gr("Defines current state of the restore process (idle/download/load/error).", "state", "cluster", "location", "snapshot_tag", "host"),
+			progress: gr("Defines current progress of the restore process.", "progress", "cluster", "snapshot_tag"),
 		},
 	}
 }
@@ -126,6 +127,7 @@ type RestoreM struct {
 	batchSize      *prometheus.GaugeVec
 	remainingBytes *prometheus.GaugeVec
 	state          *prometheus.GaugeVec
+	progress       *prometheus.GaugeVec
 }
 
 func (rm RestoreM) all() []prometheus.Collector {
@@ -133,6 +135,7 @@ func (rm RestoreM) all() []prometheus.Collector {
 		rm.batchSize,
 		rm.remainingBytes,
 		rm.state,
+		rm.progress,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


Michal-Leszczynski:
I understood that #3397 requires to add total restore progress to metrics, so just one float for the whole restore procedure. Adding progress information per node/keyspace/table basis increases granularity, but it does not allow to calculate the total restore progress.
Can you clarify @karol-kokoszka?

pkg/service/backup/restore_worker_index.go
```diff
@@ -194,11 +194,27 @@ func (w *indexWorker) workFunc(ctx context.Context, run *RestoreRun, target Rest
...
 			restoredBytes := pr.Downloaded + pr.Skipped + pr.VersionedProgress
 			w.metrics.DecreaseRemainingBytes(w.ClusterID, target.SnapshotTag, w.location, w.miwc.DC, w.miwc.NodeID,
 				pr.Keyspace, pr.Table, restoredBytes)
 
+			// Calculate restore progress in percents
+			var progress float64
+			if fm.Size == 0 {
+				progress = 100
+			} else {
+				progress = float64(restoredBytes) / float64(fm.Size) * 100
```
karol-kokoszka, line 206: @charconstpointer this formula won't work.

`restoredBytes` defines how much of the data (in the context of location, dc, node, keyspace and table) has been restored in the **current** chunk of the restore procedure.
Chunk is just a couple of SSTable files sent to a particular node that calls load&stream later on.

`fm.Size` defines the size of all SSTables from the location.

pkg/service/backup/restore_worker_index.go
```diff
@@ -194,11 +194,27 @@ func (w *indexWorker) workFunc(ctx context.Context, run *RestoreRun, target Rest
...
+			}
```
karol-kokoszka, line 207: @charconstpointer this formula won't work.
`restoredBytes` is just a chunk of restored keyspace/table from backup location.

You must remember the progress value for current index.
To do it, please just add `progress` var to `indexWorker` struct, then 
`w.progress += float64(restoredBytes) / float64(fm.Size) * 100`

pkg/metrics/backup.go
```diff
@@ -195,6 +198,31 @@ func (rm RestoreM) DecreaseRemainingBytes(clusterID uuid.UUID, snapshotTag strin
 	rm.remainingBytes.With(l).Sub(float64(restoredBytes))
 }
 
+// RestoreLabels is a set of labels for restore metrics.
+type RestoreLabels struct {
```
karol-kokoszka, line 202: Nice.
The same set of labels is used by remaining_bytes metric.
Please update method signature to accept `RestoreLabels` instead of `clusterID uuid.UUID, snapshotTag string, location backupspec.Location,
	dc, node, keyspace, table string,`

pkg/service/backup/restore_worker_index.go
charconstpointer: now that I think about it, `w.progress` will exceed 100% do we have to divide it by `len(w.hosts)`?

charconstpointer: Applied the propsed changes, now it reports 0-100 per unique metric label set
Screenshot 2023-06-28 at 13 55 56

karol-kokoszka: As per our slack conversation.
Having access to the size of the data to restore + `restored_bytes` var you are able to track the progress.
Accumulated progress must be shared between all hosts that participate in the restore.

charconstpointer: Seems like the workaround we've discussed works fine
Screenshot 2023-07-03 at 11 33 20

karol-kokoszka (approved):
:+1: I just removed table, keyspace and location labels

Issue #3400, opened 2023-05-25
docs: Issue in page ScyllaDB Manager Upgrade


tomer-sandler:
I would like to report an issue in page https://manager.docs.scylladb.com/branch-3.1/upgrade/index
It's relevant also for SM 3.0 upgrade page

### Problems
1. Centos/RHEL and Debian/Ubuntu show the same repo
![image](https://github.com/scylladb/scylla-manager/assets/25585995/54348dcf-62e2-411b-ba8b-85cabef16919)

2. SM 3.1 upgrade page shows the instructions of SM 3.0 upgrade

###  Suggest a fix
1. Correct repo/pkgs for SM 3.1 for Ubuntu 20.04
https://www.scylladb.com/download/?platform=ubuntu-20.04&version=manager-3.1#manager
![image](https://github.com/scylladb/scylla-manager/assets/25585995/f654d00b-3925-4d79-a62c-9729d1338dd1)

2. Correct repo/ pkgs for SM 3.0 for Ubuntu 20.04
https://www.scylladb.com/download/?platform=ubuntu-20.04&version=manager-3.0#manager
![image](https://github.com/scylladb/scylla-manager/assets/25585995/bd067580-1663-418b-bcda-20a8dc0aea89)

3. We should have upgrade docs matrix for various OS (like we have for Scylla Core)
  - RHEL/CentOS
  - Ubuntu 16/18/20/22

annastuchlik:
Thanks for reporting. I've added a [separate issue](https://github.com/scylladb/scylla-manager/issues/3401) to add a matrix.

annastuchlik:
I started using the procedure, and it's totally pointless to include all links - they are only used to illustrate the outcome of the commands. 
The step is about the command to display the contents of ``scylla-manager.repo`` (the content may change between the releases, so the max we can do is to document an example).

tomer-sandler:
Our customers go to Scylla Manager upgrade/fresh install for Ubuntu (Debian) and get instructions for CentOS/RHEL.
This was reported by 3 different customers, thus we should definitely do something about it @annastuchlik 

![image](https://github.com/scylladb/scylla-manager/assets/25585995/7110ff9b-b24e-4916-85f8-d6c29108a0c5)

tzach:
> Centos: https://downloads.scylladb.com/rpm/centos/scylladb-manager-3.0
> Ubuntu: http://downloads.scylladb.com/deb/ubuntu/scylladb-manager-3.0
[...]

PR #3410, opened 2023-06-06
doc: clarify the Manager upgrade instructions

annastuchlik:
Fixes https://github.com/scylladb/scylla-manager/issues/3400

This commit updates the information about displaying `scylla-manager.repo` during upgrade.

### DIFF
--- a/docs/source/upgrade/index.rst
+++ b/docs/source/upgrade/index.rst
@@ -87,29 +87,40 @@ It should have a status of *“Active: inactive (dead)”*.
 Upgrade the ScyllaDB Manager Server and Client 
 ------------------------------------------------
 
-**Before you Begin**
+#. **On the Manager server**, update the Manager repo file. Go to ScyllaDB Manager in 
+   the `ScyllaDB Download Center <https://www.scylladb.com/download/#manager>`_, 
+   and select your platform and the version to which you want to upgrade to display the relevant command.
+   
+   The following examples show how to update the repo for Manager 3.1:
 
 Confirm that the settings for ``scylla-manager.repo`` are correct.
 
-**On the Manager Server** display the contents of the scylla-manager.repo and confirm the version displayed is the desired version, in this example 3.0.
+      .. group-tab:: Example for Centos
 
-CentOS, Red Hat:
+           .. code:: console
+              :class: hide-copy-button
 
-.. code:: sh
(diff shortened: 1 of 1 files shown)


docs/source/upgrade/index.rst
```diff
@@ -91,27 +91,28 @@ Upgrade the ScyllaDB Manager Server and Client
 
 Confirm that the settings for ``scylla-manager.repo`` are correct.
 
-**On the Manager Server** display the contents of the scylla-manager.repo and confirm the version displayed is the desired version, in this example 3.0.
+**On the Manager Server**, display the contents of ``scylla-manager.repo`` to verify the version you're currently running:
```
annastuchlik, line 94: Please verify this change. I THINK this is the reason why we display scylla-manager.repo, but perhaps I'm wrong. 
The previous instruction is not helpful.

docs/source/upgrade/index.rst
```diff
@@ -91,27 +91,28 @@ Upgrade the ScyllaDB Manager Server and Client
 
 Confirm that the settings for ``scylla-manager.repo`` are correct.
```
annastuchlik, line 92: What does it mean that the settings are correct? What if they are not?
Does it refer to the following step, i.e., "confirm the settings are correct"="check if the version is as expected"?

I'd like to clarify this section, but I'm not sure how.

annastuchlik:
@roydahan @karol-kokoszka This is a minor PR, but I need your input - please see my comments above.

annastuchlik:
@karol-kokoszka @roydahan This PR fixes a customer-related issue. Please help by reviewing the entire PR and answering my questions above.

docs/source/upgrade/index.rst
karol-kokoszka: If the question is why `scylla-manager.repo` was displayed in unformatted form, then yes.

docs/source/upgrade/index.rst
karol-kokoszka: Looks that we are not using `scylla-manager.repo` name in other parts of instruction.
This file keeps the definition of the repository which is used to install scylla-manager packages.

By `confirm that the settings for scylla-manager.repo are correct` I understand that the user should make sure that this repo file keep the definition of expected scylla-manager version.

Take a look here https://www.scylladb.com/download/?platform=ubuntu-22.04&version=manager-3.1#manager
In **2. Procedure  / Manager 3.1** there is the instruction of how to download the expected file.

karol-kokoszka:
@annastuchlik just one comment regarding `scylla-manager.repo` from my side.

Issue #3415, opened 2023-06-11
Possible bug in 3.1 backup retention


A-Posthuman:
Since upgrading from Manager 3.0 to 3.1 I noticed my AWS S3 usage climbing steadily. After examining it, it appears the old backups aren't being removed properly? Or something changed in how things operate, I am not clear?

I have updated my backup task to all add "--retention-days 2", but I still see older backups kept. I also tried deleting my entire S3 bucket to restart from scratch, and remade my backup task in 3.1, but still seeing this behavior.

My backup task looks like: sctool backup -c clustername -L 's3:scylla-cluster-1-snapshots' -s '2023-05-29T04:00:00Z' -i 24h --retention 2 --retention-days 2

I have run the backup validate task with the option to purge orphaned items, and it found 0 items to purge.

Here is the backup as described by the update command: sctool backup update -c clustername --dry-run

> NOTICE: dry run mode, backup is not scheduled
> 
> Data Centers:
>   - us-east
> 
> Keyspaces:
>   - system_auth all (4 tables)
>   - system_distributed all (4 tables)
>   - system_distributed_everywhere all (1 table)
>   - system_traces all (5 tables)
>   - mytables all (36 tables)
>   - scylla_manager all (14 tables)
>   - system_schema all (15 tables)
> 
> Disk size: ~5.130T
> 
> Locations:
>   - s3:scylla-cluster-1-snapshots
> 
> Bandwidth Limits:
>   - 100 MiB/s
> 
> Snapshot Parallel Limits:
>   - All hosts in parallel
> 
> Upload Parallel Limits:
>   - All hosts in parallel
> 
> Retention: Last 2 backups

Here is the backup list, as you can see it is keeping way more than 2 backups/2 days: sctool backup list -c clustername

> NOTICE: this may take a while, we are reading metadata from backup location(s)
> backup/dc5c6378-ec03-40cd-a83d-65582470b10c
> Snapshots:
>   - sm_20230610040002UTC (5.126T, 3 nodes)
>   - sm_20230609040002UTC (5.122T, 3 nodes)
>   - sm_20230608040001UTC (5.114T, 3 nodes)
>   - sm_20230606105943UTC (5.104T, 3 nodes)
>   - sm_20230603040000UTC (5.053T, 3 nodes)
>   - sm_20230602040002UTC (5.039T, 3 nodes)
>   - sm_20230601040002UTC (5.021T, 3 nodes)
[...]

PR #3428, opened 2023-06-21
Fix broken schema from SM 3.1.0-rc0

Michal-Leszczynski:
As described in #3415, it's not possible to simply reinstall SM 3.1.0-rc0 into SM 3.1.0 because of failing schema migration at startup.
This PR documents the procedure of moving from broken SM 3.1.0-rc0 to the correct version.


Michal-Leszczynski:
@karol-kokoszka now SM no longer tries to fix those changes itself, but instead we give user a documented procedure describing how to fix this problem.

karol-kokoszka (approved):
:+1:

PR #3431, opened 2023-06-23
backup: general service description

### LINKED ISSUE #3421: Document `backup` procedure with Readme.md
We'd like to have documentation of the backup procedure kept in Readme.md file.
Backup and restore are a part of the same service right now (the plan is to split it in the future), so the Readme.md will be shared with restore documentation (for the moment).

The package is here https://github.com/scylladb/scylla-manager/tree/master/pkg/service/backup

The goal is to have the overview of the algorithm together with the description of service constructor and the description of the most important methods.

karol-kokoszka:
Fixes #3421 

Go to https://github.com/scylladb/scylla-manager/tree/backup_service_readme/pkg/service/backup to see the output.

---

Please make sure that:


pkg/service/backup/README.md
```diff
@@ -0,0 +1,60 @@
+# Backup service
+
+### Useful links
+- official documentation of what the backup in scylla-manager is - https://manager.docs.scylladb.com/stable/backup/index.html
+- `sctool backup` command - https://manager.docs.scylladb.com/stable/sctool/backup.html
+- scylla manager agent config file (look for `Backup general configuration.`) - https://manager.docs.scylladb.com/stable/config/scylla-manager-agent-config.html
+- schylla manager config file (look for `Backup service configuration`) - https://manager.docs.scylladb.com/stable/config/scylla-manager-config.html
```
charconstpointer, line 7:
```suggestion
- scylla manager config file (look for `Backup service configuration`) - https://manager.docs.scylladb.com/stable/config/scylla-manager-config.html
```

pkg/service/backup/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+#### Scheduling the backup ####
+
+![Scheduling the backup](scylla-manager-backup.drawio.svg)
+
+Backup service is responsible for handling backup tasks scheduled either via Scylla Manager REST API or via sctool (CLI client for REST API).
+
+#### Performing the backup ####
+
+```
```
charconstpointer, line 19: you can annotate code blocks with `go` keyword to enable syntax highlighting like so https://www.markdownguide.org/extended-syntax/#syntax-highlighting

This applies to all go snippets in this PR

pkg/service/backup/README.md
```diff
@@ -0,0 +1,60 @@
...
+func NewService(session gocqlx.Session, config Config, metrics metrics.BackupMetrics, clusterName ClusterNameFunc, scyllaClient scyllaclient.ProviderFunc,
+	clusterSession SessionFunc, logger log.Logger,
+) (*Service, error)
+```
+
+
+Entrypoint on the backup service that start the process of backing up the data is
```
charconstpointer, line 26:
```suggestion
Entrypoint on the backup service that starts the process of backing up the data is
```

pkg/service/backup/README.md
```diff
@@ -0,0 +1,60 @@
+# Backup service
+
+### Useful links
+- official documentation of what the backup in scylla-manager is - https://manager.docs.scylladb.com/stable/backup/index.html
+- `sctool backup` command - https://manager.docs.scylladb.com/stable/sctool/backup.html
+- scylla manager agent config file (look for `Backup general configuration.`) - https://manager.docs.scylladb.com/stable/config/scylla-manager-agent-config.html
+- schylla manager config file (look for `Backup service configuration`) - https://manager.docs.scylladb.com/stable/config/scylla-manager-config.html
+
+### General picture
```
charconstpointer, line 9:
```suggestion
### Overview of backup processes
```

charconstpointer (approved):
Great job Karol, happy to see our docs expanded!

Some very minor nitpicks. I'll give it another look later on

pkg/service/backup/README.md
charconstpointer, line 7: Maybe start this uppercase to be consistent within the document and also decide wether to use `Scylla`

Michal-Leszczynski:
Nice!!

I'm wondering if it would be good to document each step of the backup procedure itself (e.g. snapshot, index, ..., purge).
It could contain details like how SM handles versioned files etc.

In terms of versioned files, maybe it would be good to add information about them to the [backup layout specification](https://manager.docs.scylladb.com/stable/backup/specification.html)?

Right now, the first half of this README contains more general information that's common for other services (e.g repair and restore in the future), so maybe it could be elevated to the `service` pkg instead of the `service/backup`?

karol-kokoszka:
@Michal-Leszczynski

> I'm wondering if it would be good to document each step of the backup procedure itself (e.g. snapshot, index, ..., purge).
It could contain details like how SM handles versioned files etc.

I was thinking about that, but my impression is that we would go too much into the details. Code is some sort of documentation as well. 

> In terms of versioned files, maybe it would be good to add information about them to the [backup layout specification](https://manager.docs.scylladb.com/stable/backup/specification.html)?

Yeah, I saw it as well. There is an issue for that https://github.com/scylladb/scylla-manager/issues/3432

> Right now, the first half of this README contains more general information that's common for other services (e.g repair and restore in the future), so maybe it could be elevated to the service pkg instead of the service/backup?

It's true. First sequence diagram describes the scheduler service actually. I'll add the README.md to scheduler and move this pic.

pkg/service/backup/README.md
karol-kokoszka: I'll remove first picture from here and move it to scheduler service.

dkropachev (approved):

PR #3434, opened 2023-06-23
feat(integration-tests): add ipv6 test environment

### LINKED ISSUE #3419: IPv6 testing env (new docker-compose)
We need to have a docker-compose that creates IPv6 environment.
Right now, our integration-tests are checking SM behavior against IPv4 setup only, see https://github.com/scylladb/scylla-manager/blob/master/testing/docker-compose.yaml

The idea behind this issue is to have:
- another docker-compose that creates the IPv6 env
- docker-compose for IPv6 should mirror the current compose file for IPv4 (with 2 networks see this PR https://github.com/scylladb/scylla-manager/pull/3418)
- another targets in Makefile that will trigger integration tests execution against IPv6 env (see https://github.com/scylladb/scylla-manager/blob/97ee657da22654ca08ab93ca8e31ca0c10efed08/Makefile#L105-L110)
- IPv6 integration tests must be included into GH actions https://github.com/scylladb/scylla-manager/tree/master/.github/workflows

dkropachev:
Closes #3419 
Add ipv6 test environment.
In order to target ipv6 environment you have to add `IPV6=true`, i.e.:
`IPV6=true make start-dev-env`
`IPV6=true make pkg-integration-test PKG=./pkg/ping/cqlping`
...

Honestly don't like it, so if you have better idea - please let me know


testing/Makefile
```diff
@@ -37,21 +64,36 @@ else
...
+	@until [ 4 -le $$($(NODETOOL) status | grep -c "UN") ]; do echo -n "."; sleep 2; done ; echo ""
+	@echo "==> Second nodes from DC1 and DC2 are ready"
+	@. ./.env && CURRENT_UID=$(CURRENT_UID) CURRENT_GID=$(CURRENT_GID) $(COMPOSE) -f docker-compose.yaml -f $(COMPOSE_FILE) up -d
+	@echo "==> Waiting for sm db cluster"
+	@until [ 1 -le $$($(SM_NODETOOL) status | grep -c "UN") ]; do echo -n "."; sleep 2; done ; echo ""
+	@echo "==> SM cluster is ready"
+
+	@echo "==> Waiting for rest of the cluster"
+	@until [ 6 -le $$($(NODETOOL) status | grep -c "UN") ]; do echo -n "."; sleep 2; done ; echo ""
+	@echo "==> Cluster is ready"
```
karol-kokoszka, line 84: Why to wait for dc1_node_1 and dc2_node_1 first ? For *node_2 second and the rest at the end ?
What is the idea behind it ?

karol-kokoszka:
> In order to target ipv6 environment you have to add IPV6=true, i.e.:
IPV6=true make start-dev-env
IPV6=true make pkg-integration-test PKG=./pkg/ping/cqlping
...

> Honestly don't like it, so if you have better idea - please let me know

I have nothing against it.
The only other option I see is to have different targets for IPv4 and IPv6 in the Makefile, but I'm OK with what you propose.

testing/Makefile
karol-kokoszka: Can we bring back the previous approach ?

dkropachev: Scylla bootstrapping is very fragile, in general, it is not recommended to bootstrap more than one node at a time.
Since it is brand new, empty cluster we get away with bootstraping non-seed nodes in parallel.
But not with seeds nodes, since timeout on seeds is way stricter.
Locally I have 60% chance of failing bootstrap with the way we did it before.

dkropachev:
I have accidentially merged https://github.com/scylladb/scylla-manager/pull/3436 that was based of this guy.
Let's close this one, having in mind that it is actually merged.

PR #3440, opened 2023-06-28
Don't block backup on dumping schema error

### LINKED ISSUE #3433: Scylla manager fails to backup custom UDA
Creating backup using scylla manager fails with the error

```await schema: get schema: cql keyspace traffickeyspace metadata: template: aggregate:4:11: executing "aggregate" at <ne $i 0>: error calling ne: incompatible types for comparison```

We had a UDA in our keyspace, when we dropped it and retried getting the backup, it was successful. The UDA has the following schema
```
CREATE AGGREGATE ks.average(double)
SFUNC avgstate STYPE tuple<int, double>
FINALFUNC avgfinal
INITCOND (0,0.0);

CREATE FUNCTION ks.avgfinal(
  state tuple<int, double>)
CALLED ON NULL INPUT
RETURNS double
LANGUAGE lua
as $$
    r=0
    r=state[2]
    r=r/state[1]
    return r
    $$;

CREATE FUNCTION ks.avgstate(
  state tuple<int, double>,
  val double)
CALLED ON NULL INPUT
RETURNS frozen<tuple<int, double>>
LANGUAGE lua
AS $$
  return { state[1]+1, state[2]+val }
  $$;
```

The aggregate function works as expected, only encountering the error while taking backup using syclla manager.

Scylla Manager Version : 3.1
Scylla DB Version : 5.2

### PR DESCRIPTION
**Pull request by Michal-Leszczynski**
As we do not use backed-up schema file for restoring purposes, we can always continue backup, even when this step fails.
Fixes #3433

### DIFF
--- a/pkg/service/backup/service.go
+++ b/pkg/service/backup/service.go
@@ -807,17 +807,21 @@ func (s *Service) Backup(ctx context.Context, clusterID, taskID, runID uuid.UUID
 		return nil
 	}
 	stageFunc := map[Stage]func() error{
-		StageAwaitSchema: func() error {
+		StageAwaitSchema: func() error { // nolint: unparam
 			clusterSession, err := s.clusterSession(ctx, clusterID)
 			if err != nil {
-				w.Logger.Info(ctx, "No CQL cluster session, backup of schema as CQL files would be skipped", "error", err)
+				w.Logger.Info(ctx, "No CQL cluster session, backup of schema as CQL files will be skipped", "error", err)
 				return nil
 			}
 			defer clusterSession.Close()
 
 			w.AwaitSchemaAgreement(ctx, clusterSession)
 
-			return w.DumpSchema(ctx, clusterSession)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/service/backup/service.go [hunk 1]
```diff
@@ -807,17 +807,21 @@ func (s *Service) Backup(ctx context.Context, clusterID, taskID, runID uuid.UUID
 		return nil
 	}
 	stageFunc := map[Stage]func() error{
-		StageAwaitSchema: func() error {
+		StageAwaitSchema: func() error { // nolint: unparam
```
karol-kokoszka, line 810: Why `nolint: unparam` ?

karol-kokoszka (approved):

pkg/service/backup/service.go
Michal-Leszczynski: Otherwise we get the following error while running `make check`:
```
pkg/service/backup/service.go:810:28: (*Service).Backup$3 - result 0 (error) is always nil (unparam)
                StageAwaitSchema: func() error {
                                         ^
```

PR #3442, opened 2023-06-28
Document sstable versioning

### LINKED ISSUE #3432: Backup specification is not updated with versioning added in 3.1
https://manager.docs.scylladb.com/stable/backup/specification.html is missing info about the versioning.

Here is the issue that requested to add versioning https://github.com/scylladb/scylla-manager/issues/3288
... and here is the PR https://github.com/scylladb/scylla-manager/pull/3321

Michal-Leszczynski:
This PR adds missing information about backup sstable versioning.
Fixes #3432

### DIFF
--- a/docs/source/backup/specification.rst
+++ b/docs/source/backup/specification.rst
@@ -138,7 +138,16 @@ Please find below a part listing of sst directory showing a single table of a si
                        │                       ├── mc-5-big-Scylla.db
                        │                       ├── mc-5-big-Statistics.db
                        │                       ├── mc-5-big-Summary.db
-                       │                       └── mc-5-big-TOC.txt
+                       │                       ├── mc-5-big-TOC.txt
+                       │                       ├── mc-5-big-CompressionInfo.db.sm_20210809095541UTC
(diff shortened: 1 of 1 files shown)


docs/source/backup/specification.rst
```diff
@@ -155,6 +164,15 @@ Under the node directory each table version has it's directory, the path is stru
 The directory contains all the table files.
 Some files may be used in more than one backup.
 
+You may also find that some of those files have the snapshot tag suffix (e.g. ``.sm_20210809095541UTC``).
+Even though sstables (data files) are immutable by nature, it's sometimes possible to have two different sstables originating
+from the same table and with the same node ID. In this case, older versions of sstable are appended with the snapshot tag
```
tzach, line 169: Worthwhile mentioning this can only happen if you replace a node. 
It will give the issue context.

Also, once https://github.com/scylladb/scylladb/issues/6403 is fixed, two nodes won't have the same ID (just FYI, there is no need to include it here)

karol-kokoszka (approved):

annastuchlik (commented):
@Michal-Leszczynski Thanks for adding the docs!

PR #3447, opened 2023-07-03
Validate that all units are present after resume

Michal-Leszczynski:
We didn't explicitly validate that all tables are still present in the restore destination cluster after resume. 
It wasn't big of an issue (restore would eventually run into deleted table and return some error), but it's always better to check it before resuming the restore and not in the middle of it.

Related [issue](https://github.com/scylladb/scylla-dtest/issues/3286) from dtests.


karol-kokoszka (approved):
:+1:

PR #3448, opened 2023-07-03
Fix restore progress display with nil progress

Michal-Leszczynski:
The following [dtest](https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-dtest/lastCompletedBuild/testReport/manager_restore_tests/TestScyllaMgmtRestore/test_restore_alter_batch_size/) was querying restore progress right after task creation, which resulted in getting nil RestoreProgress.Progress, which caused panic when displaying restore progress. This PR fixes this behavior.


karol-kokoszka (approved):
:+1:

Issue #3458, opened 2023-07-11
Prometheus metric cardinality issue


dkropachev:
We are using snapshot id as a label for some prometheus metrics:
```
// SetRemainingBytes sets restore "remaining_bytes" metric.
func (rm RestoreM) SetRemainingBytes(labels RestoreBytesLabels, remainingBytes int64) {
	l := prometheus.Labels{
		"cluster":      labels.ClusterID,
		"snapshot_tag": labels.SnapshotTag,
		"location":     labels.Location,
		"dc":           labels.DC,
		"node":         labels.Node,
		"keyspace":     labels.Keyspace,
		"table":        labels.Table,
	}
	rm.remainingBytes.With(l).Set(float64(remainingBytes))
}

```

```
// DecreaseRemainingBytes decreases restore "remaining_bytes" metric.
func (rm RestoreM) DecreaseRemainingBytes(labels RestoreBytesLabels, restoredBytes int64) {
	l := prometheus.Labels{
		"cluster":      labels.ClusterID,
		"snapshot_tag": labels.SnapshotTag,
		"location":     labels.Location,
		"dc":           labels.DC,
		"node":         labels.Node,
		"keyspace":     labels.Keyspace,
		"table":        labels.Table,
	}
	rm.remainingBytes.With(l).Sub(float64(restoredBytes))
}
```

Which is destine to blow [metric cardinality](https://grafana.com/blog/2022/02/15/what-are-cardinality-spikes-and-why-do-they-matter/) indefinitely since `shapshot_tag` is unique for every given backup.

My understanding that this should address this issue by removing `snapshot_tag` from the metrics, given the fact that in all ocasions it is enough information in rest labels to figure which repair/backup metric related to.

dkropachev:
@amnonh , is it something that we see in production, is there a occasions when prometheus metrics from `scylla-manager` went off hand?

dkropachev:
@Michal-Leszczynski, could you please share your thoughts on it

Michal-Leszczynski:
I don't know how exactly are those metrics utilized by users/monitoring/cloud, but I would say that there is no need to include snapshot tag as a label here.

PR #3462, opened 2023-07-11
fix(backup): remove snapshot_id from metric labels

dkropachev:
Having metric vector with label that is getting new value every giving time is blowing cardinality and causes insert slowdown and failures on prometheus side, it also increases bill you paid for cloud monitoring. 

https://grafana.com/blog/2022/02/15/what-are-cardinality-spikes-and-why-do-they-matter/

Closes https://github.com/scylladb/scylla-manager/issues/3458

### DIFF
--- a/pkg/metrics/backup.go
+++ b/pkg/metrics/backup.go
@@ -36,11 +36,13 @@ func NewBackupMetrics() BackupMetrics {
 				"purge_deleted_files", "cluster", "host"),
 		},
 		Restore: RestoreM{
-			batchSize: gr("Cumulative size of the batches of files taken by the host to restore the data.", "batch_size", "cluster", "host"),
+			batchSize: gr("Cumulative size of the batches of files taken by the host to restore the data.", "batch_size",
+				"cluster", "host"),
 			remainingBytes: gr("Remaining bytes of backup to be restored yet.", "remaining_bytes",
-				"cluster", "snapshot_tag", "location", "dc", "node", "keyspace", "table"),
-			state:    gr("Defines current state of the restore process (idle/download/load/error).", "state", "cluster", "location", "snapshot_tag", "host"),
-			progress: gr("Defines current progress of the restore process.", "progress", "cluster", "snapshot_tag"),
+				"cluster", "location", "dc", "node", "keyspace", "table"),
+			state: gr("Defines current state of the restore process (idle/download/load/error).", "state",
+				"cluster", "location", "host"),
+			progress: gr("Defines current progress of the restore process.", "progress", "cluster"),
 		},
 	}
 }
@@ -169,38 +171,35 @@ func (rm RestoreM) DecreaseBatchSize(clusterID uuid.UUID, host string, size int6
 // SetRemainingBytes sets restore "remaining_bytes" metric.
 func (rm RestoreM) SetRemainingBytes(labels RestoreBytesLabels, remainingBytes int64) {
 	l := prometheus.Labels{
-		"cluster":      labels.ClusterID,
-		"snapshot_tag": labels.SnapshotTag,
-		"location":     labels.Location,
-		"dc":           labels.DC,
-		"node":         labels.Node,
-		"keyspace":     labels.Keyspace,
-		"table":        labels.Table,
+		"cluster":  labels.ClusterID,
+		"location": labels.Location,
+		"dc":       labels.DC,
+		"node":     labels.Node,
+		"keyspace": labels.Keyspace,
+		"table":    labels.Table,
 	}
 	rm.remainingBytes.With(l).Set(float64(remainingBytes))
 }
 
 // RestoreBytesLabels is a set of labels for restore metrics.
 type RestoreBytesLabels struct {
-	ClusterID   string
-	SnapshotTag string
-	Location    string
-	DC          string
-	Node        string
-	Keyspace    string
-	Table       string
+	ClusterID string
+	Location  string
+	DC        string
+	Node      string
+	Keyspace  string
+	Table     string
 }
 
 // DecreaseRemainingBytes decreases restore "remaining_bytes" metric.
 func (rm RestoreM) DecreaseRemainingBytes(labels RestoreBytesLabels, restoredBytes int64) {
 	l := prometheus.Labels{
-		"cluster":      labels.ClusterID,
-		"snapshot_tag": labels.SnapshotTag,
-		"location":     labels.Location,
-		"dc":           labels.DC,
-		"node":         labels.Node,
-		"keyspace":     labels.Keyspace,
-		"table":        labels.Table,
+		"cluster":  labels.ClusterID,
+		"location": labels.Location,
+		"dc":       labels.DC,
+		"node":     labels.Node,
+		"keyspace": labels.Keyspace,
+		"table":    labels.Table,
 	}
 	rm.remainingBytes.With(l).Sub(float64(restoredBytes))
 }
@@ -208,16 +207,14 @@ func (rm RestoreM) DecreaseRemainingBytes(labels RestoreBytesLabels, restoredByt
 // RestoreProgressLabels is a set of labels for restore "progress" metric.
 // RestorePrgoressLabels does not contain DC and Node labels since we only care about global restore progress.
 type RestoreProgressLabels struct {
-	ClusterID   string
-	SnapshotTag string
+	ClusterID string
 }
 
 // SetProgress sets restore "progress" metric,
 // progress should be a value between 0 and 100, that indicates global restore progress.
 func (rm RestoreM) SetProgress(labels RestoreProgressLabels, progress float64) {
 	l := prometheus.Labels{
-		"cluster":      labels.ClusterID,
-		"snapshot_tag": labels.SnapshotTag,
+		"cluster": labels.ClusterID,
 	}
 	rm.progress.With(l).Set(progress)
 }
@@ -237,12 +234,11 @@ const (
 )
 
 // SetRestoreState sets restore "state" metric.
-func (rm RestoreM) SetRestoreState(clusterID uuid.UUID, location backupspec.Location, snapshotTag, host string, state RestoreState) {
+func (rm RestoreM) SetRestoreState(clusterID uuid.UUID, location backupspec.Location, host string, state RestoreState) {
 	l := prometheus.Labels{
-		"cluster":      clusterID.String(),
-		"location":     location.String(),
-		"snapshot_tag": snapshotTag,
-		"host":         host,
+		"cluster":  clusterID.String(),
+		"location": location.String(),
+		"host":     host,
 	}
 	rm.state.With(l).Set(float64(state))
 }
--- a/pkg/service/backup/restore_worker_index.go
+++ b/pkg/service/backup/restore_worker_index.go
@@ -131,14 +131,14 @@ func (w *indexWorker) workFunc(ctx context.Context, run *RestoreRun, target Rest
 		h := &w.hosts[n]
 		defer func() {
 			if err != nil {
-				w.metrics.SetRestoreState(run.ClusterID, w.location, target.SnapshotTag, h.Host, metrics.RestoreStateError)
+				w.metrics.SetRestoreState(run.ClusterID, w.location, h.Host, metrics.RestoreStateError)
 				return
 			}
-			w.metrics.SetRestoreState(run.ClusterID, w.location, target.SnapshotTag, h.Host, metrics.RestoreStateIdle)
+			w.metrics.SetRestoreState(run.ClusterID, w.location, h.Host, metrics.RestoreStateIdle)
 		}()
 		for {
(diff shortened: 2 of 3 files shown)


Michal-Leszczynski:
Do we have a feedback from monitoring team that it is ok to remove this label?
Or at least that they are aware of this change.

karol-kokoszka:
Please not remove this label yet until it's confirmed that we kill prometheus with that.
If you want to remove the label, then please introduce another metric that will keep the snapshot_tag as a value, but it must be labelled with cluster_id then.

dkropachev:
> Please not remove this label yet until it's confirmed that we kill prometheus with that. 

It is not really necessary, you can always assume task and cluster from the information preset in rest of the labels.
Having label that change value every time is not exactly useful.

> If you want to remove the label, then please introduce another metric that will keep the snapshot_tag as a value, but it must be labelled with cluster_id then.

Not quite get it, could you please clarify it.

karol-kokoszka:
> Having metric vector with label that is getting new value every giving time is blowing cardinality and causes insert slowdown and failures on prometheus side, it also increases bill you paid for cloud monitoring.

Blowing cardinality is not just about the number of values for a single label, but about number of values for an active series.
Timeseries is considered active when (https://prometheus.io/docs/prometheus/1.8/storage/#settings-for-high-numbers-of-time-series):
> A series that has recently received new samples will have an open incomplete head chunk. Once that chunk is completely filled, or the series hasn't received samples in a while, the head chunk is closed and becomes a chunk waiting to be appended to its corresponding series file, i.e. it is waiting for persistence. After the chunk has been persisted to disk, it becomes evictable, provided it is not currently used by a query. Prometheus will evict evictable chunks from memory to satisfy the configured target heap size. A series with an open head chunk is called an active series. This is different from a memory series, which also includes series without an open head chunk but still other chunks in memory (whether waiting for persistence, used in a query, or evictable). A series without any chunks in memory may be archived, upon which it ceases to have any mandatory memory footprint.

`snapshot_tag` is added to restore metrics and it doesn't change throughout the full restore process. There is no cardinality spike. IIUC spike is when the number of timeseries increases suddenly. For these restore metrics it would happen if there are many restores of the same cluster in parallel, what is not the case.

>> If you want to remove the label, then please introduce another metric that will keep the snapshot_tag as a value, but it must be labelled with cluster_id then.

> Not quite get it, could you please clarify it.

I'm talking about another metric that would indicate that the restore of the particular snapshot is ongoing for given cluster.
Like `restore_ongoing{cluster={}}` keeping the snapshot_id value.

dkropachev:
> > Having metric vector with label that is getting new value every giving time is blowing cardinality and causes insert slowdown and failures on prometheus side, it also increases bill you paid for cloud monitoring.
> 
> Blowing cardinality is not just about the number of values for a single label, but about number of values for an active series. Timeseries is considered active when (https://prometheus.io/docs/prometheus/1.8/storage/#settings-for-high-numbers-of-time-series):
> 
> > A series that has recently received new samples will have an open incomplete head chunk. Once that chunk is completely filled, or the series hasn't received samples in a while, the head chunk is closed and becomes a chunk waiting to be appended to its corresponding series file, i.e. it is waiting for persistence. After the chunk has been persisted to disk, it becomes evictable, provided it is not currently used by a query. Prometheus will evict evictable chunks from memory to satisfy the configured target heap size. A series with an open head chunk is called an active series. This is different from a memory series, which also includes series without an open head chunk but still other chunks in memory (whether waiting for persistence, used in a query, or evictable). A series without any chunks in memory may be archived, upon which it ceases to have any mandatory memory footprint.

Not exactly, cardinality is cardinality, it does not link directly to active or non-active time series.
What happens is that high cardinality can lead to having different problems.
Only one of them is high delay on a metric write, due to high number of active time series.
Other problems, say delay on metrics read and per metric payments are still present even in scenarios when time series are not active.

> `snapshot_tag` is added to restore metrics and it doesn't change throughout the full restore process. There is no cardinality spike. IIUC spike is when the number of timeseries increases suddenly. For these restore metrics it would happen if there are many restores of the same cluster in parallel, what is not the case.
> 
> > > If you want to remove the label, then please introduce another metric that will keep the snapshot_tag as a value, but it must be labelled with cluster_id then.
> 
> > Not quite get it, could you please clarify it.
> 
> I'm talking about another metric that would indicate that the restore of the particular snapshot is ongoing for given cluster. Like `restore_ongoing{cluster={}}` keeping the snapshot_id value.

In my understanding it is misuse of the tool, `metric` is not a metric, it is a `metric vector`, when metric vector meets new value for the lable, it creates a new metric and uses it to store data in it.
As result, in what you propose, `s-m` is going to create a new metric for every given restore, with only one purpose - to make it look nice on the monitoring UI.

Since, unfortunately there is no way to have a textual metric or easily annotate a metric, i would suggest to keep it logs and if one wants to see in nice in monotiring, it would make monitoring extract these events from logs:
https://grafana.com/blog/2019/10/10/new-in-grafana-6.4-loki-annotations/
https://docs.datadoghq.com/service_management/events/

karol-kokoszka:
@dkropachev You backed this PR with the link describing what issues may appear in Grafana Cloud when there is cardinality spike. It describes the meaning of the cardinality as well. Have you read it ? 
https://grafana.com/blog/2022/02/15/what-are-cardinality-spikes-and-why-do-they-matter/

>You can have lower cardinality (1:5 label-value ratio), standard cardinality (1:80 label-value ratio), or high cardinality (1:10,000 label-value ratio). 

Overall cardinality of the `snapshot_tag` label is number of restores made on the cluster (using different snapshot_tag). I would say `snapshot_tag` is between lower cardinality and the standard cardinality.

Now, what is the cardinality spike

> Something our customers bring up a lot is sudden spikes in cardinality, which is when a metric with medium or lower cardinality suddenly transforms into a metric with high cardinality. 

> In [Grafana Cloud](https://grafana.com/products/cloud/?pg=blog&plcmt=body-txt), Grafana Cloud Metrics is billed on the number of [Active Series (and DPM](https://grafana.com/docs/grafana-cloud/fundamentals/active-series-and-dpm/?pg=blog&plcmt=body-txt)). An active series is considered active when data points have been added within the last 15 to 30 minutes. If you experience a sudden spike in cardinality that exponentially increases the number of metrics you’re sending, your bill will also increase. So keeping cardinality under control can have some real cost benefits as well.

Cardinality here is under control. We won't have spikes due to the `snapshot_tag` label. 
Besides, we don't even use the Grafana Cloud. Billing described here is not the case for us.

Now, refering to the last answer:

> Not exactly, cardinality is cardinality, it does not link directly to active or non-active time series.
What happens is that high cardinality can lead to having different problems.
Only one of them is high delay on a metric write, due to high number of active time series.
Other problems, say delay on metrics read and per metric payments are still present even in scenarios when time series are not active.

It is linked to active time series. Active time series are kept in the memory otherwise they are archived.
We don't pay for it, because we are not using Grafana Cloud.
Besides, you wrote 

> Only one of them is high delay on a metric write, due to high number of active time series.

So, it is linked to active or non-active time series or not ?

> As result, in what you propose, s-m is going to create a new metric for every given restore, with only one purpose - to make it look nice on the monitoring UI.

No, the purpose is not to look nice on the monitoring UI. The purpose is proper debugging. Especially QA tests.

I don't want to remove `snapshot_tag` label yet. I'm using it. This PR is not going to be merged.

dkropachev:
> @dkropachev You backed this PR with the link describing what issues may appear in Grafana Cloud when there is cardinality spike. It describes the meaning of the cardinality as well. Have you read it ? https://grafana.com/blog/2022/02/15/what-are-cardinality-spikes-and-why-do-they-matter/
> 
> > You can have lower cardinality (1:5 label-value ratio), standard cardinality (1:80 label-value ratio), or high cardinality (1:10,000 label-value ratio).
> 
> Overall cardinality of the `snapshot_tag` label is number of restores made on the cluster (using different snapshot_tag). I would say `snapshot_tag` is between lower cardinality and the standard cardinality.

Of course I did, there is one problem in this article, it does immiate jump from `one label cardinality`:
> To put it simply: Cardinality is the overall count of values for one label. In the example above, the label status_code has a cardinality of 5, environment has a cardinality of 2, and the overall cardinality of the metric server_responses is 10.

To a `metric cardinality`:
> Since Prometheus creates one series per label combination, you will end up with a lot of series for a single metric if you have a lot of users.

As result one might think that cardinality is a number related to one particular label, which is not true, it is a metric value, and in their low/normal/high cardinality chart they mean `metric cardinality` not `label cardinality`, which is calculated from multiplication of cardinality for all the labels on the metric.

Now let's calculate it for `remainingBytes`:
```
		"cluster":      labels.ClusterID,
		"location":     labels.Location,
		"dc":           labels.DC,
		"node":         labels.Node,
		"keyspace":     labels.Keyspace,
		"table":        labels.Table,
```
Let's do that not take into account `snapshot_tag` cardinality, to add it later to see the impact, say customer have 1 cluster with 10  keyspaces, 10 tables in each, running on 6 nodes with 2 dc and 2 backup locations, 10x10x6x2x2 = 2400.
Now, let's add `snapshot_tag` into the equasion, on every run `remainingBytes` metric cardinality is going to jump by 2400.
10 restores and you have 24000, two times more than what they call `high cardinality`.

> Now, what is the cardinality spike
> 
> > Something our customers bring up a lot is sudden spikes in cardinality, which is when a metric with medium or lower cardinality suddenly transforms into a metric with high cardinality.
> 
> > In [Grafana Cloud](https://grafana.com/products/cloud/?pg=blog&plcmt=body-txt), Grafana Cloud Metrics is billed on the number of [Active Series (and DPM](https://grafana.com/docs/grafana-cloud/fundamentals/active-series-and-dpm/?pg=blog&plcmt=body-txt)). An active series is considered active when data points have been added within the last 15 to 30 minutes. If you experience a sudden spike in cardinality that exponentially increases the number of metrics you’re sending, your bill will also increase. So keeping cardinality under control can have some real cost benefits as well.
> 
> Cardinality here is under control. We won't have spikes due to the `snapshot_tag` label. Besides, we don't even use the Grafana Cloud. Billing described here is not the case for us.

Exactly!
Grafana cloud does lot's of optimizations, it keeps active timeseries in memory to speedup writing.
While non-active time series on a disk, and charges, as of now, only for active series, while other monitoring providers are not so generous, such as `datadog` that [charges for non-active](https://docs.datadoghq.com/account_management/billing/custom_metrics/?tab=countrate):
> Your monthly billable count for custom metrics (reflected on the Usage page) is calculated by taking the total of all distinct custom metrics for each hour in a given month, and dividing it by the number of hours in the month to compute a monthly average value.

Grafana have even a tool to fight cardinality - [adaptive metrics](https://grafana.com/products/cloud/metrics/prometheus-cardinality-optimization)
But that does not mean that we should disregard cardinality issue, it means opposite, if they bill on per cardinality bases, if they have a product to fight cardinality, it means we should be more cautious about it and avoid blowing timeseries, especially indefinitely. 

> Now, refering to the last answer:
> 
> > Not exactly, cardinality is cardinality, it does not link directly to active or non-active time series.
> > What happens is that high cardinality can lead to having different problems.
> > Only one of them is high delay on a metric write, due to high number of active time series.
> > Other problems, say delay on metrics read and per metric payments are still present even in scenarios when time series are not active.
> 
> It is linked to active time series. Active time series are kept in the memory otherwise they are archived. We don't pay for it, because we are not using Grafana Cloud. Besides, you wrote
> > Only one of them is high delay on a metric write, due to high number of active time series.
> 
> So, it is linked to active or non-active time series or not ?

I undersand it could be confusing, since all over the place these terms are used interchangeably and I did that as well.
`cardinality` is a mathematical term for `number of elements in a set`. 
`metric cardinality` in general, is a number unique combinations of label values.
I want to draw your attention to the fact that `metric cardinality` is just a potential, calculation, property, while `active/non-active time series` are the state, to go from one to another there has to be an action.

In order for `metric cardinality` to manifest in `active time series` with 1:1 ratio metric vector has to be submitted with every combination of the label values that have been in initial calculation for given period.
In order for `metric cardinality` to manifest in `non-active time series` it has to be done once in retention period.

High `metric cardinality` not necesserily leads to high `active time series`, neither it necesserily leads high `non-active time series`.
In our case, as you mentioned, high `metric cardinality` won't lead to high `active time series`, but **could** lead to `non-active` time series.

> 
> > As result, in what you propose, s-m is going to create a new metric for every given restore, with only one purpose - to make it look nice on the monitoring UI.
> 
> No, the purpose is not to look nice on the monitoring UI. The purpose is proper debugging. Especially QA tests.
> 
> I don't want to remove `snapshot_tag` label yet. I'm using it. This PR is not going to be merged.

Only drawback here is that in order to see the tag you need to go to the logs, which is not a blocking thing at all.
If we set up loki properly we will get grafana annotations that make `snapshot_tag` present there, and not only it, we can extract host of events from the `s-m` logs that are currently not being shown.
Also it is not an excuse to expose us and customers to host of the problems related to infinit cardinality.

it is a calloborative process and I am ok with whatever, I gave my rationale to the best of my abilities, I am 99% certain that it is correct and reflects the reality.

PR #3464, opened 2023-07-12
fix(tests): make rest of the test parallel

dkropachev:
I noticed that some of the tests are running sequentially without reason.


karol-kokoszka (approved):
:+1:

PR #3466, opened 2023-07-12
Restore mv and si

### LINKED ISSUE #3370: Automate materialized views and secondary indexes restoration
Current procedure of cluster restore requires end user to drop Materialized Views and Secondary Indexes after schema restoration and to recreate both after the repair.

We should automate it and make it to be a part of restore procedure supported by scylla manager.
To do so:
- before table restoration, scylla-manager must check `system_schema.views` table against any existing views. 
- scylla-manager must save materialized views and secondary indexes definitions in cache + place information about them in `progress` command output of restore task
- after the whole restore procedure, materialized views and secondary indexes must be recreated

Michal-Leszczynski:
This PR implements automatic restoration of MV and SI in restore procedure.
Main points:
- `--keyspace` parameter defines ONLY base tables that should be restored (and that's why we always exclude views from it). ALL views of restored base tables are always restored in order to ensure correct cluster state (e.g. we don't want to allow a situation where user wants to restore only view, without base table, as it doesn't make sense). 
- Restoration of views is performed by dropping them (we still require them to exists in the schema at the beginning of restore), performing restore with repair, and then recreating them at the end.
- The information about dropped views schema definition and restored tables original `tombstone_gc` is now present in the `sctool progress` output.

Strange behavior has been observed in terms of creating views on restored table.
This is an example scenario:
- make a backup of a cluster consisting of a single table
- on a different cluster, recreate keyspace (possibly with different replication strategy) and table schema via CQL
- restore (restore tables with SM) contents of backed-up table into second cluster via load and stream (with primary replica only)
- repair restored table
- create materialized view on this table
- wait  for SUCCESS on all hosts in `system_distributed.view_build_status` 
- validate that view contains all necessary rows - FAILED - it contains only a fraction

What's interesting is that this scenario doesn't reproduce if we recreate schema on second cluster using backed-up schema SSTables (restore schema with SM) and not via correct CQL definitions.

Moreover, if we repair view or base table, it will contain all necessary rows.
Tests `TestRestoreFullViewSSTableSchemaIntegration` and `TestRestoreTablesViewCQLSchemaIntegration` seems to confirm this theory.

Update: this failing example scenario does not replicate on Scylla 5.2.4 (we are currently using Scylla 5.0.3).

Update: the root cause was that in tests we restore into keyspace with different replication.

Docs issue #3488 

Fixes #3370


Michal-Leszczynski:
Here are example outputs of updated `sctool` commands:

Michal-Leszczynski:
```
[miles@fedora scylla-manager]$ ./sctool restore -c myc -L s3:restoretest-demo --restore-tables -K mks.mt -T sm_20230712162511UTC --dry-run
NOTICE: dry run mode, restore is not scheduled

Restored tables:
  - mks: 1.313M (1 table)

Restored views:
  - mks.testmv
  - mks.testmv2

Disk size: ~1.313M

Locations:
  - s3:restoretest-demo

Snapshot Tag:   sm_20230712162511UTC
Batch Size:     2
Parallel:       0
```

Michal-Leszczynski:
```
[miles@fedora scylla-manager]$ ./sctool progress restore/f657e88b-2fca-4bd2-85b3-fb18ce6cb9dc -c myc 
Restore progress
Run:            8e895dce-215c-11ee-be1b-0892040e83bb
Status:         DONE
Start time:     13 Jul 23 11:06:30 CEST
End time:       13 Jul 23 11:07:48 CEST
Duration:       1m17s
Progress:       100% | 100%
Snapshot Tag:   sm_20230712162511UTC

╭──────────┬─────────────┬────────┬─────────┬────────────┬────────╮
│ Keyspace │    Progress │   Size │ Success │ Downloaded │ Failed │
├──────────┼─────────────┼────────┼─────────┼────────────┼────────┤
│ mks      │ 100% | 100% │ 1.313M │  1.313M │     1.313M │      0 │
╰──────────┴─────────────┴────────┴─────────┴────────────┴────────╯

Restored views
View: mks.testmv, Status: SUCCESS
View: mks.testmv2, Status: SUCCESS

Post-restore repair progress
Run:            8e895dce-215c-11ee-be1b-0892040e83bb
Status:         DONE
Start time:     13 Jul 23 11:06:30 CEST
End time:       13 Jul 23 11:07:48 CEST
Duration:       1m17s
Progress:       100%
Datacenters:    
  - dc1
  - dc2

╭──────────┬───────┬──────────┬──────────╮
│ Keyspace │ Table │ Progress │ Duration │
├──────────┼───────┼──────────┼──────────┤
│ mks      │    mt │ 100%     │ 0s       │
╰──────────┴───────┴──────────┴──────────╯
```

Michal-Leszczynski:
```
[miles@fedora scylla-manager]$ ./sctool progress restore/f657e88b-2fca-4bd2-85b3-fb18ce6cb9dc -c myc --details
Restore progress
Run:            8e895dce-215c-11ee-be1b-0892040e83bb
Status:         RUNNING (recreating restored views)
Start time:     13 Jul 23 11:06:30 CEST
Duration:       1m1s
Progress:       100% | 100%
Snapshot Tag:   sm_20230712162511UTC

╭──────────┬─────────────┬────────┬─────────┬────────────┬────────╮
│ Keyspace │    Progress │   Size │ Success │ Downloaded │ Failed │
├──────────┼─────────────┼────────┼─────────┼────────────┼────────┤
│ mks      │ 100% | 100% │ 1.313M │  1.313M │     1.313M │      0 │
╰──────────┴─────────────┴────────┴─────────┴────────────┴────────╯

Keyspace: mks
╭───────┬─────────────┬────────┬─────────┬────────────┬────────┬─────────────────────────┬─────────────────────────┬───────────────────╮
│ Table │ Progress    │ Size   │ Success │ Downloaded │ Failed │ Started at              │ Completed at            │ tombstone_gc mode │
├───────┼─────────────┼────────┼─────────┼────────────┼────────┼─────────────────────────┼─────────────────────────┼───────────────────┤
│ mt    │ 100% | 100% │ 1.313M │  1.313M │     1.313M │      0 │ 13 Jul 23 11:06:49 CEST │ 13 Jul 23 11:07:19 CEST │ timeout           │
╰───────┴─────────────┴────────┴─────────┴────────────┴────────┴─────────────────────────┴─────────────────────────┴───────────────────╯

Restored views
View: mks.testmv, Status: SUCCESS, Schema definition:
  CREATE MATERIALIZED VIEW IF NOT EXISTS mks.testmv AS
    SELECT *
    FROM mks.mt
    WHERE b IS NOT null
    PRIMARY KEY (b, a)
    WITH CLUSTERING ORDER BY (a ASC)
    AND bloom_filter_fp_chance = 0.01
    AND caching = {'keys':'ALL','rows_per_partition':'ALL'}
    AND comment = ''
    AND compaction = {'class':'SizeTieredCompactionStrategy'}
    AND compression = {'sstable_compression':'org.apache.cassandra.io.compress.LZ4Compressor'}
    AND crc_check_chance = 1
    AND dclocal_read_repair_chance = 0
    AND default_time_to_live = 0
    AND gc_grace_seconds = 864000
    AND max_index_interval = 2048
    AND memtable_flush_period_in_ms = 0
    AND min_index_interval = 128
    AND read_repair_chance = 0
    AND speculative_retry = '99.0PERCENTILE'
View: mks.testmv2, Status: UNKNOWN, Schema definition:
  CREATE MATERIALIZED VIEW IF NOT EXISTS mks.testmv2 AS
    SELECT *
    FROM mks.mt
    WHERE b IS NOT null
    PRIMARY KEY (b, a)
    WITH CLUSTERING ORDER BY (a ASC)
    AND bloom_filter_fp_chance = 0.01
    AND caching = {'keys':'ALL','rows_per_partition':'ALL'}
    AND comment = ''
    AND compaction = {'class':'SizeTieredCompactionStrategy'}
    AND compression = {'sstable_compression':'org.apache.cassandra.io.compress.LZ4Compressor'}
    AND crc_check_chance = 1
    AND dclocal_read_repair_chance = 0
    AND default_time_to_live = 0
    AND gc_grace_seconds = 864000
    AND max_index_interval = 2048
    AND memtable_flush_period_in_ms = 0
    AND min_index_interval = 128
    AND read_repair_chance = 0
    AND speculative_retry = '99.0PERCENTILE'

Post-restore repair progress
Run:            8e895dce-215c-11ee-be1b-0892040e83bb
Status:         RUNNING
Start time:     13 Jul 23 11:06:30 CEST
Duration:       1m1s
Progress:       100%
Intensity:      0
Parallel:       0
Datacenters:    
  - dc1
  - dc2

╭──────────┬───────┬──────────┬──────────╮
│ Keyspace │ Table │ Progress │ Duration │
├──────────┼───────┼──────────┼──────────┤
│ mks      │    mt │ 100%     │ 0s       │
╰──────────┴───────┴──────────┴──────────╯

Host: 192.168.200.11
╭──────────┬───────┬──────────┬──────────────┬─────────┬───────┬─────────────────────────┬─────────────────────────┬──────────╮
│ Keyspace │ Table │ Progress │ Token Ranges │ Success │ Error │              Started at │            Completed at │ Duration │
├──────────┼───────┼──────────┼──────────────┼─────────┼───────┼─────────────────────────┼─────────────────────────┼──────────┤
│ mks      │ mt    │     100% │            1 │       1 │     0 │ 13 Jul 23 11:07:20 CEST │ 13 Jul 23 11:07:20 CEST │       0s │
╰──────────┴───────┴──────────┴──────────────┴─────────┴───────┴─────────────────────────┴─────────────────────────┴──────────╯

Host: 192.168.200.12
╭──────────┬───────┬──────────┬──────────────┬─────────┬───────┬─────────────────────────┬─────────────────────────┬──────────╮
│ Keyspace │ Table │ Progress │ Token Ranges │ Success │ Error │              Started at │            Completed at │ Duration │
├──────────┼───────┼──────────┼──────────────┼─────────┼───────┼─────────────────────────┼─────────────────────────┼──────────┤
│ mks      │ mt    │     100% │            1 │       1 │     0 │ 13 Jul 23 11:07:20 CEST │ 13 Jul 23 11:07:20 CEST │       0s │
╰──────────┴───────┴──────────┴──────────────┴─────────┴───────┴─────────────────────────┴─────────────────────────┴──────────╯

Host: 192.168.200.13
╭──────────┬───────┬──────────┬──────────────┬─────────┬───────┬─────────────────────────┬─────────────────────────┬──────────╮
│ Keyspace │ Table │ Progress │ Token Ranges │ Success │ Error │              Started at │            Completed at │ Duration │
├──────────┼───────┼──────────┼──────────────┼─────────┼───────┼─────────────────────────┼─────────────────────────┼──────────┤
│ mks      │ mt    │     100% │            1 │       1 │     0 │ 13 Jul 23 11:07:20 CEST │ 13 Jul 23 11:07:20 CEST │       0s │
╰──────────┴───────┴──────────┴──────────────┴─────────┴───────┴─────────────────────────┴─────────────────────────┴──────────╯

```

pkg/managerclient/model.go
```diff
@@ -1062,12 +1067,14 @@ func (rp RestoreProgress) Render(w io.Writer) error {
 		}
 	}
 
+	rp.addViewProgress(w)
```
karol-kokoszka, line 1070: Please render view progress after the post-repair progress, so that it follows the overall restore progress.

pkg/managerclient/restore.go
```diff
@@ -4,21 +4,25 @@ package managerclient
...
-	RestoreStageEnableTG   = "ENABLE_TGC"
-	RestoreStageDone       = "DONE"
+	RestoreStageInit          = "INIT"
+	RestoreStageDropViews     = "DROP_VIEWS"
+	RestoreStageDisableTGC    = "DISABLE_TGC"
+	RestoreStageData          = "DATA"
+	RestoreStageRepair        = "REPAIR"
+	RestoreStageEnableTG      = "ENABLE_TGC"
+	RestoreStageRecreateViews = "RECREATE_VIEWS"
+	RestoreStageDone          = "DONE"
```
karol-kokoszka, line 14: Enums in go are handled differently.
https://go.dev/play/p/6a3t9StJIup

pkg/scyllaclient/model.go
```diff
@@ -239,3 +239,30 @@ func makeScyllaFeatures(endpointStates []*models.EndpointState) map[string]Scyll
...
+
+// Index returns status position in ViewBuildStatusOrder.
+func (s ViewBuildStatus) Index() int {
+	for i, stage := range ViewBuildStatusOrder {
+		if s == stage {
+			return i
+		}
+	}
+	panic("Unknown status: " + s)
+}
```
karol-kokoszka, line 268: It's equivalent to https://go.dev/play/p/wTMQowni_rX
And the Index() method is not needed then.

pkg/service/backup/backupspec/stage.go
```diff
@@ -65,21 +65,25 @@ type RestoreStage string
...
 var RestoreStageOrder = []RestoreStage{
 	StageRestoreInit,
+	StageRestoreDropViews,
 	StageRestoreDisableTGC,
 	StageRestoreData,
 	StageRestoreRepair,
 	StageRestoreEnableTGC,
+	StageRestoreRecreateViews,
 	StageRestoreDone,
 }
```
karol-kokoszka, line 89: Same as here https://github.com/scylladb/scylla-manager/pull/3466/files#r1265528159

pkg/service/backup/service_backup_restore_integration_test.go
```diff
@@ -1077,11 +1271,23 @@ func grantPermissionsToUser(s gocqlx.Session, target RestoreTarget, user string)
 	var ks, t string
 	iter := s.Query("SELECT keyspace_name, table_name FROM system_schema.tables", nil).Iter()
 	for iter.Scan(&ks, &t) {
+		// Regular tables require ALTER permission.
 		if f.Check(ks, t) {
 			if err = s.ExecStmt(fmt.Sprintf("GRANT ALTER ON %s.%s TO %s", ks, t, user)); err != nil {
 				return errors.Wrap(err, "grant alter permission")
 			}
 		}
+		// Views of restored base tables require DROP and CREATE permissions.
```
karol-kokoszka, line 1280: Needs to be documented https://manager.docs.scylladb.com/master/restore/restore-tables.html#prerequisites

karol-kokoszka (commented):
Nice!
What about the comment from the PR description "This PR will require a follow-up PR in terms of documenting those facts in docs." ?
Please include that into this PR or remove "Fixes #..." from the description so that issue is not closed when only this PR is merged.

pkg/scyllaclient/client_scylla.go
```diff
@@ -942,6 +942,31 @@ func (c *Client) FlushTable(ctx context.Context, keyspace, table string) error {
...
+	if len(resp.Payload) == 0 {
+		return StatusUnknown, nil
+	}
+
+	minStatus := StatusSuccess
+	for _, v := range resp.Payload {
+		status := ViewBuildStatus(v.Value)
+		if status.Index() < minStatus.Index() {
+			minStatus = status
+		}
```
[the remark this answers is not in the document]
Michal-Leszczynski: Yes, [nodetool viewbuildstatus](https://opensource.docs.scylladb.com/stable/operating-scylla/nodetool-commands/viewbuildstatus.html) returns status for each node in the cluster (view might have been successfully built on one node, while the other is still processing it's part of data). For simplicity, I decided to return only 1 status, which is the "earliest" status from any node (e.g when view is being built on one node and is already built on another node, this function returns "STARTED" and not the "SUCCESS" status). The reason for that is:
- it simplifies waiting on view, as we are just awaiting single "SUCCESS" status
- `progress` display does not seem like it needs per host granularity display (user can still use nodetool to get more info if needed)

pkg/managerclient/restore.go
Michal-Leszczynski: It is an logical enumeration, as the order of those strings define the order of performed restore stages.
I'm not gonna lie, I added this comment just so that linter does not complain, so I don't mind changing it to something that you think would be more fitting.

pkg/scyllaclient/model.go
Michal-Leszczynski: Not really, as this implementation lacks a function to convert string into the enum type, which is needed when parsing Scylla response. So your suggestion requires 2 conversions (string -> int, int -> string) and the current one requires only 1 (string -> int).

I made this implementation mainly because that's how we handle string enums in SM (e.g. backup stage enum), so do you think that we should change the way in which we handle it in other places as well?

pkg/service/backup/service_backup_restore_integration_test.go
Michal-Leszczynski: That's true!
Generally there are more thing to change in restore docs because of this PR, but I was going to wait with that until we agree that this implementation is what we are wanting to go through with.

Michal-Leszczynski:
I think that it's easier for the docs team to have separate PR for docs changes, so that the implementation commits and conversation does not get in their way (also it allows for more dedicated labeling).

What about including a separate issue for changing the docs (more clutter)? Or do you prefer to keep one issue for both and close it only when both PR are merged (more attention needed in terms of linking PR to issue).

pkg/scyllaclient/model.go
karol-kokoszka: > Not really, as this implementation lacks a function to convert string into the enum type,

The function below does the conversion.
By adding String() string function to the struct, you are defining the conversion to string.

If someone will do fmt.Println("%s", vbs)
Then the String() method is executed.

func (vbs ViewBuildStatus) String() string {
	switch vbs {
	case StatusUnknown:
		return "UNKNOWN"
	case StatusStarted:
		return "STARTED"
	case StatusSuccess:
		return "SUCCESS"
	}
	return ""
}

karol-kokoszka:
@Michal-Leszczynski I don't mind if the docs are the separate issue. If you want to have another issue, then please create it and link it in this PR.

karol-kokoszka (approved):
:+1:

pkg/scyllaclient/model.go
Michal-Leszczynski: >>Not really, as this implementation lacks a function to convert string into the enum type,

> The function below does the conversion.

But what about the conversion string -> int?
Scylla returns us a few raw strings, we want to compare them in order to choose the "earliest" one in terms of enumeration. How would you implement `ViewBuildStatus` with your implementation?

karol-kokoszka: > But what about the conversion string -> int?
Scylla returns us a few raw strings, we want to compare them in order to choose the "earliest" one in terms of enumeration. How would you implement ViewBuildStatus with your implementation?

Take a look at the example once again: https://go.dev/play/p/DuNMrNa06tk
ViewBuildStatus is of int type
`type ViewBuildStatus int`
So you can compare it with int explicitly.

By adding String() string method to any struct, you are instructing compiler how to convert the value to string.
fmt.Println(var) expects string. If ViewBuildStatus would not have String() method then fmt.Println(var) does something like fmt.Println(fmt.Sprintf("%d", var)), and the int value is printed.

Michal-Leszczynski: The String() method is int -> string conversion.
But the input to this problem are string and not ints.
How would you fill the following code fragment with your implementation:
```go

type ViewBuildStatus int

// ViewBuildStatus enumeration (keeps the order)
const (
	StatusUnknown ViewBuildStatus = iota
	StatusStarted
	StatusSuccess
)

func (vbs ViewBuildStatus) String() string {
	switch vbs {
	case StatusUnknown:
		return "UNKNOWN"
	case StatusStarted:
		return "STARTED"
	case StatusSuccess:
		return "SUCCESS"
	}
	return ""
}

func main() {
  s1 := "STARTED"
  s2 := "SUCCESS"
  
  // compare statuses and print the "earliest" one
}
```

karol-kokoszka: Like that https://go.dev/play/p/rU3JPkiUpHU
There is a constructor added.

But I actually see now that it's just a syntax sugar...
I'm fine with your proposal, but I don't like the fact that Index() method iterates over the slice.

Michal-Leszczynski:
Changed progress display:
```
[miles@fedora scylla-manager]$ ./sctool progress -c myc restore/24433ddc-c499-41e0-a1ff-04e2c980adb1 --details
Restore progress
Run:            8e75340f-2568-11ee-8e1b-0892040e83bb
Status:         RUNNING (recreating restored views)
Start time:     18 Jul 23 14:42:29 CEST
Duration:       1m5s
Progress:       100% | 100%
Snapshot Tag:   sm_20230712162511UTC

╭──────────┬─────────────┬────────┬─────────┬────────────┬────────╮
│ Keyspace │    Progress │   Size │ Success │ Downloaded │ Failed │
├──────────┼─────────────┼────────┼─────────┼────────────┼────────┤
│ mks      │ 100% | 100% │ 1.313M │  1.313M │     1.313M │      0 │
╰──────────┴─────────────┴────────┴─────────┴────────────┴────────╯

Keyspace: mks
╭───────┬─────────────┬────────┬─────────┬────────────┬────────┬─────────────────────────┬─────────────────────────┬───────────────────╮
│ Table │ Progress    │ Size   │ Success │ Downloaded │ Failed │ Started at              │ Completed at            │ tombstone_gc mode │
├───────┼─────────────┼────────┼─────────┼────────────┼────────┼─────────────────────────┼─────────────────────────┼───────────────────┤
│ mt    │ 100% | 100% │ 1.313M │  1.313M │     1.313M │      0 │ 18 Jul 23 14:42:53 CEST │ 18 Jul 23 14:43:23 CEST │ timeout           │
╰───────┴─────────────┴────────┴─────────┴────────────┴────────┴─────────────────────────┴─────────────────────────┴───────────────────╯

Post-restore repair progress
Run:            8e75340f-2568-11ee-8e1b-0892040e83bb
Status:         RUNNING
Start time:     18 Jul 23 14:42:29 CEST
Duration:       1m5s
Progress:       100%
Intensity:      0
Parallel:       0
Datacenters:    
  - dc1
  - dc2

╭──────────┬───────┬──────────┬──────────╮
│ Keyspace │ Table │ Progress │ Duration │
├──────────┼───────┼──────────┼──────────┤
│ mks      │    mt │ 100%     │ 0s       │
╰──────────┴───────┴──────────┴──────────╯

Host: 192.168.200.11
╭──────────┬───────┬──────────┬──────────────┬─────────┬───────┬─────────────────────────┬─────────────────────────┬──────────╮
│ Keyspace │ Table │ Progress │ Token Ranges │ Success │ Error │              Started at │            Completed at │ Duration │
├──────────┼───────┼──────────┼──────────────┼─────────┼───────┼─────────────────────────┼─────────────────────────┼──────────┤
│ mks      │ mt    │     100% │            1 │       1 │     0 │ 18 Jul 23 14:43:24 CEST │ 18 Jul 23 14:43:24 CEST │       0s │
╰──────────┴───────┴──────────┴──────────────┴─────────┴───────┴─────────────────────────┴─────────────────────────┴──────────╯

Host: 192.168.200.12
╭──────────┬───────┬──────────┬──────────────┬─────────┬───────┬─────────────────────────┬─────────────────────────┬──────────╮
│ Keyspace │ Table │ Progress │ Token Ranges │ Success │ Error │              Started at │            Completed at │ Duration │
├──────────┼───────┼──────────┼──────────────┼─────────┼───────┼─────────────────────────┼─────────────────────────┼──────────┤
│ mks      │ mt    │     100% │            1 │       1 │     0 │ 18 Jul 23 14:43:24 CEST │ 18 Jul 23 14:43:24 CEST │       0s │
╰──────────┴───────┴──────────┴──────────────┴─────────┴───────┴─────────────────────────┴─────────────────────────┴──────────╯

Host: 192.168.200.13
╭──────────┬───────┬──────────┬──────────────┬─────────┬───────┬─────────────────────────┬─────────────────────────┬──────────╮
│ Keyspace │ Table │ Progress │ Token Ranges │ Success │ Error │              Started at │            Completed at │ Duration │
├──────────┼───────┼──────────┼──────────────┼─────────┼───────┼─────────────────────────┼─────────────────────────┼──────────┤
│ mks      │ mt    │     100% │            1 │       1 │     0 │ 18 Jul 23 14:43:24 CEST │ 18 Jul 23 14:43:24 CEST │       0s │
╰──────────┴───────┴──────────┴──────────────┴─────────┴───────┴─────────────────────────┴─────────────────────────┴──────────╯

Restored views
View: mks.testmv, Status: SUCCESS, Schema definition:
  CREATE MATERIALIZED VIEW IF NOT EXISTS mks.testmv AS
    SELECT *
    FROM mks.mt
    WHERE b IS NOT null
    PRIMARY KEY (b, a)
    WITH CLUSTERING ORDER BY (a ASC)
    AND bloom_filter_fp_chance = 0.01
    AND caching = {'keys':'ALL','rows_per_partition':'ALL'}
    AND comment = ''
    AND compaction = {'class':'SizeTieredCompactionStrategy'}
    AND compression = {'sstable_compression':'org.apache.cassandra.io.compress.LZ4Compressor'}
    AND crc_check_chance = 1
    AND dclocal_read_repair_chance = 0
    AND default_time_to_live = 0
    AND gc_grace_seconds = 864000
    AND max_index_interval = 2048
    AND memtable_flush_period_in_ms = 0
    AND min_index_interval = 128
    AND read_repair_chance = 0
    AND speculative_retry = '99.0PERCENTILE'
View: mks.testmv2, Status: UNKNOWN, Schema definition:
  CREATE MATERIALIZED VIEW IF NOT EXISTS mks.testmv2 AS
    SELECT *
    FROM mks.mt
    WHERE b IS NOT null
    PRIMARY KEY (b, a)
    WITH CLUSTERING ORDER BY (a ASC)
    AND bloom_filter_fp_chance = 0.01
    AND caching = {'keys':'ALL','rows_per_partition':'ALL'}
    AND comment = ''
    AND compaction = {'class':'SizeTieredCompactionStrategy'}
    AND compression = {'sstable_compression':'org.apache.cassandra.io.compress.LZ4Compressor'}
    AND crc_check_chance = 1
    AND dclocal_read_repair_chance = 0
    AND default_time_to_live = 0
    AND gc_grace_seconds = 864000
    AND max_index_interval = 2048
    AND memtable_flush_period_in_ms = 0
    AND min_index_interval = 128
    AND read_repair_chance = 0
    AND speculative_retry = '99.0PERCENTILE'

```

pkg/scyllaclient/model.go
Michal-Leszczynski: Please take a look at the commit `utils/slice: introduce Index` and tell me if that looks better to you. It still relies on a slice, but at least it's "immutable" and we have generic function for stage/status enumeration.

Michal-Leszczynski:
Rebased on master with changes to progress display and enhanced approach to `Index` function.

karol-kokoszka (approved):
:+1:

PR #3481, opened 2023-07-17
Use many Scylla versions in gh actions

### LINKED ISSUE #3471: Test many Scylla versions
Right now our gh actions run test only on the Scylla with version hardcoded in `testing/.env`. 
Perhaps we should use [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) to run tests on more versions (at least the oldest and newest supported versions).

Because of restore task, SM relies on the internals of Scylla more than before, so it would be nice to monitor, whether some Scylla assumptions used while writing SM code haven't changed in the meantime.

Michal-Leszczynski:
This PR changes gh actions workflow so that it runs against many Scylla versions.
Fixes #3471

### DIFF
--- a/.github/actions/test-setup/action.yml
+++ b/.github/actions/test-setup/action.yml
@@ -1,13 +1,18 @@
 name: "Set up testing dependencies"
 description: "Set up GO, package dependencies and install-dependencies"
+
 inputs:
+  scylla-version:
+    description: "Specifies Scylla version for test environment"
+    required: false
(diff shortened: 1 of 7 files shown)


dkropachev (approved):

Michal-Leszczynski:
Possible cause of failures #3479

dkropachev:
@karol-kokoszka , should we merge it ?

Issue #3472, opened 2023-07-14
Test env setup takes a lot of time


Michal-Leszczynski:
After ipv6 changes introduced in [commit](1819fae5), `make start-dev-env` takes a significant time to set up testing environment. When I want run some tests on different Scylla versions, it is more time efficient for me to just:
- `git checkout 59a303e8` (one commit before described changes has been made)
- `make start-dev-env`
- `git checkout <working-branch>`

Why is this set-up taking such a long time in comparison to the previous one? If ipv6 env cannot be set-up faster, then could we heave an option in makefile to set up a simple ipv4 env faster, like it was done before?

@dkropachev @karol-kokoszka

karol-kokoszka:
It has nothing common with ipv6 actually. Makefile just chooses explicitly which nodes to start and is doing it in chain. 
https://github.com/scylladb/scylla-manager/blame/fd8417dafa0f9a6194c84dca556c7bdcf37b3c64/testing/Makefile#L72-L86

I think we can rollback to previous apporach.

dkropachev:
There is official prodecidure on cluster bootstraping https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/create-cluster-multidc.html
That says:
`Starting Scylla nodes, starting with the seed node, wait until it is in UN state, and repeat for the other nodes.`

The reason why it is so, is that scylla does not handle some edge cases that pops up in parallel bootstraping well.
As result you node boostraping endup in failure and cannot be proceeded.

In my personal experience I had 30% chance that cluster bootstrap fails before https://github.com/scylladb/scylla-manager/pull/3436, now i have not seen it at all.

Given the fact that this is also part of pipeline I would not recomend falling back to old bootstraping proceedure.

PR #3483, opened 2023-07-17
fix(testing): make scylla load faster

dkropachev:
After https://github.com/scylladb/scylla-manager/pull/3436 cluster nodes are bootstrap one by one to ensure stability As result test env boostraping too long.
Fortunately there is a way to make skylla load faster via `skip_wait_for_gossip_to_settle` which makes it not to wait for gossip to settle, which readuces load time 5 fold

Closes https://github.com/scylladb/scylla-manager/issues/3472


Michal-Leszczynski (approved):
After looking at the gh actions short exec time I thought, that we accidentally don't run all the tests.
But it looks like we do! That's an amazing life quality improvement!

mykaul:
There's a reason this (skip_wait_for_gossip_to_settle ) is not the default, though... Beware ;-)

Issue #3485, opened 2023-07-18
Make scylla restart stable


dkropachev:
After making scylla [skip gossip settling](https://github.com/scylladb/scylla-manager/pull/3483) restart become unstable.
Test can think that node is ready to operate while it is not.
It results with test failures like this:
```
--- When: query contents of restored table
    service_backup_restore_integration_test.go:1008: Cannot achieve consistency level for cl ONE. Requires 1, alive 0
```

dkropachev:
https://github.com/scylladb/scylla-manager/pull/3483 was reverted

PR #3486, opened 2023-07-18
fix(restore-tests): make restartScylla to check on other nodes

dkropachev:
After making scylla [skip gossip settling](https://github.com/scylladb/scylla-manager/pull/3483) restart become unstable.
Test can think that node is ready to operate while it is not.

Closes https://github.com/scylladb/scylla-manager/issues/3485


Michal-Leszczynski:
Possible cause of failures #3479

Issue #3425, opened 2023-06-19
`--parallel` flag to ensure single repair job per host


karol-kokoszka:
Current definition of `--parallel` flag is as follow https://manager.docs.scylladb.com/stable/sctool/repair.html#parallel

> The maximum number of Scylla repair jobs that can run at the same time (on different token ranges and replicas). Each node can take part in at most one repair at any given moment. By default the maximum possible parallelism is used. The effective parallelism depends on a keyspace replication factor (RF) and the number of nodes. The formula to calculate is is as follows: number of nodes / RF, ex. for 6 node cluster with RF=3 the maximum parallelism is 2.

We have to ensure that only a single repair job is executed on a host. Definition of the `--parallel` flag stays as-is.

~~We want to change it to~~
~~> The maximum number of Scylla repair jobs that can run at the same time on a single host.~~

~~Given a scenario where:~~
~~- there are 4 nodes A, B, C, D~~
~~- replication factor is set to 2~~
~~- woth ranges r1, r2, r3, r4, r5, r6~~
~~- ranges are distributed as follow~~
~~r1 are on nodes A, B~~
~~r2 -> A, C~~
~~r3 -> A. D~~
~~r4 -> B, C~~
~~r5 -> B, D~~
~~r6 -> C, D~~
~~.... with parallel set to 1 only following combinations are possible...~~
~~r1(A, B) together with r6(C, D)~~
~~r2(A, C) together with r5(B, D) etc~~

~~We must assure that max `--parallel` number of jobs are executed in parallel on a single host~~

**Default for `parallel`: 1**

mykaul:
I would not change the meaning of an existing flag. I think it's better to introduce a new flag and deprecate this one.

vladzcloudius:
@asias FYI.

I agree with @mykaul - the current definition of `--parallel` and `--intensity` are totally ok.

However it seems the implementation is currently broken.

Two questions:
1) Since what SM version.
2) Why we are not fixing it ASAP?

vladzcloudius:
ref https://github.com/scylladb/scylla-manager/issues/3463

karol-kokoszka:
Current definition of `--parallel` doesn't match the SM behavior.

It says:
> Each node can take part in at most one repair at any given moment.

#3463 shows that it's not true.

In new implementation of the repair we wan't to guarantee that max configured value number of jobs can be send to a single host.
[...]

PR #3496, opened 2023-07-20
repair: configurable max number of jobs per host

### LINKED ISSUE #3480: Ensure only single repair job per host is executed during the repair (backport to < 3.2)
We must ensure that this what is written in scylla-manager docs
https://manager.docs.scylladb.com/stable/sctool/repair.html#parallel
is met by the implementation.

> The maximum number of Scylla repair jobs that can run at the same time (on different token ranges and replicas). **Each node can take part in at most one repair at any given moment**. By default the maximum possible parallelism is used. The effective parallelism depends on a keyspace replication factor (RF) and the number of nodes. The formula to calculate is is as follows: number of nodes / RF, ex. for 6 node cluster with RF=3 the maximum parallelism is 2

**Each node can take part in at most one repair at any given moment** is not met for now -> see https://github.com/scylladb/scylla-manager/issues/3463#issuecomment-1631789229

3.2 release is going to update the overall repair in scylla-manager, but we must support previous versions as well.
As the repair in 3.2 is kind of bigger refactor, this issue is supposed to guarantee the max number of repair jobs execution on single host for SM < 3.2 and the fix must be backported to all supported SM versions.

+ Add integration test veryfing the number of jobs being executed at the moment.

karol-kokoszka:
Fixes #3480 

There is additional flag provided for the `sctool repair` command that is responsible for ensuring that there is no more than `single_host_parallelism` number of repair jobs executed against the host.

---

Please make sure that:


Michal-Leszczynski:
What about including `single_host_parallelism` in `repair --dry-run` output?

pkg/service/repair/service_repair_integration_test.go
```diff
@@ -483,6 +529,109 @@ func TestServiceGetTargetIntegration(t *testing.T) {
...
+	WriteData(t, clusterSession, ks1, 5, t1, t2)
+	WriteData(t, clusterSession, ks2, 5, t1, t2)
+	WriteData(t, clusterSession, ks3, 5, t1, t2)
+
+	t.Run("repair schedules only one job per host at any given time", func(t *testing.T) {
+		ctx := context.Background()
+
+		target := h.generateTarget(map[string]any{
+			"fail_fast":               true,
+			"single_host_parallelism": 1,
```
Michal-Leszczynski, line 557: It would be nicer to use `maxJobsOnHost` here instead of straight 1 (possible var rename).

pkg/service/repair/service_repair_integration_test.go
```diff
@@ -483,6 +529,109 @@ func TestServiceGetTargetIntegration(t *testing.T) {
...
+					muHIJ.Unlock()
+
+					muJPH.Lock()
+					defer muJPH.Unlock()
+
+					for _, host := range hosts {
+						jobsPerHost[host]--
+					}
+				}
+			}
```
Michal-Leszczynski, line 621: Wasn't this test flaky at the beginning? What has changed?

Michal-Leszczynski:
In terms of general repair refactor, we will preserve `--single-host-parallelism` and won't change the meaning of `--parallel` as issue #3425 suggests?

pkg/service/repair/service_repair_integration_test.go
karol-kokoszka: makes sense, fixed

pkg/service/repair/service_repair_integration_test.go
karol-kokoszka: it was, and actually I found the reason after couple of debugging hours...
repair_job_id is given per master host, not per cluster

karol-kokoszka:
> In terms of general repair refactor, we will preserve --single-host-parallelism and won't change the meaning of --parallel as issue https://github.com/scylladb/scylla-manager/issues/3425 suggests?

@Michal-Leszczynski That's the idea. It simplifies a lot.

karol-kokoszka:
> What about including single_host_parallelism in repair --dry-run output?

But the --dry-run doesn't include information about task configuration. Does it ?

Michal-Leszczynski:
> But the --dry-run doesn't include information about task configuration. Does it ?

It prints things like ignored hosts, dcs, schedule, so it might be nice to print the `parallel` configuration there (`--parallel`, `--intensity`, `--single-host-parallelism`)? 

Similarly, backup's `--dry-run` prints things like `retention` etc.

But as this isn't anything required for this flag's functionality, we can add it (if we decide so) later with the general repair refactor.

pkg/service/repair/service.go
```diff
@@ -888,6 +917,18 @@ func (i *intensityHandler) SetParallel(ctx context.Context, parallel int) error
 	return nil
 }
 
+// SetSingleHostParallelism sets repair max parallelism per host value.
+func (i *intensityHandler) SetSingleHostParallelism(ctx context.Context, singleHostParallelism int) error {
+	if singleHostParallelism < defaultSingleHostParallelism {
```
Michal-Leszczynski, line 922: Shouldn't it be lesser than 0, not the default value?

karol-kokoszka:
@Michal-Leszczynski I'm merging it, need to see the dtest output. Please place comments if you find something suspicious in the PR.

pkg/service/repair/service.go
karol-kokoszka: defaultSingleHostParallelism is const = 0

swagger/gen/scylla-manager/client/scylla_manager_client.go
```diff
@@ -9,7 +9,6 @@ import (
 	"github.com/go-openapi/runtime"
 	httptransport "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
-	"github.com/scylladb/scylla-manager/v3/swagger/gen/scylla-manager/client/operations"
```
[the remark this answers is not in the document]
Michal-Leszczynski: How is it used?
GH actions tests passed and checking locally, it didn't have an impact on `make check` or SM start-up.

pkg/service/repair/service.go
vladzcloudius: @Michal-Leszczynski was right - the default should have been 1.
@karol-kokoszka I wonder what's the point of a review if you ignore completely relevant reviewers comments like here?

karol-kokoszka: @vladzcloudius it cannot be default to 1. The main reason for that is "We don't want to change the current behavior of the repair in the manager". By default, there is no limit on this parallelism. This change cannot be invasive for the current users.
We checked with the SCT tests, that setting this parallelism to 1 lowers the performance of the overall repair.

vladzcloudius: But there IS limit for this parameter and it's documented and it's exactly `1`.
@karol-kokoszka I don't think it makes sense to discuss the PR any longer. 

We are going to take it with @tzach as soon as he returns.
Until then I suggest to not add any additional noise in the public repo.

Please, wait for an update from Tzach about this parameter in general and changes in repair APIs in particular.

Issue #3424, opened 2023-06-19
Deprecate the <0,1> intensity flag support


karol-kokoszka:
https://manager.docs.scylladb.com/stable/sctool/repair.html#intensity

`--intensity`
   - 0: Set number of token ranges per repair job automatically.
   - 1, 2, 3, 4, .... : Specify number of ranges per repair job manually.
   - **(0,1): The decimal (0,1) value support in --intensity should be deprecated. Allow the user to specify, but ignore it, setting to 1 internally.**

The --intensity option controls how many token ranges a repair job repairs.

For example, a node owns 512 token ranges, max_repair_ranges_in_parallel is 6.

Setting --intensity to 1, 1 token ranges will be sent to Scylla core per repair job.

Setting --intensity to 2, 2 token ranges will be sent to Scylla core per repair job.

Setting --intensity to 6, 6 token ranges will be sent to Scylla core per repair job.

TODO:
- Update the repair service and remove all the logic connected with (0,1) float intensity level
- Update the documentation https://manager.docs.scylladb.com/stable/sctool/repair.html#intensity

vladzcloudius:
@karol-kokoszka could you, please, clarify what's the point of this change?

Michal-Leszczynski:
We asked @asias about repair and what would be the best way for SM to orchestrate it.
It turned out that shard count shouldn't be taken into consideration when calculating the amount of token ranges repaired by Scylla in a single job (although it should be used for choosing the repair master).

Intensity lesser than 1 is defined as:
`It can be a decimal between (0,1), in that case it specifies percent of shards that can be repaired in parallel on a repair master node.`
So that's why we wanted to deprecate it.

On the other hand, it was also said that:
> From Asias: There is no limitation of sending how many token ranges per repair job. Sending less than max_repair_ranges_in_parallel per repair job allows users to run repair less aggressively, which is the main reason we want to introduce the option in the first place. We could just send max_repair_ranges_in_parallel per repair job, but this means we will send a lot of repair jobs if there are too many token ranges per node. I think it makes more sense to send some percentage of token ranges per repair job, say 5% of the total token ranges we need to repair per repair job. This works for all cases no matter how many token ranges a node owns. Internally, scylla will cap the number of ranges to repair in parallel anyway. Sending more token ranges per repair job than  max_repair_ranges_in_parallel will not have performance impact. It only has an impact on the fail and resume granularity. 

[...]

Issue #3423, opened 2023-06-19
Update the repair procedure


karol-kokoszka:
This is the epic that groups all tasks related to `repair` procedure fixes and updates.
https://manager.docs.scylladb.com/stable/repair/

The requirements for `repair` procedure are as follow:
- Repair keyspace by keyspace (in the context of manager repair task) ordered by their total size
- Repair table by table (in the context of repairing the keyspace)
- Repair MV and si after the base table (only when CQL credentials are available)
- Execute max `--parallel` repair job per host
- Start repair job on the replica node with smallest shard count
- Repair system tables before user tables

Full design doc is here https://docs.google.com/document/d/1mBaufbomXU6hDO_25KkhC7CO65AbFgKpHVpB5Mei8xA/edit
(visible for people from ScyllaDB org only).

The idea is to update/align current implementation of repair to the given requirements.

```[tasklist]
### Tasks
- [ ] https://github.com/scylladb/scylla-manager/issues/3424
- [ ] https://github.com/scylladb/scylla-manager/issues/3425
- [ ] https://github.com/scylladb/scylla-manager/issues/3362
- [ ] https://github.com/scylladb/scylla-manager/issues/3367
- [ ] https://github.com/scylladb/scylla-manager/issues/3426
- [ ] https://github.com/scylladb/scylla-manager/issues/3463
- [ ] https://github.com/scylladb/scylla-manager/issues/3515
```

karol-kokoszka:
Epic should be started with this issue https://github.com/scylladb/scylla-manager/issues/3424
It will remove quite much of not needed (and complicated) code.

PR #3506, opened 2023-07-27
service.go: fix repair semantics regression

### LINKED ISSUE #3463: Too high repair parallelism
Version: 3.1.1
Cluster size: 9
RF: 3
parallel: 0
max_repair_ranges_in_parallel: 6
intensity: 1

*Description*
According to SM doc pages (https://manager.docs.scylladb.com/stable/sctool/repair.html#id8) the parallelism for a KS with RF=3 and cluster with 9 nodes is supposed to be 3.

However we see that the effective parallelism is much more than that:

```
$ run scylla_hosts curl  http://127.0.0.1:10000/storage_service/active_repair
Elapsed: 00:01 - running on Scylla nodes: 'curl http://127.0.0.1:10000/storage_service/active_repair'...-
running on Scylla nodes: 'curl http://127.0.0.1:10000/storage_service/active_repair': done. Took: 1.43 seconds.
╒══════════════════════════════════════════╤═══════════════════════════════════════╤══════════════╕
│ Scylla host                              │ Result                                │ Duration     │
╞══════════════════════════════════════════╪═══════════════════════════════════════╪══════════════╡
│ scylla-xxx-xxxx-xxxx-asia-southeast1-a-1 │ []                                    │ 0.68 seconds │
├──────────────────────────────────────────┼───────────────────────────────────────┼──────────────┤
│ scylla-xxx-xxxx-xxxx-asia-southeast1-b-1 │ [5,4,3,2,1]                           │ 0.67 seconds │
├──────────────────────────────────────────┼───────────────────────────────────────┼──────────────┤
│ scylla-xxx-xxxx-xxxx-asia-southeast1-c-1 │ [14,13,12,11,10,9,8,7,6]              │ 0.67 seconds │
├──────────────────────────────────────────┼───────────────────────────────────────┼──────────────┤
│ scylla-xxx-xxxx-xxxx-asia-southeast1-a-3 │ [5,4,3,2,1]                           │ 0.68 seconds │
├──────────────────────────────────────────┼───────────────────────────────────────┼──────────────┤
│ scylla-xxx-xxxx-xxxx-asia-southeast1-b-3 │ []                                    │ 0.68 seconds │
├──────────────────────────────────────────┼───────────────────────────────────────┼──────────────┤
│ scylla-xxx-xxxx-xxxx-asia-southeast1-c-3 │ [5,4,3,2,1]                           │ 0.69 seconds │
├──────────────────────────────────────────┼───────────────────────────────────────┼──────────────┤
│ scylla-xxx-xxxx-xxxx-asia-southeast1-a-4 │ []                                    │ 0.68 seconds │
├──────────────────────────────────────────┼───────────────────────────────────────┼──────────────┤
│ scylla-xxx-xxxx-xxxx-asia-southeast1-b-4 │ [151,150,149,148,147,146,145,144,143] │ 0.68 seconds │
├──────────────────────────────────────────┼───────────────────────────────────────┼──────────────┤
│ scylla-xxx-xxxx-xxxx-asia-southeast1-c-4 │ [37,36,35,33,34]                      │ 0.67 seconds │
╘══════════════════════════════════════════╧═══════════════════════════════════════╧══════════════╛
```

As we can see there are 4*5 + 2*9 = 38 repair jobs running at the same time. 
This doesn't seem to correlate with the documentation referenced above.

vladzcloudius:
Restore the repair task sanity.

Make the code correspond design and Documentation.
See more in the patches description.

Fixes #3463 and #3424

### DIFF
--- a/pkg/service/repair/controller.go
+++ b/pkg/service/repair/controller.go
@@ -172,28 +172,39 @@ func (c *rowLevelRepairController) TryBlock(hosts []string) (bool, allowance) {
 	return true, a
 }
 
+func (c *rowLevelRepairController) parallelLimit() int {
+	p := c.intensity.Parallel()
+
+	if p <= defaultParallel {
+		return c.intensity.MaxParallel()
+	}
+
+	return p
+}
+
 func (c *rowLevelRepairController) shouldBlock(hosts []string, intensity float64) bool {
 	// ALLOW if nothing is running
 	if !c.Busy() {
 		return true
 	}
 
-	// DENY if any host has max nr of jobs already running
+	// DENY if any host has a job running already. We don't allow more than a single repair job per host.
 	for _, h := range hosts {
-		if c.jobs[h] >= c.limits[h].Default {
+		if c.jobs[h] >= 1 {
 			return false
 		}
 	}
 
-	// DENY if there is too much parallel replicas being repaired
-	if parallel := c.intensity.Parallel(); parallel != defaultParallel {
-		hash := replicaHash(hosts)
-		if _, ok := c.busyReplicas[hash]; !ok {
-			l := len(c.busyReplicas) + 1
-			if l > parallel || l > c.intensity.MaxParallel() {
-				return false
-			}
-		}
+	// DENY (Sanity) if allJobs + 1 > MaxParallel.
+	// However, this should never happen due to "up to 1 repair task per replica" and the fact that MaxParallel is
+	// calculated the way that allows up to 1 repair task per host.
+	if c.allJobs+1 > c.intensity.MaxParallel() {
+		return false
+	}
+
+	// DENY if allJobs is greater than the configured parallelism
+	if c.allJobs+1 > c.parallelLimit() {
+		return false
 	}
 
 	// DENY if all workers are busy
@@ -210,6 +221,8 @@ func (c *rowLevelRepairController) block(hosts []string, ranges int) {
 		c.jobs[h]++
 		c.busyRanges[h] += ranges
 	}
+	//TODO: the value for every replica set should never be greater than 1
+	// We should build some test around this fact.
 	c.busyReplicas[replicaHash(hosts)]++
 }
 
--- a/pkg/service/repair/controller.go
+++ b/pkg/service/repair/controller.go
@@ -92,19 +92,14 @@ func (c *defaultController) allowance(hosts []string) allowance {
 		Replicas: hosts,
 		Ranges:   math.MaxInt32,
 	}
-	if i < 1 {
-		a.ShardsPercent = i
-	}
 
 	for _, h := range hosts {
 		var v int
 		switch {
 		case i == maxIntensity:
 			v = c.limits[h].Max
-		case i < 1:
-			v = c.limits[h].Default
 		default:
-			v = int(i) * c.limits[h].Default
+			v = int(i)
 		}
 		if v < a.Ranges {
 			a.Ranges = v
@@ -190,28 +185,6 @@ func (c *rowLevelRepairController) shouldBlock(hosts []string, intensity float64
 		}
 	}
 
-	// DENY if any host already runs at the full intensity.
-	// This may happen if we decrease intensity, for a while we need to reduce
-	// number of jobs to reduce number of busy ranges.
-	ranges := c.rangesForIntensity(hosts, intensity)
-	for _, h := range hosts {
-		var (
-			r  = c.busyRanges[h] + ranges
-			ok bool
-		)
-		switch {
-		case intensity == maxIntensity:
-			ok = r <= c.limits[h].Max
-		case intensity <= 1:
-			ok = r <= c.limits[h].Default
-		default:
-			ok = r <= int(intensity)*c.limits[h].Default && r <= c.limits[h].Max
-		}
-		if !ok {
-			return false
-		}
-	}
-
 	// DENY if there is too much parallel replicas being repaired
 	if parallel := c.intensity.Parallel(); parallel != defaultParallel {
 		hash := replicaHash(hosts)
@@ -245,9 +218,7 @@ func (c *rowLevelRepairController) allowance(hosts []string, intensity float64) 
 		Replicas: hosts,
 		Ranges:   c.rangesForIntensity(hosts, intensity),
 	}
-	if intensity < 1 {
-		a.ShardsPercent = intensity
-	}
+
 	return a
 }
 
@@ -255,27 +226,17 @@ func (c *rowLevelRepairController) rangesForIntensity(hosts []string, intensity 
 	switch {
 	case intensity == maxIntensity:
 		ranges = math.MaxInt32
-		for _, h := range hosts {
-			if v := c.limits[h].Max / c.limits[h].Default; v < ranges {
-				ranges = v
-			}
-		}
-		if ranges == 0 {
-			ranges = 1
-		}
-	case intensity < 1:
-		// If intensity < 1 return all token ranges (nr. of shards) in a single
-		// call. This is to avoid multiple workers repairing the same host in
-		// parallel. Single worker will offload work from shards using
-		// the legacy repair logic i.e. repair shard by shard.
-		ranges = math.MaxInt32
-		for _, h := range hosts {
-			if v := c.limits[h].Default; v < ranges {
-				ranges = v
-			}
-		}
 	default:
 		ranges = int(intensity)
+	}
+
+	for _, h := range hosts {
+		if v := c.limits[h].Max; v < ranges {
+			ranges = v
+		}
+	}
+	if ranges == 0 {
+		ranges = 1
 	}
 	return
 }
--- a/pkg/service/repair/service.go
+++ b/pkg/service/repair/service.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"strings"
 	"sync"
 
@@ -609,7 +610,7 @@ func (s *Service) newIntensityHandler(ctx context.Context, clusterID uuid.UUID, 
 	ih = &intensityHandler{
 		logger:      s.logger.Named("control"),
 		intensity:   atomic.NewFloat64(intensity),
-		parallel:    atomic.NewInt64(int64(parallel)),
+		parallel:    atomic.NewInt64(int64(math.Min(float64(parallel), float64(maxParallel)))),
 		maxParallel: maxParallel,
 	}
 
@@ -886,12 +887,13 @@ func (i *intensityHandler) SetParallel(ctx context.Context, parallel int) error 
 		return service.ErrValidate(errors.Errorf("setting invalid parallel value %d", parallel))
 	}
 
-	i.logger.Info(ctx, "Setting repair parallel", "value", parallel, "previous", i.parallel.Load())
-	i.parallel.Store(int64(parallel))
-
 	if parallel > i.maxParallel {
 		i.logger.Info(ctx, "Requested parallel value will be capped to maximum possible", "requested", parallel, "maximum", i.maxParallel)
+		parallel = i.maxParallel
 	}
+
+	i.logger.Info(ctx, "Setting repair parallel", "value", parallel, "previous", i.parallel.Load())
+	i.parallel.Store(int64(parallel))
 
 	return nil
 }
--- a/pkg/service/repair/service.go
+++ b/pkg/service/repair/service.go
@@ -565,9 +565,15 @@ func (s *Service) hostRangeLimits(ctx context.Context, client *scyllaclient.Clie
 			return errors.Wrapf(err, "%s: get shard count", h)
 		}
 
+		// Since we don't issue repairs for per-shard ranges anymore floating point 'intensity' becomes irrelevant.
+		// We will allow up to a single repair task per replica host as our documentation mandates and will allow
+		// 1 or more token ranges per repair task. The maximum amount ranges per task is limited by
+		// s.maxRepairRangesInParallel().
 		v := rangesLimit{
-			Default: int(shards),
-			Max:     s.maxRepairRangesInParallel(shards, totalMemory),
+			Default: 1,
+			//TODO: collect max_repair_ranges_in_parallel from scylla hosts using a not-yet-existing API
+			// In the meantime use good-enough value
+			Max: s.maxRepairRangesInParallel(shards, totalMemory),
 		}
 		s.logger.Info(ctx, "Host repair intensity limit", "host", h, "limit", v)
 
--- a/pkg/service/repair/worker.go
+++ b/pkg/service/repair/worker.go
@@ -10,7 +10,6 @@ import (
 	"github.com/scylladb/go-log"
 	"github.com/scylladb/scylla-manager/v3/pkg/dht"
 	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
-	"github.com/scylladb/scylla-manager/v3/pkg/util/parallel"
 )
 
 var errTableDeleted = errors.New("table deleted during repair")
@@ -80,13 +79,7 @@ func (w *worker) Run(ctx context.Context) error {
 }
 
 func (w *worker) handleJob(ctx context.Context, job job) error {
-	var err error
-	if w.repairType == TypeRowLevel && job.Allowance.ShardsPercent == 0 {
-		err = w.rowLevelRepair(ctx, job)
-	} else {
-		err = w.legacyRepair(ctx, job)
-	}
-	return err
+	return w.rowLevelRepair(ctx, job)
 }
 
 func (w *worker) runRepair(ctx context.Context, job job) error {
@@ -143,58 +136,6 @@ func (w *worker) rowLevelRepair(ctx context.Context, job job) error {
 		w.logger.Error(ctx, "Run row-level repair", "error", err)
 	}
 	return err
-}
-
-func (w *worker) legacyRepair(ctx context.Context, job job) error {
-	p := w.hostPartitioner[job.Host]
-
-	// In a very rare case when there is a strange partitioner
-	// do not split to shards.
-	if p == nil {
-		err := w.runRepair(ctx, job)
-		if err != nil {
-			w.logger.Error(ctx, "Run legacy repair, no sharding due to unsupported partitioner", "error", err)
-		}
-		return err
-	}
-
-	// Calculate max parallel shard repairs
-	limit := int(p.ShardCount())
-	if job.Allowance.ShardsPercent != 0 {
-		l := float64(limit) * job.Allowance.ShardsPercent
-		if l < 1 {
-			l = 1
-		}
-		limit = int(l)
-
-		w.logger.Debug(ctx, "Limiting parallel shard repairs", "total", p.ShardCount(), "limit", limit)
-	}
-
-	// Split ranges to shards
-	shardRanges, err := splitToShardsAndValidate(job.Ranges, p)
-	if err != nil {
-		return errors.Wrap(err, "split to shards")
-	}
-
-	f := func(i int) error {
-		if ctx.Err() != nil {
-			return nil
-		}
-
-		if len(shardRanges[i]) == 0 {
-			return nil
-		}
-
-		shardJob := job
-		shardJob.Ranges = shardRanges[i]
-		return w.runRepair(log.WithFields(ctx, "subranges_of_shard", i), shardJob)
-	}
-
-	notify := func(i int, err error) {
-		w.logger.Error(ctx, "Run legacy repair", "error", err)
-	}
-
-	return parallel.Run(len(shardRanges), limit, f, notify)
 }
 
 func (w *worker) waitRepairStatus(ctx context.Context, id int32, host, keyspace, table string) error {
--- a/pkg/service/repair/controller_test.go
+++ b/pkg/service/repair/controller_test.go
@@ -240,9 +240,7 @@ func TestRowLevelRepairController(t *testing.T) {
 			if a.Ranges != controllerTestDefaultRangesLimit {
 				t.Fatalf("TryBlock()=%v, expected %d ranges", a, controllerTestDefaultRangesLimit)
 			}
-			if a.ShardsPercent != 0.5 {
-				t.Fatalf("TryBlock()=%v, expected %f shards percent", a, 0.5)
-			}
+
 			queue = append(queue, a)
 		}
 		if ok, _ := ctl.TryBlock([]string{"a", "b", "c"}); ok {


Michal-Leszczynski:
It doesn't look related to #3464.

Moreover, we this patch won't really work the way you intended.
Current implementation uses repair controller in order to check if given job repair should be scheduled. Basically, repair worker asks if it can repair `intensity` amount of token ranges and controller checks if it fits under the limits (that you changed). So when `intensity > 1`, this will always fail.
On the other hand, controller has some additional logic, which always allows to schedule a repair job if there are no other repair jobs running in the cluster and that's why the tests have passed (but this means that there is only 1 repair job per cluster running, so repair will be really slow).

A quick patch for this was introduced to branch 3.1 via #3496.

vladzcloudius:
> It doesn't look related to #3464.

Right. I meant #3463 of course.
Thanks for looking at it, @Michal-Leszczynski,  and, to clarify, I didn't mean this to be merged as v1.
This is my "toy" Go attempt. So, I'm sure I'm going to need a lot of help here.
More below.

> 
> Moreover, we this patch won't really work the way you intended. Current implementation uses repair controller in order to check if given job repair should be scheduled. Basically, repair worker asks if it can repair `intensity` amount of token ranges and controller checks if it fits under the limits (that you changed). So when `intensity > 1`, this will always fail. 

I'm well aware of what you have described above and this was exactly my intention in this patch (you could see that from the inline comment).

However in following updates I'm going to bring back an integer semantics of `intensity` and restore the `parallel` semantics and requirements in full.

In particular the current ` rowLevelRepairController::shouldBlock(...)` has an outdated logic that is based on the old design where a per-shard range repair was an atomic repair item. And hence all the accounting.

Let's fix it here. More below.

> On the other hand, controller has some additional logic, which always allows to schedule a repair job if there are no other repair jobs running in the cluster and that's why the tests have passed (but this means that there is only 1 repair job per cluster running, so repair will be really slow).

Right. The logic that allocates workers.
Let's fix that.

> 
> A quick patch for this was introduced to branch 3.1 via #3496.

That patch doesn't help us much because it seems rather bogus to me since it still mixes and confuses "token ranges" and "token tasks". 
Specifically I suspect that with that patch there would still be num_shards token ranges repaired in a single repair task.

karol-kokoszka:
@vladzcloudius the PR ensuring single (or predefined number) job to be executed is already merged to branch-3.1.
What do you want to achieve with that PR ? I'm not gonna merge that.

vladzcloudius:
> @vladzcloudius the PR ensuring single (or predefined number) job to be executed is already merged to branch-3.1. What do you want to achieve with that PR ? I'm not gonna merge that.

@karol-kokoszka The PR you are referring is not solving the problem in question which is: restore the documented SM repair functionality and constraints.

Your PR does something else - something only you understand.

We are going to have a call with Tzach first thing next week and me and Asias are going to insist on reverting your PR. See more details on corresponding GH issues.

As to this PR - this is a demonstration how the documented functionality can be restored. 

cc @mykaul @harel-z @ruthea

vladzcloudius:
@asias take a glimpse. This PR is supposed to implement your proposal from the "Repair 2023 Design".

The only thing that is missing is the way to fetch `max_repair_ranges_in_parallel` values from nodes.
AFAICT today it's not exposed anywhere except for a syslog, right?
If "yes", then I'll open a GH issue and send a `scylla` PR for that adds a corresponding REST API.

@eliransin you probably want to keep and eye on this as well.

asias:
> @asias take a glimpse. This PR is supposed to implement your proposal from the "Repair 2023 Design".
> 
> The only thing that is missing is the way to fetch `max_repair_ranges_in_parallel` values from nodes. AFAICT today it's not exposed anywhere except for a syslog, right? If "yes", then I'll open a GH issue and send a `scylla` PR for that adds a corresponding REST API.

IIRC, SM calculates max_repair_ranges_in_parallel directly using the same method as in scylla core:

auto max_memory_repair = memory::stats().total_memory() * 0.1; // 10% of per shard memory size
max_repair_memory_per_range = 32 * 1024 * 1024;
max_repair_ranges_in_parallel = max_repair_memory / max_repair_memory_per_range   / 4;

We could expose with api as well. 

> @eliransin you probably want to keep and eye on this as well.

vladzcloudius:
> > @asias take a glimpse. This PR is supposed to implement your proposal from the "Repair 2023 Design".
> > The only thing that is missing is the way to fetch `max_repair_ranges_in_parallel` values from nodes. AFAICT today it's not exposed anywhere except for a syslog, right? If "yes", then I'll open a GH issue and send a `scylla` PR for that adds a corresponding REST API.
> 
> IIRC, SM calculates max_repair_ranges_in_parallel directly using the same method as in scylla core:
> 
> auto max_memory_repair = memory::stats().total_memory() * 0.1; // 10% of per shard memory size max_repair_memory_per_range = 32 * 1024 * 1024; max_repair_ranges_in_parallel = max_repair_memory / max_repair_memory_per_range / 4;
> 
> We could expose with api as well.
> 
> > @eliransin you probably want to keep and eye on this as well.

Correct, you can see how I nucked the corresponding call in this PR. I didn't like it. But maybe I should bring it back for now. Still better than a hard coded value...;) 
I think that there must be a single "source of thruth" for this value and it must be scylla.
If tomorrow you change scylla logic around it it must be transparent to SM.

pkg/service/repair/service.go
```diff
@@ -589,16 +585,6 @@ func (s *Service) hostRangeLimits(ctx context.Context, client *scyllaclient.Clie
 	return out, err
 }
 
-func (s *Service) maxRepairRangesInParallel(shards uint, totalMemory int64) int {
```
tarzanek, line 592: revert and use above function, as per Asias this is OK
(it can be reverted to REST call later, so TODO can be moved here)

pkg/service/repair/controller.go
```diff
@@ -202,25 +197,17 @@ func (c *rowLevelRepairController) shouldBlock(hosts []string, intensity float64
...
-	if parallel := c.intensity.Parallel(); parallel != defaultParallel {
-		hash := replicaHash(hosts)
-		if _, ok := c.busyReplicas[hash]; !ok {
-			l := len(c.busyReplicas) + 1
-			if l > parallel || l > c.intensity.MaxParallel() {
-				return false
-			}
-		}
+	// DENY if there is too many repair tasks running
+	if parallel := c.intensity.Parallel(); parallel+1 > c.intensity.MaxParallel() {
```
tarzanek, line 209: you stop considering busy replicas here
isn't parallel a static var?

vladzcloudius: It turns out that this (and the condition above) was a total BS.
And one simply needs to compare `c.allJobs` (counter of currently running repair tasks) vs `c.intensity.Parallel()` - the actual value of `--parallel` for this repair task.

vladzcloudius:
@tarzanek @igorribeiroduarte @asias please, re-review. 
This PR is meant to be merged: I rebased and split it into meaningful patches.

tzach:
I'm worried about this patch updating the default behavior of *all* customers. 
This is something we try not to do in a patch release.  For sure, it makes sense for some customers, but I still need to get a blessing from QA and @asias it makes sense in *all* cases.

I agree with @vladzcloudius we have a difference between docs and logic we need to fix, but the issue has been there for years and works reasonably for most customers. We need to validate we will not harm a subset of the deployments.

pkg/service/repair/controller.go
```diff
@@ -177,57 +173,51 @@ func (c *rowLevelRepairController) TryBlock(hosts []string) (bool, allowance) {
...
-		}
+	// DENY if allJobs is greater than the configured parallelism
+	if c.allJobs+1 > c.parallelLimit() {
+		//fmt.Println("Too high parallelism (vs Parallel). Blocking.")
+		return false
 	}
 
 	// DENY if all workers are busy
 	if c.allJobs == c.maxWorkerCount {
+		//fmt.Println("All workers are busy. Blocking.")
```
tarzanek, line 216: instead of comments isn't there some debug flag or logger (debug) level where you can put this?

tarzanek:
@tzach it should of course be tested, but I don't expect regression, but rather performance improvement of repairs for all customers (since right now it's unpredictable for clusters with more nodes - if you are lucky, your repair won't block with scheduling parallel tasks on same node and it will be done quickly
if you are unlucky, you repair takes a long time due to racing for resources by same shards for multiple repair tasks on same node)
- repair time should get more predictable once this *bug fix* is merged - and we should avoid seeing repair "stuck" for weird reasons
(I wrote bug fix and consider this as a bug fix - since the behaviour IS broken now, it breaks the simple rule which is that 1 node runs only 1 parallel task - and @asias will bless this)

so QA should mainly test this patch in positive sense on e.g. 3 node clusters with lots of cpus (>16 ?) per node and of course classic positive 3 node cluster with low amount of cpus per node

and negative test cases - multi node clusters (3, 4 nodes per rack) again with more cpus
and
Vlad found it on a certain customer setup and it also could be taken as a model (so on top of above there are MVs to make the repair more funny)

mykaul:
> @tzach it should of course be tested, but I don't expect regression, but rather performance improvement of repairs for all customers (since right now it's unpredictable for clusters with more nodes - if you are lucky, your repair won't block with scheduling parallel tasks on same node and it will be done quickly if you are unlucky, you repair takes a long time due to racing for resources by same shards for multiple repair tasks on same node) repair time should get more predictable once this _bug fix_ is merged - and we should avoid seeing repair "stuck" for weird reasons (I wrote bug fix and consider this as a bug fix - since the behaviour IS broken now, it breaks the simple rule which is that 1 node runs only 1 parallel task - and @asias will bless this)
> 
> so QA should mainly test this patch in positive sense on e.g. 3 node clusters with lots of cpus (>16 ?) per node and of course classic positive 3 node cluster with low amount of cpus per node

Ideally, tests should be added to the PR. To begin with.

pkg/service/repair/controller.go
vladzcloudius: I didn't want to create too much "noise" in this PR.
I was planning to do this in the followup patches.
If it bothers you I can drop all this patch.

vladzcloudius:
> > @tzach it should of course be tested, but I don't expect regression, but rather performance improvement of repairs for all customers (since right now it's unpredictable for clusters with more nodes - if you are lucky, your repair won't block with scheduling parallel tasks on same node and it will be done quickly if you are unlucky, you repair takes a long time due to racing for resources by same shards for multiple repair tasks on same node) repair time should get more predictable once this _bug fix_ is merged - and we should avoid seeing repair "stuck" for weird reasons (I wrote bug fix and consider this as a bug fix - since the behaviour IS broken now, it breaks the simple rule which is that 1 node runs only 1 parallel task - and @asias will bless this)
> > so QA should mainly test this patch in positive sense on e.g. 3 node clusters with lots of cpus (>16 ?) per node and of course classic positive 3 node cluster with low amount of cpus per node
> 
> Ideally, tests should be added to the PR. To begin with.

I've tested this PR ma

> I'm worried about this patch updating the default behavior of _all_ customers. This is something we try not to do in a patch release. For sure, it makes sense for some customers, but I still need to get a blessing from QA and @asias it makes sense in _all_ cases.

What are you talking about, @tzach?

> 
> I agree with @vladzcloudius we have a difference between docs and logic we need to fix, but the issue has been there for years and works reasonably for most customers. 

No it didn't, Tzach. We have customers complaining about "too aggressive repairs" that impact tail latencies all the time.

Please, check the amount of corresponding GH issues on the matter.

> We need to validate we will not harm a subset of the deployments.

Could you, please, be more specific about what you think can harm and what?

vladzcloudius:
> > @tzach it should of course be tested, but I don't expect regression, but rather performance improvement of repairs for all customers (since right now it's unpredictable for clusters with more nodes - if you are lucky, your repair won't block with scheduling parallel tasks on same node and it will be done quickly if you are unlucky, you repair takes a long time due to racing for resources by same shards for multiple repair tasks on same node) repair time should get more predictable once this _bug fix_ is merged - and we should avoid seeing repair "stuck" for weird reasons (I wrote bug fix and consider this as a bug fix - since the behaviour IS broken now, it breaks the simple rule which is that 1 node runs only 1 parallel task - and @asias will bless this)
> > so QA should mainly test this patch in positive sense on e.g. 3 node clusters with lots of cpus (>16 ?) per node and of course classic positive 3 node cluster with low amount of cpus per node
> 
> Ideally, tests should be added to the PR. To begin with.

Ideally R&D would be fixing it and not me, @mykaul.

I've tested it manually quite extensively and there are quite a lot of CI tests maintained by our SM mainteners. If some new tests need to be added - we have the whole QA and SM teams for that. I'm not familiar with the SM testing framework but I can definitely explain how the corresponding test can be defined, @mykaul.

To remind you, @mykaul @tzach, we need these issues to be fixed ASAP since we are bleeding in the field. And it was expected that R&D would do the fixing. However you know exactly what is going on with that.
So, I spent a few hours on my time on that to write and test this thing. However I expect the relevant R&D (!!!) members do what's left - like tests.
I really don't have too much bandwidth to learn your testing frameworks and work on tests.

vladzcloudius:
> I'm worried about this patch updating the default behavior of _all_ customers. This is something we try not to do in a patch release. For sure, it makes sense for some customers, but I still need to get a blessing from QA and @asias it makes sense in _all_ cases.
> 
> I agree with @vladzcloudius we have a difference between docs and logic we need to fix, but the issue has been there for years and works reasonably for most customers. We need to validate we will not harm a subset of the deployments.

And most importantly, @tzach, what "patch release" you are referring here?
This is a PR towards the 'master'. You are welcome to backport it to whatever release you want.

vladzcloudius:
@tzach @mykaul unless you have specific comments about this PR's code i suggest we merge it to the master and then your teams can test it, backport it or not.

I'll take a look on failed CI tests before that and if these failures are related to this PR i'll fix tests. 

Any objections?

vladzcloudius:
For example, the same tests that were used/added in this (https://github.com/scylladb/scylla-manager/pull/3496) PR would cover the this PR as well, @mykaul

vladzcloudius:
@karol-kokoszka could you tell me how I can either see the detailed output of the failed test of run the test myself?

vladzcloudius:
@karol-kokoszka @Michal-Leszczynski could you, please, tell me how I can see the detailed output of the failed CI test(s) or (preferably) if I can run these tests locally?

vladzcloudius:
FYI: the `make test` fails with a clean `master` with the following error:

```
ok      github.com/scylladb/scylla-manager/v3/pkg/util/tickrun  (cached)        coverage: 100.0% of statements
--- FAIL: TestTodayMidnight (0.00s)
    timeutc_test.go:13: invalid hour 2023-08-08 20:00:00 -0400 EDT
FAIL
```

And I don't see how it can possibly pass unless you run the above in the UTC timezone.

Michal-Leszczynski:
Just so it's clear, SM devs are also preparing a fix for that (and many other repair issues #3423) and we plan to have implementation finished by the end of the week.

> @karol-kokoszka @Michal-Leszczynski could you, please, tell me how I can see the detailed output of the failed CI test(s) or (preferably) if I can run these tests locally?

I usually just go to actions, select your gh actions run, select failed test scenario, cogwheel, view raw logs and then I look for "--- FAIL" in the output. Running all the tests locally is just:
```
make start-dev-env
make check
make test
```
Testing only given pkg can be done by `make pkg-integration-test PKG=./pkg/service/repair`

tzach:
> Just so it's clear, SM devs are also preparing a fix for that (and many other repair issues https://github.com/scylladb/scylla-manager/issues/3423) and we plan to have implementation finished by the end of the week.

Great!

Is the solution all in one PR / Commit?
It would be helpful to backport only the fix for #3463 to 3.1 to give a quick solution for user issue and keep the rest of the improvement to 3.2

pkg/service/repair/controller.go
```diff
@@ -3,6 +3,7 @@
 package repair
 
 import (
+	//"fmt"
```
mykaul, line 6: ?

pkg/service/repair/controller.go
```diff
@@ -177,57 +173,51 @@ func (c *rowLevelRepairController) TryBlock(hosts []string) (bool, allowance) {
...
 	if !c.Busy() {
 		return true
 	}
 
-	// DENY if any host has max nr of jobs already running
+	// DENY if any host has a job running already. We don't allow more than a single repair job per host.
 	for _, h := range hosts {
-		if c.jobs[h] >= c.limits[h].Default {
+		if c.jobs[h] >= 1 {
+			//fmt.Printf("all jobs: %d, Too many jobs for %s (%d). Blocking.\n", c.allJobs, h, c.jobs[h])
```
mykaul, line 195: ?

pkg/service/repair/service.go
```diff
@@ -565,9 +566,15 @@ func (s *Service) hostRangeLimits(ctx context.Context, client *scyllaclient.Clie
...
 
+		// Since we don't issue repairs for per-shard ranges anymore 'intensity' becomes irrelevant
+		// Until we do a proper cleanup of 'intensity' related code let's restrict its ability
+		// to change the number of token ranges and allow only a single token range per host at a time
+		// as our documentation mandates.
 		v := rangesLimit{
-			Default: int(shards),
-			Max:     s.maxRepairRangesInParallel(shards, totalMemory),
+			Default: 1,
+			//TODO: collect max_repair_ranges_in_parallel from scylla hosts using a not-yet-existing API
```
mykaul, line 575:
```suggestion
			//TODO: collect max_repair_ranges_in_parallel from scylla hosts using the new API (see https://github.com/scylladb/scylladb/pull/14886 )
```

tarzanek (approved):
reconsider if fmt debug patch is needed, otherwise lgtm

vladzcloudius:
> Just so it's clear, SM devs are also preparing a fix for that (and many other repair issues #3423) and we plan to have implementation finished by the end of the week.

Not sure what you are trying to say here, @Michal-Leszczynski.
This PR is fixing 2 issues out of those you referenced.
If you have comments on this PR you should state them.

Re-doing the same thing somebody already did is counterproductive in general.

> 
> > @karol-kokoszka @Michal-Leszczynski could you, please, tell me how I can see the detailed output of the failed CI test(s) or (preferably) if I can run these tests locally?
> 
> I usually just go to actions, select your gh actions run, select failed test scenario, cogwheel, view raw logs and then I look for "--- FAIL" in the output. Running all the tests locally is just:
> 
> ```
> make start-dev-env
> make check
> make test
> ```
> 
> Testing only given pkg can be done by `make pkg-integration-test PKG=./pkg/service/repair`

I'll try. Thanks.

pkg/service/repair/controller.go
vladzcloudius: resolved (I removed the last patch from this PR that adds those printouts).
I'll consider sending a followup with a proper logging.

pkg/service/repair/service.go
vladzcloudius: Let me merge this.

pkg/service/repair/controller.go
vladzcloudius: I removed the last patch from the original PR that was adding these printouts.

pkg/service/repair/controller.go
```diff
@@ -92,19 +92,14 @@ func (c *defaultController) allowance(hosts []string) allowance {
...
 	for _, h := range hosts {
 		var v int
 		switch {
 		case i == maxIntensity:
 			v = c.limits[h].Max
-		case i < 1:
-			v = c.limits[h].Default
 		default:
-			v = int(i) * c.limits[h].Default
+			v = int(i)
```
asias, line 102: So here, if i == 0, we set it to max_repair_ranges_in_parallel. Otherwise, we set it to the number specified by the user.

pkg/service/repair/controller.go
```diff
@@ -177,50 +172,39 @@ func (c *rowLevelRepairController) TryBlock(hosts []string) (bool, allowance) {
...
 	// ALLOW if nothing is running
 	if !c.Busy() {
 		return true
 	}
 
-	// DENY if any host has max nr of jobs already running
+	// DENY if any host has a job running already. We don't allow more than a single repair job per host.
 	for _, h := range hosts {
-		if c.jobs[h] >= c.limits[h].Default {
+		if c.jobs[h] >= 1 {
```
asias, line 193: I am curious what was c.limits[h].Default?

pkg/service/repair/controller.go
```diff
@@ -177,50 +172,39 @@ func (c *rowLevelRepairController) TryBlock(hosts []string) (bool, allowance) {
...
-		default:
-			ok = r <= int(intensity)*c.limits[h].Default && r <= c.limits[h].Max
-		}
-		if !ok {
-			return false
-		}
+	// DENY (Sanity) if allJobs + 1 > MaxParallel.
+	// However, this should never happen due to "up to 1 repair task per replica" and the fact that MaxParallel is
+	// calculated the way that allows up to 1 repair task per host.
+	if c.allJobs+1 > c.intensity.MaxParallel() {
```
asias, line 201: I assume c.allJobs tracks the number of pending jobs that are currently running. 

nit: We can use >= to avoid the plus one thing. Saying something in the comments like, we already have c.allJobs running and we can not have more than MaxParallel() jobs, so we need to block the request. 
```
if c.allJobs >= c.intensity.MaxParallel()  {
}
```

pkg/service/repair/controller.go
```diff
@@ -177,50 +172,39 @@ func (c *rowLevelRepairController) TryBlock(hosts []string) (bool, allowance) {
...
+		return false
```
asias, line 202: Btw, it seems the function returns a reversed state. shouldBlock() returns false to perform the  block? Why?

pkg/service/repair/worker.go
```diff
@@ -80,13 +79,7 @@ func (w *worker) Run(ctx context.Context) error {
...
 
 func (w *worker) handleJob(ctx context.Context, job job) error {
-	var err error
-	if w.repairType == TypeRowLevel && job.Allowance.ShardsPercent == 0 {
-		err = w.rowLevelRepair(ctx, job)
-	} else {
-		err = w.legacyRepair(ctx, job)
-	}
-	return err
+	return w.rowLevelRepair(ctx, job)
```
asias, line 82: It is safe to drop the legacy partition level repair. All supported releases have row level repair now.

asias:
I am not familiar with the SM code base, but the change looks reasonable to me. I have some comments inline. 

It is good that the number of changes is limited. 

```
$ git diff --stat 9b3527fc538381216ec7076a033f8d7363e91c6d
 pkg/service/repair/controller.go      | 92 +++++++++++++++++++++++++++++++++-----------------------------------------------------------
 pkg/service/repair/controller_test.go |  4 +---
 pkg/service/repair/service.go         | 20 ++++++++++++++------
 pkg/service/repair/worker.go          | 61 +------------------------------------------------------------
 4 files changed, 49 insertions(+), 128 deletions(-)
```

pkg/service/repair/service.go
```diff
@@ -565,9 +566,15 @@ func (s *Service) hostRangeLimits(ctx context.Context, client *scyllaclient.Clie
...
 		}
 
+		// Since we don't issue repairs for per-shard ranges anymore 'intensity' becomes irrelevant
+		// Until we do a proper cleanup of 'intensity' related code let's restrict its ability
+		// to change the number of token ranges and allow only a single token range per host at a time
+		// as our documentation mandates.
 		v := rangesLimit{
-			Default: int(shards),
-			Max:     s.maxRepairRangesInParallel(shards, totalMemory),
+			Default: 1,
```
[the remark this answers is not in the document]
vladzcloudius: Great. That was exactly my idea. We are defaulting to `1` which is both to preserve the documented behavior and to be on a conservative side by default - this can be seen as a "least noisy fastest mode (since `parallel` defaults to 0)".

And then we can change the intensity all the way up to Max (I tested that it works manually).

pkg/service/repair/controller.go
vladzcloudius: This wasn't my decision - I'm just fixing what was written before me ;)
But I agree, some functions names do not match their semantics in a straight forward way.

But logic wise the code does what it's supposed to: https://github.com/scylladb/scylla-manager/blob/master/pkg/service/repair/controller.go#L49

pkg/service/repair/controller.go
vladzcloudius: Correct the result would be the same.
However I followed the previous code approach on this and wanted to keep different condition blocks (some of when remained) to look similar.

And hence the +1 thingy.
In practice, this is not a real-time application anyway so it's not about the performance to begin with.

And I kinda think that `if current_counter+1> max_counter; then {don't do something} else {++current_counter;}` is more readable than `>=` option - it semantically hints that we are checking a condition that if held is going to cause the `current_counter` to be incremented - thereby self-explaining its (`current_counter`) semantics.

pkg/service/repair/controller.go
vladzcloudius: AFAIR it's initialized in this hunk: https://github.com/scylladb/scylla-manager/pull/3506#discussion_r1289704818

So, technically, before this PR we allowed up to "number of shards" concurrent repair jobs in a single replica (other conditions aside).

vladzcloudius: AFAIR it's initialized in this hunk: https://github.com/scylladb/scylla-manager/pull/3506#discussion_r1289704818

So, technically, before this PR we allowed up to "number of shards" concurrent repair jobs in a single replica (other conditions aside).

This was the central part of the "bug" - SM code was mixing the "repair ranges math" with the "repair jobs" math.

It most probably came from the original repair code that was per-shard-ranges based. And probably it made sense then. But for a row-level repair things can and should be significantly simplified.

pkg/service/repair/controller.go
vladzcloudius: correct

pkg/service/repair/controller.go
```diff
@@ -177,50 +172,39 @@ func (c *rowLevelRepairController) TryBlock(hosts []string) (bool, allowance) {
 	return true, a
 }
 
+func (c *rowLevelRepairController) parallelLimit() int {
+	p := c.intensity.Parallel()
+
+	if p <= defaultParallel {
```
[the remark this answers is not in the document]
vladzcloudius: correct again!

vladzcloudius:
> I am not familiar with the SM code base, but the change looks reasonable to me. I have some comments inline.
> 
> It is good that the number of changes is limited.
> 
> ```
> $ git diff --stat 9b3527fc538381216ec7076a033f8d7363e91c6d
>  pkg/service/repair/controller.go      | 92 +++++++++++++++++++++++++++++++++-----------------------------------------------------------
>  pkg/service/repair/controller_test.go |  4 +---
>  pkg/service/repair/service.go         | 20 ++++++++++++++------
>  pkg/service/repair/worker.go          | 61 +------------------------------------------------------------
>  4 files changed, 49 insertions(+), 128 deletions(-)
> ```

I responded to all your comments, @asias.
Please, let me know if there are other questions/comments.

vladzcloudius:
@tgrabiec your comments were addressed too. Please, re-review.

pkg/service/repair/controller.go
asias: Yes, it is existing issue and is outside the scope of the PR.

pkg/service/repair/controller.go
asias: No, this does not bother me much and I was not mentioning this for any perf reasons. I just want to make sure I understand the code correctly (the plus one thing).

pkg/service/repair/controller.go
asias: Yes, this is the most important change in this PR.

asias (approved):

tarzanek (approved):

Michal-Leszczynski:
@vladzcloudius can we close this PR?

Issue #3505, opened 2023-07-26
[Doc]: contradicting content in the repair page


vladzcloudius:
Page URL: https://manager.docs.scylladb.com/stable/sctool/repair.html

*Description*
The description of newly added `--single-host-parallelism` parameter (https://manager.docs.scylladb.com/stable/sctool/repair.html#single-host-parallelism) and the existing `--parallel` parameter (https://manager.docs.scylladb.com/stable/sctool/repair.html#id8) are contradicting.

The former suggests that there may be more than a single task running on a given host, while the later clearly states this is forbidden.

This is related to https://github.com/scylladb/scylla-manager/issues/3463

vladzcloudius:
FYI @tzach @annastuchlik

annastuchlik:
@karol-kokoszka It looks like the sentence "Each node can take part in at most one repair at any given moment." in the description or `--parallel` is incorrect. Please confirm, and I'll remove it.

karol-kokoszka:
@annastuchlik This sentence `Each node can take part in at most one repair at any given moment.` is not correct. We are not ensuring it.

vladzcloudius:
> @annastuchlik This sentence `Each node can take part in at most one repair at any given moment.` is not correct. We are not ensuring it.

@karol-kokoszka it's the bug in SM that we must fix - not document it: https://github.com/scylladb/scylla-manager/issues/3463

@annastuchlik the sentence in question is correct. The need to remove the `--single-host-parallelism` from our documentation ASAP.

There are still discussions about the design which mentions this parameter and this parameter is very much not liked by me and @asias.

Again, I suggest we remove this from our public documentation till further notice.

vladzcloudius:
@annastuchlik and of course, please, revert the https://github.com/scylladb/scylla-manager/pull/3509.

vladzcloudius:
cc @mykaul

Michal-Leszczynski:
As the flag `--single-host-parallelism` isn't a part of SM 3.2 release, I think we can close this issue.
#3509 was effectively reverted by the #3526.

PR #3508, opened 2023-07-27
doc: fix the description of repair parallel

annastuchlik:
Fixes https://github.com/scylladb/scylla-manager/issues/3505

This commit fixes the description for the `parallel` option of `sctool repair`:
- The incorrect information about one-repair-per-node is removed.
- The language is fixed for better clarity.

@karol-kokoszka I think it should be backported to version 3.1 and 3.0.

### DIFF
--- a/docs/source/sctool/partials/sctool_repair.yaml
+++ b/docs/source/sctool/partials/sctool_repair.yaml
@@ -105,10 +105,10 @@ options:
   default_value: "0"
   usage: |
     The maximum number of Scylla repair jobs that can run at the same time (on different token ranges and replicas).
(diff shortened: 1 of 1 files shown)


karol-kokoszka (approved):

karol-kokoszka (changes requested):
fix should be done in different place, sctool_repair.yaml is the generated file
I'll fix that

karol-kokoszka:
I won't fix that, because it comes from fork.
Let me create another PR.

karol-kokoszka:
Closing in favor of https://github.com/scylladb/scylla-manager/pull/3509

PR #3518, opened 2023-08-09
repair: ensure max single repair job per host

### LINKED ISSUE #3425: `--parallel` flag to ensure single repair job per host

karol-kokoszka:
Fixes #3425 

---

Please make sure that:

### DIFF
--- a/pkg/service/repair/controller.go
+++ b/pkg/service/repair/controller.go
@@ -4,7 +4,6 @@ package repair
 
 import (
 	"math"
-	"strconv"
 )
 
 // controller informs generator if repair can be executed on a given set of
@@ -16,23 +15,16 @@ type controller interface {
 	MaxWorkerCount() int
 }
 
-// rowLevelRepairController is a specialised controller for row-level repair.
-// While defaultController sets token ranges equal to the number of shards times
-// intensity in one allowance the rowLevelRepairController splits that to the
-// number of shards into chunks that can be executed independently.
-// Each node can handle at most one repair task per shard.
-// The number of ranges returned in allocation typically equals the intensity.
-// If parallel is set it caps the number of distinct replica sets that can be
-// repaired at the same time.
+// rowLevelRepairController is a responsible for ensuring that just a single repair
+// job is scheduled against single host. It guarantees as well that no more than X (--parallel flag)
+// replicas are repaired in parallel.
+// It's responsible for controlling amount of token ranges included into the job.
 type rowLevelRepairController struct {
 	intensity      *intensityHandler
 	limits         hostRangesLimit
 	maxWorkerCount int
-
-	allJobs      int
-	jobs         map[string]int
-	busyRanges   map[string]int
-	busyReplicas map[uint64]int
+	allJobs        int
+	jobs           map[string]int
 }
 
 var _ controller = &rowLevelRepairController{}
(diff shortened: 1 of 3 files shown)


pkg/service/repair/controller.go
```diff
@@ -74,7 +74,7 @@ func (c *rowLevelRepairController) shouldBlock(hosts []string, intensity int) bo
 
 	// DENY if any host has max nr of jobs already running
 	for _, h := range hosts {
-		if c.jobs[h] >= c.limits[h].Default {
+		if c.jobs[h] >= 1 {
```
mykaul, line 77: nit: the comment in line 75 is now inaccurate.

karol-kokoszka: thanks, updated the comment + removed another checks that are not relevant anymore

pkg/service/repair/controller_test.go
```diff
@@ -199,44 +155,3 @@ func TestRowLevelRepairController(t *testing.T) {
 		}
 	})
 }
-
-func TestRowLevelRepairControllerIssue2446(t *testing.T) {
```
vladzcloudius, line 203: Why this test was removed?

pkg/service/repair/controller_test.go
```diff
@@ -47,125 +47,81 @@ func (s controllerTestSuite) newRowLevelRepairController() *rowLevelRepairContro
...
-		if ok, _ := ctl.TryBlock([]string{"a", "b", "c"}); !ok {
+		const intensity = 1000
+
+		s.intensityHandler.SetIntensity(ctx, intensity)
+
+		ok, a := ctl.TryBlock([]string{"a", "b", "c"})
+		if !ok {
 			t.Fatal("TryBlock() failed to block")
 		}
+		if a.Ranges != intensity {
```
vladzcloudius, line 62: Should be ">" - exactly as the error message below states.
Repair task is allowed to pick less than "intensity" ranges, e.g. in the last hunk.

vladzcloudius (changes requested):

vladzcloudius:
There is already a PR that fixes this and other issues which was submitted some time ago: https://github.com/scylladb/scylla-manager/pull/3506

This PR pretty much copies what this patch (https://github.com/scylladb/scylla-manager/pull/3506/commits/8020beb4db314e598ae1f42d83b2e405a3e892d9) from the PR above does in terms of SM fix.

Don't you think that it makes more sense to pick the code part from the PR above and leave only the test part from this PR?

vladzcloudius:
> This PR pretty much copies what this patch (https://github.com/scylladb/scylla-manager/commit/8020beb4db314e598ae1f42d83b2e405a3e892d9) from the PR above does in terms of SM fix.

I need to take this back - it didn't. It should have though ;)

karol-kokoszka:
Closing as https://github.com/scylladb/scylla-manager/pull/3521 is ready

Issue #3514, opened 2023-08-04
Scylla Manager server crashes when starting deleted task


Michal-Leszczynski:
Steps to reproduce:
```
[miles@fedora scylla-manager]$ ./sctool repair -c myc
repair/fe31623e-dfea-40c4-80a3-36d08f16b622
[miles@fedora scylla-manager]$ ./sctool stop repair/fe31623e-dfea-40c4-80a3-36d08f16b622 -c myc --delete
[miles@fedora scylla-manager]$ ./sctool start repair/fe31623e-dfea-40c4-80a3-36d08f16b622 -c myc
```
SM logs:
```
{"L":"INFO","T":"2023-08-04T10:19:01.743+0200","N":"http","M":"PUT /api/v1/cluster/myc/task/repair/fe31623e-dfea-40c4-80a3-36d08f16b622/stop?disable=false","from":"127.0.0.1:45568","status":0,"bytes":0,"duration":"4ms","_trace_id":"jJvfQeQSQ8mRgyRlXY7KYA"}
{"L":"INFO","T":"2023-08-04T10:19:01.749+0200","N":"scheduler","M":"Task deleted","cluster_id":"dad4c44e-e04a-4747-9175-55ecc6c1b172","task_type":"repair","task_id":"fe31623e-dfea-40c4-80a3-36d08f16b622","_trace_id":"tpm9rnPmQFS2ZsZKfF_Lwg"}
{"L":"INFO","T":"2023-08-04T10:19:01.749+0200","N":"http","M":"DELETE /api/v1/cluster/myc/task/repair/fe31623e-dfea-40c4-80a3-36d08f16b622","from":"127.0.0.1:45568","status":0,"bytes":0,"duration":"5ms","_trace_id":"tpm9rnPmQFS2ZsZKfF_Lwg"}
{"L":"INFO","T":"2023-08-04T10:19:11.444+0200","N":"http","M":"PUT /api/v1/cluster/myc/task/repair/fe31623e-dfea-40c4-80a3-36d08f16b622/start?continue=true","from":"127.0.0.1:53550","status":0,"bytes":0,"duration":"3ms","_trace_id":"bkMpQziQQFKC_fnrDr01Rg"}
{"L":"INFO","T":"2023-08-04T10:19:11.444+0200","N":"scheduler.00000000","M":"Run started","task":"/00000000-0000-0000-0000-000000000000","retry":0,"_trace_id":"GTWubiPhSS-qG_m3M0MCHg"}
[...]

PR #3524, opened 2023-08-14
fix(repair): update params of disabled tasks using its names only

### LINKED ISSUE #3513: Make sure it's possible to re-enable disabled task using task name instead of just the task ID
I previously used the command "sctool stop -c my-cluster-name --disable repair/all-weekly" to temporarily disable repairs while doing some work on my cluster, and now I want to enable it, but can't find anything in the docs on how to do this.

The "sctool start" command doesn't seem to have an option for this. "sctool tasks resume" also doesn't seem to do the trick.

I tried to modify an example like this "sctool repair update -c my-cluster-name repair/all-weekly --enabled=true", but that returned "Error: no tasks of type repair". But if I do "sctool tasks -a", I do see the repair/all-weekly task in the list with an asterisk in front of it.

Let me know the trick, thanks.

karol-kokoszka:
Fixes #3513 

---

Please make sure that:

### DIFF
--- a/pkg/managerclient/client.go
+++ b/pkg/managerclient/client.go
@@ -411,6 +411,7 @@ func (c *Client) taskByName(ctx context.Context, clusterID, taskType, taskName s
 		ClusterID: clusterID,
 		Type:      &taskType,
 		Short:     pointer.BoolPtr(true),
+		All:       pointer.BoolPtr(true),
 	})
 	if err != nil {
 		return uuid.Nil, err


pkg/managerclient/client.go
```diff
@@ -406,11 +406,13 @@ func (c *Client) uniqueTaskID(ctx context.Context, clusterID, taskType string) (
 }
 
 func (c *Client) taskByName(ctx context.Context, clusterID, taskType, taskName string) (uuid.UUID, error) {
+	all := true
 	resp, err := c.operations.GetClusterClusterIDTasks(&operations.GetClusterClusterIDTasksParams{
 		Context:   ctx,
 		ClusterID: clusterID,
 		Type:      &taskType,
 		Short:     pointer.BoolPtr(true),
+		All:       &all,
```
Michal-Leszczynski, line 415: Why not use `pointer.BoolPtr(true)` like above?

karol-kokoszka: shame on me :) fixing...

Michal-Leszczynski (approved):
My only concern would be about handling deleted/disabled tasks in the other parts of code. I believe that this change makes it possible to run into #3514 when using task name, but this can still happen now when using task ID and it's a bug in different parts of code.

PR #3529, opened 2023-08-18
Improve repair tests

Michal-Leszczynski:
This PR adds test checking:
- repair pause and resume
- repair completeness (all token ranges are repaired)

And it also improves the test checking:
- repair order 
- repair table by table

### DIFF
--- a/pkg/scyllaclient/client_scylla.go
+++ b/pkg/scyllaclient/client_scylla.go
@@ -940,11 +940,11 @@ func (c *Client) DisableAutoCompaction(ctx context.Context, keyspace, table stri
 }
 
 // FlushTable flushes writes stored in MemTable into SSTables stored on disk.
-func (c *Client) FlushTable(ctx context.Context, keyspace, table string) error {
+func (c *Client) FlushTable(ctx context.Context, host, keyspace, table string) error {
 	_, err := c.scyllaOps.StorageServiceKeyspaceFlushByKeyspacePost(&operations.StorageServiceKeyspaceFlushByKeyspacePostParams{
 		Cf:       &table,
 		Keyspace: keyspace,
-		Context:  ctx,
+		Context:  forceHost(ctx, host),
 	})
 	return err
 }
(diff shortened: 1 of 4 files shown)


tzach:
We need to test repair with MV, particularly MV with small cardinality. 
@ShlomiBalalis @fgelcer FYI

mykaul:
> We need to test repair with MV, particularly MV with small cardinality. @ShlomiBalalis @fgelcer FYI

Not sure if it's part of https://github.com/scylladb/qa-tasks/issues/1450 or unrelated (in core?)

Michal-Leszczynski:
This PR is just a standalone SM repair test improvement.
This is supposed to partially make up with the lack of more sophisticated dtests / SCT for SM 3.2 repair.
Fortunately, those extended tests didn't find any bugs.

> We need to test repair with MV, particularly MV with small cardinality.
Tests in this PR only validate if SM sent requests with all possible token ranges (even in case of pause / resume) and how it handles failures returned by Scylla. So it does not validate if MV has indeed been repaired, but only if a "SUCCESSFUL" repair has been performed on all ranges.

pkg/scyllaclient/client_scylla.go
```diff
@@ -940,11 +940,11 @@ func (c *Client) DisableAutoCompaction(ctx context.Context, keyspace, table stri
 }
 
 // FlushTable flushes writes stored in MemTable into SSTables stored on disk.
-func (c *Client) FlushTable(ctx context.Context, keyspace, table string) error {
+func (c *Client) FlushTable(ctx context.Context, host, keyspace, table string) error {
 	_, err := c.scyllaOps.StorageServiceKeyspaceFlushByKeyspacePost(&operations.StorageServiceKeyspaceFlushByKeyspacePostParams{
 		Cf:       &table,
 		Keyspace: keyspace,
-		Context:  ctx,
+		Context:  forceHost(ctx, host),
```
dkropachev, line 947: I would suggest to move changes related to the tests in one commit, and fixes to the components in other, so that it would be easier to go over changes history in future.

dkropachev (changes requested):

pkg/scyllaclient/client_scylla.go
Michal-Leszczynski: done

dkropachev (approved):

karol-kokoszka (approved):

Issue #3528, opened 2023-08-17
[Repair]: control allows setting `intensity` greater than the maximum allowed


vladzcloudius:
SM version: 3.2
Cluster: 9 nodes (3 nodes in each of 3 racks)
RF: 3
max_repair_ranges_in_parallel: 2

*Description*
`sctool control` allowed to set `intensity` value 7 and SM actually sent a repair job with 7 token ranges to scylla:

Scylla log:
```
shard 0 stats: repair_reason=repair, keyspace=keyspace1, tables={standard1}, ranges_nr=7, round_nr=27, round_nr_fast_path_already_..
```

SM log:
```
Aug 17 23:19:11 diego-repair-monitor scylla-manager[70342]: {"L":"INFO","T":"2023-08-17T23:19:11.489Z","N":"repair.worker 2","M":"Repairing","keyspace":"keyspace1","table":"
standard1","master":"10.142.0.127","hosts":["10.142.0.118","10.142.0.121","10.142.0.127"],"ranges":[{"StartToken":-5968249633527889392,"EndToken":-5962250548062071906},{"Sta
rtToken":-5962250548062071906,"EndToken":-5938265995976221708},{"StartToken":-5875134265137665159,"EndToken":-5872655845602208870},{"StartToken":-5872655845602208870,"EndTok
en":-5864665924136304192},{"StartToken":-5273923641784723347,"EndToken":-5263337718691042465},{"StartToken":-5244482280309978570,"EndToken":-5241737648028349783},{"StartToke
n":-5241737648028349783,"EndToken":-5241509209815145555}],"job_id":3527,"_trace_id":"dvHUhjB_QUm_yF4E9Ynf3w"}
```

While @asias told us that it's safe and `scylla` is going to repair only up to `max_repair_ranges_in_parallel` ranges at a time in a single repair job it's very confusing to a user that assumes that he/she actually increased the repair intensity to 7 while in practice any value greater than 2 is not going to change anything.

I suggest to either trim the actually stored value to the maximum allowed or alternatively fail the corresponding `control` command with an "invalid value" error.

Michal-Leszczynski:
I think that trimming it internally could be even more confusing, so I would go with the validation error approach.
It's worth to remember that `max_repair_ranges_in_parallel` might be different for each node in the cluster.
[...]

PR #3536, opened 2023-08-24
epair params validation

Michal-Leszczynski:
---

Please make sure that:


Michal-Leszczynski:
After giving it some thoughts, I think that returning error on too big `--intensity` or `--parallel` isn't the best idea.
The reason for that is that we can set those parameters only once while repair task will run with them multiple times. Changes in cluster topology, availability or schema can influence current max parallel/intensity value, so even though some intensity was correct at one point, it can be to big at another one.

Because of that, I decided to change approach and only trim actually used intensity/parallel values inside `controller.go`.
In order to make up for that, I extended repair progress display with max values for intensity/parallel:
```
[miles@fedora scylla-manager]$ ./sctool progress -c myc repair/c45996d9-144f-46b2-bb0b-1daaae54c296
Run:            c26cb65a-4787-11ee-a9b6-0892040e83bb
Status:         RUNNING
Start time:     30 Aug 23 18:51:30 EDT
Duration:       31s
Progress:       84%
Intensity:      1/1 (max)
Parallel:       2/3
```
where `max` means that the special value 0 was used.
Those max values can be used in a scenario described in this [comment](https://github.com/scylladb/scylla-manager/issues/3528#issuecomment-1690873757).

The first 3 commits remained the same, only the 3 last ones (previously about validation) were changed.
@karol-kokoszka what do you think about this approach?

karol-kokoszka:
You mean, if someone creates repair task scheduled every couple of days, then the intensity and parallel will stay the same.
That's correct.

AFAIR it shouldn't be a problem if we send too much of the token ranges in a single call (it's about intensity param).
And it is not a problem if someone sets parallel too high, as the real value is limited by (nr_node / RF) already.

Never the less, I'm totally fine with silently lowering down the intensity value at runtime.

vladzcloudius:
> Intensity:      1/1 (max)
> Parallel:       2/3

This `max` thingy is confusing. You seem to always print the maximum value (the value on the right) and the requested value (the value on the left) - IMO you don't need to print any extra marker that denotes "left value is greater or equal to the right value".

Michal-Leszczynski:
The idea was to print `(max)` when intensity/parallel are set to special value 0. I thought that's perhaps better than writing 0/1 in such case. So if max parallel equals 3 and user runs repair with parallel 3, the `(max)` is not printed.
Here is a [link](https://github.com/scylladb/scylla-manager/pull/3563/files#diff-35d4c54ec728b4297dce77df3c8a5b8d9b8627484c6d8aaf06ff2a52ed20651e) to a PR with `sctool progress` example explaining this interpretation.

vladzcloudius:
> The idea was to print `(max)` when intensity/parallel are set to special value 0. I thought that's perhaps better than writing 0/1 in such case. So if max parallel equals 3 and user runs repair with parallel 3, the `(max)` is not printed. Here is a [link](https://github.com/scylladb/scylla-manager/pull/3563/files#diff-35d4c54ec728b4297dce77df3c8a5b8d9b8627484c6d8aaf06ff2a52ed20651e) to a PR with `sctool progress` example explaining this interpretation.

Like I have already explained - this is confusing in my opinion.

PR #3539, opened 2023-08-25
Remind to check version compatibility before running restore

### LINKED ISSUE #3533: docs: Issue in page Restore
I would like to report an issue in page https://manager.docs.scylladb.com/branch-3.1/sctool/restore

### Problem

https://manager.docs.scylladb.com/stable/sctool/restore.html#restore-schema
should explain that restoring system* tables (besides system_auth) IS a BIG problem if versions are NOT compatible
e.g. restoring 2022.2 system tables to 5.2 is a disaster
even 2023 (despite it being built on top of 5.2) restored to 5.2 WILL be a problem (since OSS lacks EE features)

###  Suggest a fix

we should explicitely and in bold warn people to use this options ONLY with compatible versions (so same flavour, EE, OSS, and only lower to higher or same to same version, anything else will break the node(s) )

Michal-Leszczynski:
Fixes #3533

### DIFF
--- a/docs/source/restore/index.rst
+++ b/docs/source/restore/index.rst
@@ -20,6 +20,11 @@ Restore and backups are scheduled in the same manner: you can start, stop, resum
 
 .. note:: Restore procedure works with any cluster topologies, so backed-up cluster can have different number of nodes than restore destination cluster.
 
+.. warning::
+    Restore works only when ScyllaDB version of backed-up cluster is the same or older than ScyllaDB version of restore destination cluster.
+
+    Mixing ScyllaDB Enterprise and ScyllaDB Open Source versions could lead to unexpected results.
+
 Restore types
 =============
 


Michal-Leszczynski:
I decided that both types of restore should be marked by this warning as it seems safer and I doubt we have any cross-version restore tests. Is this acceptable?

tarzanek:
well look @Michal-Leszczynski 
don't hesitate to marked ANYTHING that is not TESTED as UNSUPPORTED

unexpected results might be tempting for someone to try ... and ... he/she will fail, so why giving him the temptation?
Also we are pretty clear that major versions for SURE don't have compatible system tables - so there it's a clear UNSUPPORTED
with minor you might have a small chance, but why trying it?

docs/source/restore/index.rst
```diff
@@ -20,6 +20,11 @@ Restore and backups are scheduled in the same manner: you can start, stop, resum
 
 .. note:: Restore procedure works with any cluster topologies, so backed-up cluster can have different number of nodes than restore destination cluster.
 
+.. warning::
+    Restore works only when ScyllaDB version of backed-up cluster is the same or older than ScyllaDB version of restore destination cluster.
+
+    Mixing ScyllaDB Enterprise and ScyllaDB Open Source versions could lead to unexpected results.
```
tarzanek, line 26: mark as unsupported for direction
Enterprise -> OSS
(the other way it might work, but still, will you test?)
system tables are not compatible between different versions, PERIOD

official path for Enterprise to OSS is CQL migration

that said
user tables will be compatible as per SSTABLE compatibility support matrix - https://opensource.docs.scylladb.com/stable/architecture/sstable/

that's it, nothing else people need to know or try

karol-kokoszka: I agree that it's better to explicitly say 'unsupported' on Enterprise -> OSS and vice versa, although the `could lead to unexpected results.` have the same meaning.
@annastuchlik please advise if the "Mixing ScyllaDB Enterprise and ScyllaDB Open Source versions could lead to unexpected results." should be rephrased.

karol-kokoszka (approved):

PR #3541, opened 2023-08-28
Restore: run post-restore repair with max intensity

Michal-Leszczynski:
As cluster shouldn't be under any load during restore procedure, this should make it run more efficient.

### DIFF
--- a/pkg/service/backup/restore_worker_tables.go
+++ b/pkg/service/backup/restore_worker_tables.go
@@ -275,7 +275,9 @@ func (w *tablesWorker) stageRepair(ctx context.Context, run *RestoreRun, _ Resto
 		}
 	}
 	repairProps, err := json.Marshal(map[string]any{
-		"keyspace": keyspace,
+		"keyspace":  keyspace,
+		"intensity": 0,
+		"parallel":  0,
 	})
 	if err != nil {
 		return errors.Wrap(err, "parse repair properties")


mykaul:
Is that for any restore? Full restore only? Where is the issue it is solving?

Michal-Leszczynski:
> Is that for any restore? Full restore only? Where is the issue it is solving?

Created corresponding issue with more explanation. Please let me know if it make sense.

karol-kokoszka:
@Michal-Leszczynski it makes sense what's written here https://github.com/scylladb/scylla-manager/issues/3542, but the repair task have a command that lets to change the intensity and parallel at runtime -> https://manager.docs.scylladb.com/stable/sctool/repair.html#repair-control
This possibility of changing the intensity at runtime is a safety valve. 

I'm OK with introducing this change to repair with max intensity only if the same mechanism is introduced to the repair task executed as a post-restore repair.

Michal-Leszczynski:
Right now post restore repair task isn't treated as a regular task, so `sctool repair control` won't work on it.
Should we make it possible?

karol-kokoszka:
We should make it possible.

Michal-Leszczynski:
Ok, so I will close this PR and introduce the possibility of using `sctool repair control` on post restore repair task with default max intensity in a separate PR.

PR #3544, opened 2023-08-28
Repair: make repair system keyspace first more robust

### LINKED ISSUE #3543: Repair: don't use whitelist as the only source of internal tables
Right now, when SM creates table repair order, it distinguishes internal keyspaces from regular ones only via predefined list (which also gives different priorities to different internal tables). If keyspace cannot be found in this list, SM treats it as a regular one. In this case, it would be better if SM also checked for `system` prefix. Right now this list might be enough, but it would be nice to make the ordering process more robust in case a new internal keyspace is introduced in Scylla.

Michal-Leszczynski:
Fixes #3543

### DIFF
--- a/pkg/service/repair/preference.go
+++ b/pkg/service/repair/preference.go
@@ -2,6 +2,8 @@
 
 package repair
 
+import "strings"
+
 // TablePreference describes partial predefined order in which tables should be repaired.
 type TablePreference interface {
 	// KSLess compares priorities of two keyspaces according to preference.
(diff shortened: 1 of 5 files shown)


pkg/service/repair/preference.go
```diff
@@ -0,0 +1,74 @@
...
+func NewInternalTablePreference() TablePreference {
+	return internalTablePreference{
+		{
+			Keyspace: "system_auth",
+			Tables: []string{
+				"role_attributes",
+				"role_members",
+			},
+		},
+		{Keyspace: "system_traces"},
```
mykaul, line 28: nit: system_traces can be the last, it's less important than any other table.

pkg/service/repair/preference.go
```diff
@@ -0,0 +1,74 @@
...
+			if t == t1 {
+				return true
+			}
+			if t == t2 {
+				return false
+			}
+		}
+	}
+	return false
+}
```
karol-kokoszka, line 74: Could be easily unit tested.
Please add test cases for KLess and TLess.

pkg/service/repair/preference.go
Michal-Leszczynski: done

pkg/service/repair/preference.go
Michal-Leszczynski: done

PR #3552, opened 2023-08-31
fix(healtcheck): make native cql to retry if connection is failed

### LINKED ISSUE #3550: cqlping.NativeCQLPing should have retry

dkropachev:
Fixes #3550 

Customers complain that healthcheck fails on cqlping having almost no load on the cluster and network working perfectly fine.

### DIFF
--- a/pkg/ping/cqlping/cqlping.go
+++ b/pkg/ping/cqlping/cqlping.go
@@ -5,15 +5,18 @@ package cqlping
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
 	"net"
 	"strconv"
 	"strings"
 	"time"
 
-	"github.com/gocql/gocql"
 	"github.com/pkg/errors"
 
+	"github.com/gocql/gocql"
+	"github.com/scylladb/go-log"
 	"github.com/scylladb/scylla-manager/v3/pkg/ping"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/retry"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/timeutc"
 )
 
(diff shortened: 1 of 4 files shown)


karol-kokoszka:
Native CQL ping is executed when scylla-manager doesn't hold CQL credentials to the cluster.
The customer who complains is the cloud one or the enterprise one ? 
All cloud clusters have credentials available already.

I'm not sure if the retry solves the problem. It couldn't connect and it's reported as FAIL what is correct.

dkropachev:
> Native CQL ping is executed when scylla-manager doesn't hold CQL credentials to the cluster. The customer who complains is the cloud one or the enterprise one ? All cloud clusters have credentials available already.
> 
> I'm not sure if the retry solves the problem. It couldn't connect and it's reported as FAIL what is correct.

Customer pointed to exactly that case, they mentioned that hey don't have credentials.
But despite that, failure of singular tcp connection request should not trigger healthcheck failure.

karol-kokoszka:
This code snippet shows that manager is logging the exact error on CQL ping failure.
Customer can check if it's the `timeout` error or something else https://github.com/scylladb/scylla-manager/blob/e51531ddb257bf19f217a60cd4922ff6975b8b8f/pkg/ping/error.go#L9

https://github.com/scylladb/scylla-manager/blob/e51531ddb257bf19f217a60cd4922ff6975b8b8f/pkg/service/healthcheck/service.go#L219-L226

The solution you want to apply makes sense only if the root cause is not the timeout on establishing connection, but something else, like immediate rejection.

Have a look here https://github.com/scylladb/scylla-manager/blob/e51531ddb257bf19f217a60cd4922ff6975b8b8f/pkg/ping/cqlping/cqlping.go#L49-L63

^^ dialer keeps exactly the same timeout as what is passed to the backoff = `config.Timeout`
`	backoff := retry.NewExponentialBackoff(time.Millisecond, config.Timeout, config.Timeout/3, 2, 0)`

If the root cause is some immediate rejection, then we will have it in logs.

karol-kokoszka:
Although, I'm fine with the PR, just fix the sanity-checks that are failing.

dkropachev:
> This code snippet shows that manager is logging the exact error on CQL ping failure. Customer can check if it's the `timeout` error or something else
> 
> https://github.com/scylladb/scylla-manager/blob/e51531ddb257bf19f217a60cd4922ff6975b8b8f/pkg/ping/error.go#L9
> 
> https://github.com/scylladb/scylla-manager/blob/e51531ddb257bf19f217a60cd4922ff6975b8b8f/pkg/service/healthcheck/service.go#L219-L226
> 
> The solution you want to apply makes sense only if the root cause is not the timeout on establishing connection, but something else, like immediate rejection.
> 
> Have a look here
> 
> https://github.com/scylladb/scylla-manager/blob/e51531ddb257bf19f217a60cd4922ff6975b8b8f/pkg/ping/cqlping/cqlping.go#L49-L63
> 
> ^^ dialer keeps exactly the same timeout as what is passed to the backoff = `config.Timeout` ` backoff := retry.NewExponentialBackoff(time.Millisecond, config.Timeout, config.Timeout/3, 2, 0)`
> 
> If the root cause is some immediate rejection, then we will have it in logs.

Sure, I have already asked to provide additional information on the issue.

dkropachev:
@karol-kokoszka , all linter complaints addressed, test is fixed, fail you see on backup is not related to the PR.

pkg/ping/cqlping/cqlping.go
```diff
@@ -33,21 +34,30 @@ var options = []byte{4, 0, 0, 0, 5, 0, 0, 0, 0}
...
 // It returns rtt and error.
 func NativeCQLPing(ctx context.Context, config Config) (rtt time.Duration, err error) {
+	backoff := retry.NewExponentialBackoff(time.Millisecond, config.Timeout, config.Timeout/3, 2, 0)
+	deadline := timeutc.Now().Add(config.Timeout)
 	t := timeutc.Now()
-	defer func() {
-		rtt = timeutc.Since(t)
-		if rtt >= config.Timeout {
-			err = ping.ErrTimeout
+	for attempt := 0; ; attempt++ {
```
karol-kokoszka, line 40: what is this variable for ? 
I see no checks made on `attempt`.
It can be replaced with for {}

pkg/ping/cqlping/cqlping.go
```diff
@@ -33,21 +34,30 @@ var options = []byte{4, 0, 0, 0, 5, 0, 0, 0, 0}
...
+		err = nativeCQLPingOnce(ctx, config, deadline)
+		if err == nil {
+			return timeutc.Since(t), nil
 		}
-	}()
+		duration := backoff.NextBackOff()
+		if duration == retry.Stop {
+			return timeutc.Since(t), ping.ErrTimeout
+		}
+		time.Sleep(duration)
```
karol-kokoszka, line 49: Please log the information about retries (attempt variable may be useful then).

dkropachev: done

pkg/ping/cqlping/cqlping.go
dkropachev: it is reused for the logging

karol-kokoszka (approved):

mykaul:
I'm hoping this doesn't hide issues...

PR #3558, opened 2023-09-07
fix(backup): generate a proper UUIDv1 for sstable identifier

tchaikov:
before this change, `RandomSSTableUUID()` just return a serial of random bits formatted like 'xxxx_xxxx_xxxxxxxxxxxxxxxxxx', where 'x' is picked from the alphabet of
"abcdefghijklmnopqrstuvwxyz0123456789".
but the generated UUID cannot be decoded by scylla, as it does not encode a UUID v1 (as per RFC4122). that's why scylla chokes when parsing the UUID created by scylla-manager, and throws `std::out_of_range (stoull)`.

in this change, we

* add github.com/google/uuid as an explicit dependency
* leverage github.com/google/uuid to generate a UUID V1, and format it with the same encoding used by scylla's `generation_type`.

Fixes https://github.com/scylladb/scylladb/issues/15226
Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>

---

Please make sure that:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/gocql/gocql v1.5.2
 	github.com/golang/mock v1.6.0
 	github.com/google/go-cmp v0.5.5
+	github.com/google/uuid v1.1.2
 	github.com/hailocab/go-hostpool v0.0.0-20160125115350-e80d13ce29ed
 	github.com/hashicorp/go-version v1.2.0
 	github.com/hbollon/go-edlib v1.5.0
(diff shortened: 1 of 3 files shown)


tchaikov:
@dkropachev hi Dimitry, sorry i failed to identify the problem while reviewing #3461 . could you please help review this change?

karol-kokoszka (approved):

karol-kokoszka:
We will include this fix into 3.2.1 patch release that we want to trigger today.

Michal-Leszczynski (approved):

PR #3569, opened 2023-09-12
chore: bump to go1.21

karol-kokoszka:
This PR bumps golang to 1.21 version.

---

Please make sure that:


Michal-Leszczynski (approved):
Nice! Let's fix the checks and merge it.

PR #3570, opened 2023-09-12
Healthcheck: don't report -1 in ping metrics when agent is down

### LINKED ISSUE #3555: CQL status helthcheck metric shows -1 even when node responds via CQL
Healthcheck tasks use different types of pings and set status metrics on success (1) and failure (-1).
Current implementation of CQL ping queries agent for node info and only after that performs proper CQL ping.
So when agent is down, CQL ping will end up in failure even when Scylla is up and responds via CQL. The same goes for other types of pings (rest, alternator).

Is this expected behavior?
Perhaps it's misleading that -1 metric value can mean that either agent or Scylla are not responding.
Maybe it would be better to introduce new metric value (-2) which would indicate that agent is down and proper ping cannot be executed?

Michal-Leszczynski:
Fixes #3555

### DIFF
--- a/pkg/scyllaclient/client_ping.go
+++ b/pkg/scyllaclient/client_ping.go
@@ -3,10 +3,13 @@
 package scyllaclient
 
 import (
+	"bytes"
 	"context"
+	"fmt"
 	"math"
 	"math/rand"
 	"net"
+	"net/http"
 	"net/url"
 	"runtime"
 	"sort"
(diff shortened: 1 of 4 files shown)


Michal-Leszczynski:
Example metrics when agent `192.168.200.11` is down.
```
# HELP scylla_manager_healthcheck_cql_status Host native port status. -2 stands for unavailable agent, -1 for unavailable Scylla and 1 for everything is fine.
# TYPE scylla_manager_healthcheck_cql_status gauge
scylla_manager_healthcheck_cql_status{cluster="694d541d-0520-4067-9ea0-437a3b03edc5",host="192.168.200.11"} -2
scylla_manager_healthcheck_cql_status{cluster="694d541d-0520-4067-9ea0-437a3b03edc5",host="192.168.200.12"} 1
scylla_manager_healthcheck_cql_status{cluster="694d541d-0520-4067-9ea0-437a3b03edc5",host="192.168.200.13"} 1
scylla_manager_healthcheck_cql_status{cluster="694d541d-0520-4067-9ea0-437a3b03edc5",host="192.168.200.21"} 1
scylla_manager_healthcheck_cql_status{cluster="694d541d-0520-4067-9ea0-437a3b03edc5",host="192.168.200.22"} 1
scylla_manager_healthcheck_cql_status{cluster="694d541d-0520-4067-9ea0-437a3b03edc5",host="192.168.200.23"} 1
```

Michal-Leszczynski:
Regarding improvement suggested in [comment](https://github.com/scylladb/scylla-manager/issues/3555#issuecomment-1708360660), I decided not to add it for now. Although this could work just fine for `healthcheck` tasks which use `nodeInfo` only for different types of pings, this could mess up the output of `sctool status` which also uses `nodeInfo` for information like uptime, memory, versions etc. Extending `nodeInfo` TTL on successful ping could easily lead to outdated information in `sctool status` and create confusion on user side.

karol-kokoszka (approved):

PR #3572, opened 2023-09-12
feat(scheduler): add suspended metric (#3567)

### LINKED ISSUE #3567: Have a metric of cluster suspension
Have a metric that takes 1 one cluster is suspended and takes 0 when it is not

Michal-Leszczynski:
Fixes #3567

### DIFF
--- a/pkg/metrics/scheduler.go
+++ b/pkg/metrics/scheduler.go
@@ -8,6 +8,7 @@ import (
 )
 
 type SchedulerMetrics struct {
+	suspended    *prometheus.GaugeVec
 	runIndicator *prometheus.GaugeVec
 	runsTotal    *prometheus.GaugeVec
 	lastSuccess  *prometheus.GaugeVec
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
@karol-kokoszka Could you take a look at that and also answer the question?

> Is it important to have this metric deliberately set to 0 when cluster is not suspended, or are we ok with no value for this metric in this case?

karol-kokoszka:
You should set the value if the cluster is not suspended as well.
Otherwise how do you want to indicate that the cluster is resumed already ? Do you want to reset this metric?

https://manager.docs.scylladb.com/stable/sctool/suspend-resume.html

karol-kokoszka:
OK, there is another scenario where the cluster just is and it was never suspended/resumed.
It's ok to have no value for the metric.

karol-kokoszka (approved):

Issue #3568, opened 2023-09-12
Give better explanation on host down error during task creation


karol-kokoszka:
With the latest changes, scylla-manager just throws basic "connection refused" error to the output whenever host is down on task creation.

The message should be more self-explanatory.
We need to replace it with 
```
Node XXXX is down. Task cannot be created.
Please use --ignore-host-down flag (https://manager.docs.scylladb.com/stable/sctool/repair.html#ignore-down-hosts) if you want to skip the repair on this node.
```

Michal-Leszczynski:
I'm just wondering about safety of using `--ignore-host-down` flag ion general. If we skip repair of down node and then it comes back to cluster it could cause data resurrection, right? So is suggesting to use this flag a good approach?

karol-kokoszka:
We are not the administrators of the customers database.
We are just delivering the tool they can use to achieve what they want.

I don't think there is a hidden suggestion in the sentence. It just lists one of the options customer can use to proceed with the repair of UN nodes.

PR #3574, opened 2023-09-12
Better explanation on host down

karol-kokoszka:
Fixes https://github.com/scylladb/scylla-manager/issues/3568

CLI error message is extended and contains information about the reason why task creation failed.

```
One of the nodes is down. Task cannot be created.
Check --ignore-down-hosts flag if you wish to exclude the node from the repair.

Error: create repair target: create repair plan: calculate tables size: fetch table disk size report
 192.168.200.12: giving up after 2 attempts: agent [HTTP 502] 
invalid response from host
Trace ID: Qs_yl1nOSXiWTwr0J7Oh6Q (grep in scylla-manager logs)
```

---

Please make sure that:

### DIFF
--- a/pkg/command/repair/cmd.go
+++ b/pkg/command/repair/cmd.go
@@ -4,9 +4,9 @@ package repair
 
 import (
 	_ "embed"
+	"errors"
 	"fmt"
 
-	"github.com/pkg/errors"
 	"github.com/scylladb/scylla-manager/v3/pkg/command/flag"
 	"github.com/scylladb/scylla-manager/v3/pkg/managerclient"
 	"github.com/spf13/cobra"
--- a/pkg/restapi/task.go
+++ b/pkg/restapi/task.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/render"
 	"github.com/pkg/errors"
+	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
 	"github.com/scylladb/scylla-manager/v3/pkg/service"
 	"github.com/scylladb/scylla-manager/v3/pkg/service/backup"
 	"github.com/scylladb/scylla-manager/v3/pkg/service/repair"
@@ -232,6 +233,11 @@ func (h *taskHandler) getTarget(w http.ResponseWriter, r *http.Request) {
 		}
 	case scheduler.RepairTask:
 		if t, err = h.Repair.GetTarget(r.Context(), newTask.ClusterID, p); err != nil {
+			if errors.Is(err, scyllaclient.ErrHostInvalidResponse) {
+				respondError(w, r, err, "One of the nodes is down. Task cannot be created.\n"+
+					"Check --ignore-down-hosts flag if you wish to exclude the node from the repair.")
+				return
+			}
 			respondError(w, r, errors.Wrap(err, "get repair target"))
 			return
 		}
@@ -295,6 +301,11 @@ func (h *taskHandler) createTask(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := h.validateTask(r.Context(), newTask, p); err != nil {
+		if newTask.Type == scheduler.RepairTask && errors.Is(err, scyllaclient.ErrHostInvalidResponse) {
+			respondError(w, r, err, "One of the nodes is down. Task cannot be created.\n"+
+				"Check --ignore-down-hosts flag if you wish to exclude the node from the repair.")
+			return
+		}
 		respondError(w, r, err)
 		return
 	}
@@ -329,6 +340,11 @@ func (h *taskHandler) updateTask(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := h.validateTask(r.Context(), newTask, newTask.Properties); err != nil {
+		if newTask.Type == scheduler.RepairTask && errors.Is(err, scyllaclient.ErrHostInvalidResponse) {
+			respondError(w, r, err, "One of the nodes is down. Task cannot be updated.\n"+
+				"Check --ignore-down-hosts flag if you wish to exclude the node from the repair.")
+			return
+		}
 		respondError(w, r, err)
 		return
 	}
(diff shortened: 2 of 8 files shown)


pkg/managerclient/client.go
```diff
@@ -193,7 +198,7 @@ func (c *Client) GetRepairTarget(ctx context.Context, clusterID string, t *Task)
 		TaskFields: makeTaskUpdate(t),
 	})
 	if err != nil {
-		return nil, err
+		return nil, errors.Join(err, ErrClientError)
 	}
```
Michal-Leszczynski, line 202: Why would you want to treat all `GetRepairTarget` errors as `ErrClientError`?

Michal-Leszczynski: The same for other places where this error is joined.

pkg/command/repair/cmd.go
```diff
@@ -65,7 +65,14 @@ func newCommand(client *managerclient.Client, update bool) *command {
...
 	cmd.RunE = func(_ *cobra.Command, args []string) error {
-		return cmd.run(args)
+		err := cmd.run(args)
+		if errors.Is(err, managerclient.ErrClientError) {
+			fmt.Fprintf(cmd.OutOrStderr(), "One of the nodes is down. Task cannot be created/updated.\n"+
+				"Please use --ignore-down-hosts flag (https://manager.docs.scylladb.com/stable/sctool/repair.html#ignore-down-hosts)"+
+				" if you want to skip the repair on this node.\n\n")
+		}
+
+		return err
```
Michal-Leszczynski, line 75: Is this link needed here? If user want to get this information, it can be obtained with `--help`. Not to say that the information is `Do not repair nodes that are down i.e. in status DN.`.

Also, I think that ifing out errors both on server and cli tool side is a double work. Can't just return a descriptive error directly from server?

Michal-Leszczynski (commented):
Wouldn't it be easier to just return descriptive error from `GetTarget`? We can do that by simply checking if all hosts include in plan are in UN state before trying to calculate table size, which fails on downed nodes.

pkg/command/repair/cmd.go
karol-kokoszka: There is a tiny problem in returning the descriptive error from server directly.
The normal approach is just to chain errors and return them the way
`error1 : error2 : .... etc` 

It's not the most humane readable way of informing end-user about the root cause of the problem.

To make it human-readable, the error must be parsed and print out in bit better organized way.

karol-kokoszka: I will change it bit different way.
Create the general error = "host is down". Put it in chain and look for this error on parsing the output on Sctool repair cmd.

pkg/restapi/task.go
```diff
@@ -232,6 +233,11 @@ func (h *taskHandler) getTarget(w http.ResponseWriter, r *http.Request) {
 		}
 	case scheduler.RepairTask:
 		if t, err = h.Repair.GetTarget(r.Context(), newTask.ClusterID, p); err != nil {
+			if errors.Is(err, scyllaclient.ErrHostInvalidResponse) {
+				respondError(w, r, err, "One of the nodes is down. Task cannot be created.\n"+
+					"Check --ignore-down-hosts flag if you wish to exclude the node from the repair.")
+				return
+			}
```
Michal-Leszczynski, line 240: It's rather difficult to maintain this approach when the same error message is present at 3 separate places.

Also, this approach requires ifing the error in 3 places now:
- in the client function
- in create/update/dry-run of a task
- in the CLI

Isn't it more complicated than just returning human readable error message from `GetTarget` directly?

pkg/scyllaclient/client_scylla.go
```diff
@@ -5,6 +5,7 @@ package scyllaclient
 import (
 	"bytes"
 	"context"
+	stdErrors "errors"
```
Michal-Leszczynski, line 8: Why do we have two error pkg imported here?

pkg/scyllaclient/client_scylla.go
```diff
@@ -862,7 +866,7 @@ func (c *Client) TableDiskSizeReport(ctx context.Context, hostKeyspaceTables Hos
 
 		size, err := c.TableDiskSize(ctx, v.Host, v.Keyspace, v.Table)
 		if err != nil {
-			return parallel.Abort(errors.Wrapf(err, v.Host))
+			return parallel.Abort(errors.Wrapf(stdErrors.Join(err, ErrHostInvalidResponse), v.Host))
```
Michal-Leszczynski, line 869: So this error should be included in every client function, but right now we include it only in 1 repair specific call?
This sounds like it could create confusion in the future.

Michal-Leszczynski (commented):
> There is a tiny problem in returning the descriptive error from server directly.
The normal approach is just to chain errors and return them the way
error1 : error2 : .... etc

But in this case we had only:
> Error: create repair target: create repair plan: calculate tables size: fetch table disk size report
 192.168.200.12: giving up after 2 attempts: agent [HTTP 502] 
invalid response from host

So we could change host down validation to obtain something similar to:
`Error: create repair target: node 192.168.200.12 is down, check --ignore-down-hosts flag if you wish to exclude the node from the repair.`

To do that, it's enough to:
- exclude `FillSize` from `newPlan`
- create `newPlan`
- iterate over plan's hosts and check their UN state - return error when there is a down node and `--ignore-hosts-down` isn't set.
- `FillSize`

I just afraid that currently proposed approach would be difficult to maintain, because it creates 3 separate places where errors have to be ifed out (see comment).

pkg/restapi/task.go
Michal-Leszczynski: Although there isn't really any ifing out in the CLI itself.

pkg/scyllaclient/client_scylla.go
karol-kokoszka: No, it's just an option. You can include it, if you want to have a possibility to check one of the root causes.

Michal-Leszczynski:
@karol-kokoszka but if you think that currently proposed approach is better than what I suggested above, then I'm ok with merging it.

pkg/restapi/task.go
karol-kokoszka: But this message is returned from the API call directly.
I need to if here, because I must find what's the root cause of an error and what to describe to the end user.

Michal-Leszczynski (approved):

PR #3575, opened 2023-09-13
Repair test: extend duration cmp limit (#3573)

### LINKED ISSUE #3573: Flaky `TestProgressManagerIntegration/progress_update_sequence_(Init,OnJobStart,OnJobEnd)`
`TestProgressManagerIntegration/progress_update_sequence_(Init,OnJobStart,OnJobEnd)` randomly failed on `5.2.4` for IPv6.
 
Output:
```
=== RUN   TestProgressManagerIntegration
=== RUN   TestProgressManagerIntegration/progress_update_sequence_(Init,OnJobStart,OnJobEnd)
--- When: run progress is initialized with incomplete values
--- Then: progress rows are initialized with zero values
--- When: OnJobStart called on progress manager
--- Then: run progress is updated with starting times
--- When: OnJobEnd is called on progress manager
--- Then: progress is updated with success and duration
    progress_integration_test.go:155:   []repair.RunProgress{
          	{
          		... // 9 identical fields
          		StartedAt:         s"2023-09-12 18:00:27.069 +0000 UTC",
          		CompletedAt:       nil,
        - 		Duration:          s"5.119714ms",
        + 		Duration:          s"6.220017ms",
          		DurationStartedAt: nil,
          		... // 1 ignored field
          	},
          	{
          		... // 9 identical fields
          		StartedAt:         s"2023-09-12 18:00:27.069 +0000 UTC",
          		CompletedAt:       nil,
        - 		Duration:          s"5.119714ms",
        + 		Duration:          s"6.220017ms",
          		DurationStartedAt: nil,
          		... // 1 ignored field
          	},
          }
```

Michal-Leszczynski:
As we can see in the issue, this test can sometimes fail on duration comparison limit. Those tests were originally written for local testing envs and were rekicked in case of failures, so I think that simply increasing the timeout should do the trick.

Fixes #3573

### DIFF
--- a/pkg/service/repair/progress_integration_test.go
+++ b/pkg/service/repair/progress_integration_test.go
@@ -30,7 +30,7 @@ func TestProgressManagerIntegration(t *testing.T) {
 		cmpopts.IgnoreUnexported(RunProgress{}),
 		UUIDComparer(),
 		NearTimeComparer(5 * time.Millisecond),
-		NearDurationComparer(1 * time.Millisecond),
+		NearDurationComparer(5 * time.Millisecond),
 	}
 
 	t.Run("progress update sequence (Init,OnJobStart,OnJobEnd)", func(t *testing.T) {


Michal-Leszczynski:
@karol-kokoszka could you take a look at that?

karol-kokoszka (approved):

PR #3582, opened 2023-09-15
Collect agent memory profile

Michal-Leszczynski:
This commit allows for collecting agent memory profiles (heap, allocs, goroutines). In order to collect profiles:
- `scylla-manager-agent.yaml` has to have `debug` field specified
- profiles are saved in `/tmp/pprof/<profile-type>/<cnt>` agent files, so they have to be extracted from agent before destroying the instance

Currently profiles are collected every 30 min and only 60 newest profiles are saved. Those profiles can be analyzed separately or as a delta.

### DIFF
--- a/pkg/cmd/agent/server.go
+++ b/pkg/cmd/agent/server.go
@@ -5,9 +5,13 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
+	"io"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
+	"path"
 	"runtime"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/agent/server.go
```diff
@@ -207,14 +211,101 @@ func (s *server) startServers(ctx context.Context) {
...
+	defer profileFile.Close()
+
+	if _, err := io.Copy(profileFile, resp.Body); err != nil {
+		return errors.Wrap(err, "save profile "+filePath)
+	}
+	return nil
+}
+
+func profilePath(profileType string, cnt int) string {
+	return path.Join("/tmp/pprof/", profileType, fmt.Sprint(cnt))
```
mykaul, line 306: This is not a good idea. You need to create a safe temporary dir/file. https://pkg.go.dev/io/ioutil#TempFile ?

karol-kokoszka:
@Michal-Leszczynski pls make sure that the profiles are preserved after the SCT test executiion and they are just not removed. I guess some changes in SCT are needed ?  Maybe @dkropachev @ShlomiBalalis can help to answer what changes may be required ?

PR #3585, opened 2023-09-24
Expand repair docs

### LINKED ISSUE #3576: docs: Issue in page Repair
I would like to report an issue in page https://manager.docs.scylladb.com/branch-3.2/sctool/repair

### Problem
Please provide additional clarity on --parallelism 0 setting.
Eg:
"The formula to calculate is is as follows: number of nodes / RF, ex. for 6 node cluster with RF=3 the maximum parallelism is 2."
1. The repair is across all keyspaces and different keyspaces may have different RF.  How does this affect repair parallelism calculation?  Is each keyspace repaired with different degree of parallelism, or something else?

2. How does this equation work in a multi-DC situation?  If I have 6 DC's with 9 nodes, am I using 57 nodes?  Is the RF from each DC for a keyspace SUM'd?  Eg. a table in every DC with RF = 3 - then it's 57/18?  

###  Suggest a fix
Please clarify the equation in light of above comments.

It would be nice if when default parallelism is used, the max parallelism used during the repair was reported in progress and history.  It makes it much easier/faster for a user to slow things down if they see how fast it's going.

### LINKED ISSUE #3578: docs: Issue in page Progress
I would like to report an issue in page https://manager.docs.scylladb.com/branch-3.2/sctool/progress

### Problem
How to generate progress is documented but interpreting the output is not.

###  Suggest a fix
Please add example output.  Include explanation success/fail nuance of progress field (which is described in prior release notes, but ideally would be in this section also).  (Thanks @noellymedina )
https://www.scylladb.com/product/release-notes/scylla-manager-release-1-4/

### LINKED ISSUE #3581: 0% Errors in repair progress report can be misleading
When a repair has a high number of ranges and a low number of errors, the report can show 100%/0%, which can be misunderstood as no errors.  It also was not obvious to me on initial use that the presense of the 2nd column meant there were actually errors and the absense meant there were none.

Adding the error column from --details to progress would make things crystal clear.
Alternately, taking the CEIL for errors so if there were even 1 error, the minimum would be 1% not zero percent would make it clear there were errors.

Thank you

Michal-Leszczynski:
Fixes #3578
Fixes #3581
Fixes #3576

### DIFF
--- a/docs/source/repair/index.rst
+++ b/docs/source/repair/index.rst
@@ -42,7 +42,6 @@ Parallel repairs
 
 Each node can take part in at most one Scylla repair job at any given moment, but Scylla Manager can repair distinct replica sets in a token ring in parallel.
 This is beneficial for big clusters.
-For example, a 9 node cluster and a keyspace with replication factor 3, can be repaired up to 3 times faster in parallel.
 The following diagram presents a benchmark results comparing different parallel flag values.
 In a benchmark we ran 9 Scylla 2020.1 nodes on AWS i3.2xlarge machines under 50% load, for details check `this blog post <https://www.scylladb.com/2020/11/12/scylla-manager-2-2-repair-revisited/>`_
 
@@ -51,6 +50,22 @@ In a benchmark we ran 9 Scylla 2020.1 nodes on AWS i3.2xlarge machines under 50%
 
 By default Scylla Manager runs repairs with full parallelism, you can change that using :ref:`sctool repair --parallel flag <sctool-repair>`.
 
+Maximal effective parallelism
+=============================
+
+Max parallelism is determined by:
+  * the constraint that each node can only take part in one ScyllaDB repair job at any given moment.
+  * ScyllaDB repair job targeting the full replica set of the repaired token range.
+
+For example, let's assume a cluster with 2 datacenters, 5 nodes each.
+When you repair the keyspace ``my_keyspace with replication = {'class': 'NetworkTopologyStrategy', 'dc1': 2, 'dc2': 3}``,
+max parallelism is equal to ``1``, because each ScyllaDB repair job targets a full replica set of the repaired token range.
+Every replica set consists of 2 nodes from ``dc1`` and 3 nodes from ``dc2``,
+so it's impossible to schedule 2 repair jobs to run simultaneously (``dc2`` lacks one more node for it to be possible).
+
+Repair is performed table by table and keyspace by keyspace,
+so max effective parallelism might change depending on which keyspace is being repaired.
+
 Repair intensity
 ================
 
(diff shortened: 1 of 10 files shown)


karol-kokoszka (approved):

docs/source/repair/index.rst
```diff
@@ -51,20 +50,43 @@ In a benchmark we ran 9 Scylla 2020.1 nodes on AWS i3.2xlarge machines under 50%
 
 By default Scylla Manager runs repairs with full parallelism, you can change that using :ref:`sctool repair --parallel flag <sctool-repair>`.
 
+Maximal effective parallelism
+=============================
+
+Max parallelism is limited by the ``each node can take part in at most one Scylla repair job at any given moment`` constraint.
```
annastuchlik, line 56: I'd reorganize and reword the description for the sake of clarity:

> Max parallelism is determined by:
> * the constraint that each node can only take part in one ScyllaDB repair job at any given moment.
> * ScyllaDB repair job targeting the full replica set of the repaired token range.
> 
> For example, let's assume a cluster with 2 datacenters, 5 nodes each. When you repair the keyspace ``my_keyspace with replication = {'class': 'NetworkTopologyStrategy', 'dc1': 2, 'dc2': 3}``, max parallelism is equal to ``1``, because each ScyllaDB repair job targets a full replica set of the repaired token range. Every replica set consists of 2 nodes from ``dc1`` and 3 nodes from ``dc2``, so it's impossible to schedule 2 repair jobs to run simultaneously (``dc2`` lacks one more node for it to be possible). 

(``dc2`` lacks one more node for it to be possible) - is it``dc2``?

Also, please add an empty space before
> Repair is performed table by table and keyspace by keyspace,
> so max effective parallelism might change depending on which 
> keyspace is being repaired.

Otherwise, the content displays on the same line (and it shouldn't, as it's not part of the example).

docs/source/repair/index.rst
```diff
@@ -51,20 +50,43 @@ In a benchmark we ran 9 Scylla 2020.1 nodes on AWS i3.2xlarge machines under 50%
...
+
 Repair intensity
 ================
 
 Intensity specifies how many token ranges can be repaired in a Scylla node at every given time.
 The default intensity is one, you can change that using :ref:`sctool repair --intensity flag <sctool-repair>`.
 
 Scylla Manager 2.2 adds support for intensity value zero.
-In that case the number of token ranges is calculated based on node memory and adjusted to the Scylla maximal number of ranges that can be repaired in parallel (see ``max_repair_ranges_in_parallel`` in Scylla logs).
+In that case the number of token ranges is calculated based on node memory and adjusted to the Scylla maximal number of ranges that can be repaired in parallel.
```
annastuchlik, line 72:
```suggestion
In that case, the number of token ranges is calculated based on node memory and adjusted to ScyllaDB's maximum number of ranges that can be repaired in parallel.
```

docs/source/repair/index.rst
```diff
@@ -51,20 +50,43 @@ In a benchmark we ran 9 Scylla 2020.1 nodes on AWS i3.2xlarge machines under 50%
...
 If you want to repair faster, try using intensity zero.
 
 Note that the less the cluster is loaded the more it makes sense to increase intensity.
 If you increase intensity on a loaded cluster it may not give speed benefits since cluster have no resources to process more repairs.
 In our experiments in a 50% loaded cluster increasing intensity from 1 to 2 gives about 10-20% boost and increasing it further will have little impact.
 
+Maximal effective intensity
```
annastuchlik, line 79: How about rewriting as follows (the empty lines are important - if not added, everything will melt into one paragraph after publication).

> Max intensity is calculated based on the ``max_repair_ranges_in_parallel`` value (present in ScyllaDB logs). This value might be different for each node in the cluster.
> 
> The overall max intensity is equal to the maximum ``max_repair_ranges_in_parallel`` value of all repaired nodes.
> 
> However, because each ScyllaDB repair job targets some subset of all nodes and ScyllaDB Manager avoids repairing more than ``max_repair_ranges_in_parallel`` on any node, the max effective intensity for a given repair job is equal to the **minimum** ``max_repair_ranges_in_parallel`` value of nodes taking part in the job.

annastuchlik: Is there any value in saying: "The overall max intensity is equal to the maximum max_repair_ranges_in_parallel value of all repaired nodes"?
I guess there is, so I've left it, but if the information is not useful to the user, let's skip it.

pkg/command/repair/repaircontrol/res.yaml
```diff
@@ -5,16 +5,21 @@ short: Change parameters while a repair is running
 long: |
   This command allows you to modify repair parameters while a repair is running.
   Note that modified parameters apply only to the currently running task and not to the future task runs.
+  Moreover, this change is applied only to newly created Scylla repair jobs (also for currently repaired table).
```
annastuchlik, line 8:
```suggestion
  Moreover, this change is applied only to newly created ScyllaDB repair jobs (and to the currently repaired table).
```

docs/source/repair/index.rst
Michal-Leszczynski: > (dc2 lacks one more node for it to be possible) - is itdc2?

Yes, because `dc2` has 5 nodes and rf=3, so it's not possible to run two repair jobs without using some node in both of them (and SM does not allow for it to happen). If `dc2` had 1 more node, it would be possible.

annastuchlik (commented):
I've left some comments and suggestions.

A general one is to use "ScyllaDB" instead of "Scylla".

docs/source/repair/index.rst
Michal-Leszczynski: I think it can be skipped.

docs/source/repair/index.rst
annastuchlik: Just making sure I understand correctly. Thanks.

PR #3592, opened 2023-09-27
Repair: fix small table optimization on big tables

### LINKED ISSUE #3591: Repair: small table optimization is used on big tables
SM should repair small, fully replicated tables in 1 Scylla repair job, but it happens that bigger tables are also optimized.

Examples:
- SM [issue](https://github.com/scylladb/scylla-manager/issues/3584#issuecomment-1732627868)
- Scylla [issue](https://github.com/scylladb/scylla-enterprise/issues/3400#issuecomment-1720870937)

Cause:
This fragment of `FillSize` method:
```go
	for i, kp := range p.Keyspaces {
		p.Keyspaces[i].Size = ksSize[kp.Keyspace]
		for j, tp := range kp.Tables {
			kp.Tables[j].Size = tableSize[kp.Keyspace+"."+tp.Table]
			// Return merged ranges for small, fully replicated table (#3128)
			if tp.Size <= smallTableThreshold && len(kp.Replicas) == 1 {
				kp.Tables[j].Optimize = true
			}
		}
	}
```
where we set `kp.Tables[j].Size` but then we check `tp.Size` which hasn't been updated.
This basically means that all fully replicated tables (no matter their size) are always optimized (repaired in 1 Scylla repair job) since SM 3.2.

Also, we need to add tests checking that bigger tables are not optimized even when they are fully replicated.

Michal-Leszczynski:
Fixes #3591

### DIFF
--- a/pkg/service/repair/plan.go
+++ b/pkg/service/repair/plan.go
@@ -225,10 +225,10 @@ func (p *plan) FillSize(ctx context.Context, client *scyllaclient.Client, smallT
 
 	for i, kp := range p.Keyspaces {
 		p.Keyspaces[i].Size = ksSize[kp.Keyspace]
-		for j, tp := range kp.Tables {
-			kp.Tables[j].Size = tableSize[kp.Keyspace+"."+tp.Table]
+		for j := range kp.Tables {
+			kp.Tables[j].Size = tableSize[kp.Keyspace+"."+kp.Tables[j].Table]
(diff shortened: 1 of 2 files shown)


karol-kokoszka (approved):

karol-kokoszka:
@Michal-Leszczynski @ShlomiBalalis please validate this branch against the test that found the problem with small table optimization before merging it

PR #3616, opened 2023-10-24
Scyllaclient: increase t/o of listing versioned files

### LINKED ISSUE #3615: Timeout on listing versioned files is to little
This is based on this [issue](https://github.com/scylladb/scylla-enterprise/issues/3462#issuecomment-1768077216).
Our custom implementation of listing remote dir resets timeout on every iteration, but since listing versioned files returns almost no files to iterate over, it's way easier for it to hit the timeout.

Michal-Leszczynski:
Fixes #3615


tarzanek (approved):
lgtm
so we will go to 1h timeout by default, good
ev. it can be configed in yaml file and increased to more

Issue #3611, opened 2023-10-18
New parameter for the repair defining max nr of repair job per node.


karol-kokoszka:
We would like to introduce the possibility to increase the max number of repair jobs executed against a single scylla node using the repair task parameter.

Right now, we have a limit of just a single repair job per node.
We want to leave it as a default, but the possibility to increase this value must be introduced.
Parameter must accept integer values > 0. {1, 2, 3, ...}

Whenever user will set the max nr of repair jobs per scylla node to be higher than 0, then scylla-manager won't respect the limit defined by https://manager.docs.scylladb.com/stable/repair/index.html#maximal-effective-parallelism . The parallelism won't be narrow down to meet what is defined in the maximal-effective-parallelism definition.

tzach:
Is  this the revert a of [Manager 3.2 update](https://forum.scylladb.com/t/release-scylla-manager-3-2-0/815) ?

> This release changes the parallelism and order of the repair job for better performance and stability. The following changes have been made:
>     *  Only one repair job is running on any host at any given time."
>     ...

karol-kokoszka:
@tzach it's not the revert.
Manager < 3.2 didn't respect any limits in terms of parallelism per node.

When we were discussing repair changes in 3.2, one of the ideas was to provide the flag that would tell what is the max number of the repair jobs that can be sent to a single node. https://github.com/scylladb/scylla-manager/issues/3425#issuecomment-1636013035  .  See  We didn't introduce the flag finally, but we decided to hard limit it to 1. See https://github.com/scylladb/scylla-manager/issues/3425#issuecomment-1663614205

Now, seems that some of the customers would like to repair more aggressively even tough they need to sacrifice the performance of the cluster during the repair.
The statement above brings us back to the idea of the flag defining max parallel repair jobs per node.

Michal-Leszczynski:
As this is another parameter controlling repair speed, it should also be integrated with the [repair control](https://manager.docs.scylladb.com/stable/sctool/repair.html#repair-control) command.

Michal-Leszczynski:
What is the expected behavior when setting `--parallel 3` and `--max-jobs-per-node 3` in a cluster where the total maximal parallelism (with 1 job per 1 node rule) is 10?

From the basic flag description, it would look like scheduling 3 repairs on 1 replica set is allowed even though there would be other, underutilized replica sets.

[...]

PR #3623, opened 2023-11-01
Add worker pool to repair

Michal-Leszczynski:
This is the first step in implementing #3611, as changing the approach to repair parallelism (while still integrating it with `sctool repair control`) requires freely control the size of the worker pool, which was lacking in the current implementation.

### DIFF
--- a/pkg/service/repair/controller.go
+++ b/pkg/service/repair/controller.go
@@ -2,8 +2,6 @@
 
 package repair
 
-import "math"
-
 // controller informs generator about the amount of ranges that can be repaired
 // on a given replica set. Returns 0 ranges when repair shouldn't be scheduled.
 type controller interface {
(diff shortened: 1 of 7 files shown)


Michal-Leszczynski:
@karol-kokoszka  could you take a look at that?

karol-kokoszka (approved):
:+1: Please trigger dtests and SCT before merging.

Michal-Leszczynski:
SCT test [passed](https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/centos-sanity-test/1052/).

Michal-Leszczynski:
All SM Dtests are currently failing because of not related [issue](https://github.com/scylladb/scylla-dtest/issues/3729).

Michal-Leszczynski:
Passed [dtests](https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-dtest/1359/).

PR #3625, opened 2023-11-02
Add SM 3.2 to restore compatibility matrix

Michal-Leszczynski:
This change in documentation is already present on master and should be backported to branch 3.2.

Fixes #3617

PR #3626, opened 2023-11-05
Test SM against Scylla 5.4.0-rc1

Michal-Leszczynski:
Addressing [comment](https://github.com/scylladb/scylla-manager/pull/3625#pullrequestreview-1713932887).
Proper changes to tested Scylla versions should be first merged to master and then cherry-picked into `branch-3.2`.


Michal-Leszczynski:
It looks like IPV6 env setup does not work and that restoring versioned schema test fails (it passes for me locally).

PR #3633, opened 2023-11-09
Update s3 compatible storage description with information about custom CAs

### LINKED ISSUE #3632: Update backup docs with information about custom CA support
As per issue title.

karol-kokoszka:
Fixes #3632 

image

---

Please make sure that:


hopugop:
The suggested certificate path and openssl update certs command are debian-specific and do not apply to RHEL or CentOS. Could we please state this is just an example and this applied to debian/ubuntu?

karol-kokoszka:
@annastuchlik your suggestions are included.
@hopugop example for Centos is added

Issue #3628, opened 2023-11-07
Update testing environment to validate support for custom CA on cloud storage


karol-kokoszka:
We would like to verify the scenario where the cloud storage is Minio using custom CA.
According to https://forum.rclone.org/t/how-to-use-a-custom-ca-certificate/8676 , it should be enough for scylla-manager-agent to have this CA available in system CAs.

The goal here is to update the testing environment, so that Minio uses custom CA, all nodes containers keeps the CA in system CA.

It addresses the issue below
https://github.com/scylladb/scylla-enterprise/issues/3534

Issue #3621, opened 2023-10-27
Scylla Manager : enhancement > repair intensity and parallelism values visibility in progress / info / --show-properties


ppalczak:
Please add visibility of repair intensity and parallelism for these 4 scenarios, as it's currently missing
1. sctool tasks --show-properties:
![image](https://github.com/scylladb/scylla-manager/assets/134054072/e612cfb9-5a7c-45d2-81f7-9f001cdb5e16)

2. sctool progress for finished and currently running tasks (for currently running tasks it's not always visible)
![image](https://github.com/scylladb/scylla-manager/assets/134054072/3b3cfb85-14be-41e3-9be0-799ddb7cbac3)

3. sctool progress --run (should display last set value)

4. sctool info repair (to see default value set for repair)
![image](https://github.com/scylladb/scylla-manager/assets/134054072/5d0d6fab-3cd9-4772-ab7b-8ff34a9069e0)

ppalczak:
Visibility of repair intensity and parallelism in currently running task for comparison
![image](https://github.com/scylladb/scylla-manager/assets/134054072/6e07c93c-e261-43cc-ab0a-c7639cb8e968)

Michal-Leszczynski:
1. As described in [SM docs](https://manager.docs.scylladb.com/stable/sctool/task.html#show-properties), `--show-properties` displays only the flags deliberately set by the user. Example:
```
[miles@fedora scylla-manager]$ ./sctool repair -c myc --intensity 20 --parallel 20
[miles@fedora scylla-manager]$ ./sctool tasks --show-properties
Cluster: myc (928bf746-aba9-4046-8c9d-c608ecd90568)
╭─────────────────────────────────────────────┬──────────────┬────────────────────────────────────────┬───────────────┬─────────┬───────┬─────────────────────────┬────────────┬─────────┬─────────────────────────┬──────────────────────────────╮
│ Task                                        │ Schedule     │ Window                                 │ Timezone      │ Success │ Error │ Last Success            │ Last Error │ Status  │ Next                    │ Properties                   │
├─────────────────────────────────────────────┼──────────────┼────────────────────────────────────────┼───────────────┼─────────┼───────┼─────────────────────────┼────────────┼─────────┼─────────────────────────┼──────────────────────────────┤
[...]

Issue #3601, opened 2023-10-08
Restore system_distributed.cdc_generation_descriptions


Michal-Leszczynski:
Real [issue](https://github.com/scylladb/scylla-enterprise/issues/3483).

As SM does not restore `cdc` related tables and blacklists `system_distributed.cdc_generation_descriptions_v2`, it should also blacklist `system_distributed.cdc_generation_descriptions` as it may also exist in Scylla versions compatible with SM restore compatibility matrix.

PR #3638, opened 2023-11-14
Patch release 3.2.4

karol-kokoszka:
Includes:
- https://github.com/scylladb/scylla-manager/issues/3597
- https://github.com/scylladb/scylla-manager/issues/3615
- https://github.com/scylladb/scylla-manager/issues/3278
- https://github.com/scylladb/scylla-manager/issues/3381
- https://github.com/scylladb/scylla-manager/issues/3628
- https://github.com/scylladb/scylla-manager/issues/3165
- https://github.com/scylladb/scylla-manager/issues/3621
- https://github.com/scylladb/scylla-manager/issues/3601
- https://github.com/scylladb/scylla-enterprise/issues/3580

---

Please make sure that:


Michal-Leszczynski:
@karol-kokoszka I understand that this PR was created just by looking over recent commits on master, cherry-picking them and hoping, that nothing was left unintentionally. I find this process annoying and difficult to validate, but unfortunately there is no other way (that I'm aware of), because of different git history caused by additional commits `release: prepare for X.X.X` present only on release branch.

Could we change the management of our release branches, so that release branches contain strictly the same git history up to the release point? The process could look like that:
- we want to make a release which contains all commits from master
- we change Scylla Manager version in master to `X.X.X`
- we rebase release branch on top of master branch
- we change Scylla Manager version in master to `X.X.X-dev`

karol-kokoszka:
@Michal-Leszczynski The current approach gives us a possibility to patch release just a single commits from the master and this is the main benefit.

We do copy all the commits from master whenever we are branching for another minor (3.x) releases.

PR #3658, opened 2023-12-07
Run github actions against Scylla 5.4.0

### LINKED ISSUE #3644: Enable RAFT for integration-tests
Scylla manager docker-compose for integration-tests starts scylla nodes with the RAFT disabled.
We need to introduce another suite where all the integration tests are executed against the nodes with RAFT enabled.

https://www.scylladb.com/2023/05/04/scylladbs-path-to-strong-consistency-a-new-milestone/

### LINKED ISSUE #3654: Github actions failing because cluster setup hangs
Recently some github actions runs are failing because they reach the 6h timeout on cluster creation step. Example:
```
2023-12-05T23:00:23.7152983Z ==> Waiting for first nodes from DC1 and DC2 to spin up
2023-12-06T04:59:54.8577560Z ##[error]The operation was canceled.
```
It would be nice to know if the root cause is connected to:
- Scylla regression (from what I have seen, this only happens with Scylla 5.2.X)
- SM cluster setup script
- github actions runners

### LINKED ISSUE #3656: Update gocql
We use (https://github.com/scylladb/scylla-manager/blob/55fbc265faf6f834c0456937df2591048c640ded/go.mod#L120 ) a VERY old version of our Scylla GoCQL driver! We should update to the latest, on a regular basis.

Michal-Leszczynski:
This PR contains a bunch of changes related to fixing and improving stability of integration tests. The main goal was to introduce Scylla 5.4 to test suite and fix any encountered issues.

This PR also resulted in creating two possible Scylla issues:
- https://github.com/scylladb/scylladb/issues/16349
- https://github.com/scylladb/scylladb/issues/16332

Fixes #3644
Fixes #3654
Fixes #3656


Michal-Leszczynski:
All alternator tests are failing with 5.4.0, but it looks like a [Scylla issue](https://github.com/scylladb/scylladb/issues/16332).

Michal-Leszczynski:
Failing restore schema tests:
- 5.2 - restoring versioned schema fails
- 5.4 - the same but also restoring schema + tables fails

Michal-Leszczynski:
It looks like there is a difference in how restore schema performs.
With Scylla 5.2 when SM restores schema to nodes in dc1 and performs rolling-restart, nodes from dc2 pick up uploaded schema. With Scylla 5.4 nodes from dc2 have the same schema version as before the restore. This looks like the reason for failing schema tests.

Michal-Leszczynski:
Also, from some local experiments, it looks like the trick with `skip_wait_for_gossip_to_settle: 0` works fine with improved cluster setup, but it still creates issues when running restore schema tests (cluster restart fails).

The tests in github actions will run without this trick, but I will add a way to enable it in `make start-dev-env`, so that in case someone wants to work with anything except restore schema, the cluster setup will take noticeably shorter time to finish.

Michal-Leszczynski:
In terms of restore schema problems described in this [issue](https://github.com/scylladb/scylladb/issues/16349), the exact approach that SM should take is still discussed, but for now it looks like restore schema isn't stable with Scylla 5.4.0 and `consistent_cluster_management: true`.

The only thing that this PR changes in this regard, is that when someone tries to restore schema with location that has annotated dc, SM will ignore this dc and try to use ALL nodes with location access for restoring schema. This change just improves stability but does not change any behavior observable by the user, so I think that we can merge it (it also fixes our tests, but they don't check all corner cases that might fail with Scylla 5.4.0 and  `consistent_cluster_management: true`).

haluz (approved):
LGTM

karol-kokoszka (approved):
:+1:

PR #3669, opened 2023-12-20
Run gh actions on enterprise Scylla

Michal-Leszczynski:
Just running to see test output, but enterprise releases should definitely be included in test matrix.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -30,7 +30,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        scylla-version: [ 5.2.11, 5.4.0 ]
+        scylla-version: [ 'scylla:5.2.13', 'scylla:5.4.1', 'scylla-enterprise:2023.1.3', 'scylla-enterprise:2024.1.0-rc3' ]
         ip-family: [ IPV4, IPV6 ]
         raft-enabled: [ true, false ]
     uses: ./.github/workflows/integration-tests.yaml
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -30,6 +30,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # Due to Scylla issue #16349 it is expected that restore-schema
+        # jobs with the following config will fail (4 jobs total):
+        # - scylla-version: scylla:5.2.X,               raft-enabled: true
+        # - scylla-version: scylla-enterprise:2024.1.X, raft-enabled: true
         scylla-version: [ 'scylla:5.2.13', 'scylla:5.4.1', 'scylla-enterprise:2023.1.3', 'scylla-enterprise:2024.1.0-rc3' ]
         ip-family: [ IPV4, IPV6 ]
         raft-enabled: [ true, false ]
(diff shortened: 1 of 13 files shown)


Michal-Leszczynski:
Results of running SM tests against enterprise 2024.1.0-rc1:
- SM repaired `system_replicated_keys` at the end of system tables (it wasn't directly specified in repair order because it's in enterprise only) - now it is repaired after `system_auth`.
- trying to start the first node in the cluster with IPV6 ends with `2023-12-21 12:21:22,270 INFO spawnerr: can't find command '/usr/bin/scylla-manager-agent'
`

tzach:
You should hit https://github.com/scylladb/scylladb/issues/16349 as well

Michal-Leszczynski:
@karol-kokoszka In [commit](https://github.com/scylladb/scylla-manager/pull/3669/commits/f168e61d6aed532d4adf3b3aa79dd62331f08815) I removed the workaround (that doesn't work in real life scenarios) for the tests to work with Scylla 5.4 with raft schema, so now it's visible in test failures. As expected, the only failed jobs are restore schema jobs with newer Scylla (5.4 / 2024.1) and raft schema enabled.

There are 2 questions regarding this PR:
1. Should we skip restore schema jobs with configuration that we know should fail, or should we leave them as they are and validate that these are the only failed checks before merging PR? I would leave them as they are so that we keep them in mind all the time and fix them when a proper solution is delivered on the core side.
2. This PR increases the number of jobs to 81! It seems like a lot to me, but perhaps it's fine and we shouldn't worry about that. Anyway, I think that it's not necessary to check each configuration (SCYLLA_VERSION/IP_FAMILY/RAFT_ENABLED) before merging a PR, so maybe we could remove some configurations from the matrix (e.g. don't test both IP_FAIMLY=IPV6 & RAFT_ENABLED=true and IP_FAMILY=IPV4 and RAFT_ENABLED=true)?

Apart from those questions, this PR is ready for review.

karol-kokoszka:
> There are 2 questions regarding this PR:

> Should we skip restore schema jobs with configuration that we know should fail, or should we leave them as they are and validate that these are the only failed checks before merging PR? I would leave them as they are so that we keep them in mind all the time and fix them when a proper solution is delivered on the core side.

Please keep the checks. It's a scylla-manager independent problem. Although, pls add the comment in code, that the test fails on the particular configuration.

> This PR increases the number of jobs to 81! It seems like a lot to me, but perhaps it's fine and we shouldn't worry about that. Anyway, I think that it's not necessary to check each configuration (SCYLLA_VERSION/IP_FAMILY/RAFT_ENABLED) before merging a PR, so maybe we could remove some configurations from the matrix (e.g. don't test both IP_FAIMLY=IPV6 & RAFT_ENABLED=true and IP_FAMILY=IPV4 and RAFT_ENABLED=true)?
Apart from those questions, this PR is ready for review.

As long as they are free of charge, we are fine.
According to this https://github.com/orgs/community/discussions/70492, they do.
We are using `ubuntu-lates`, which is standard free-of-charge for public repo runner.

karol-kokoszka (approved):
:+1: , just add the comment to restore schema tests about the environment where it doesn't pass and why

PR #3670, opened 2023-12-22
Repair: use small_table_optimization param

Michal-Leszczynski:
WIP: as small_table_optimization isn't part of 2024.1.0-rc2, we have to wait for some Scylla release that contains this new repair param for proper testing.


mykaul:
@Michal-Leszczynski - I thought it's in 2024.2 (6.0) ?

Michal-Leszczynski:
Yes but this PR was stale because we were unable to test it before and now it is conflicting with refactor needed for tablet repair. I will recreate this PR when tablet repair is done and tested on SM side.

Issue #3671, opened 2023-12-25
repair fails after 2.5 hours on enterprise performance tests


fruch:
## Issue description

- [ ] This issue is a regression.
- [x] It is unknown if this issue is a regression.

During performance test of repair operation on enterprise latest build `2024.2.0~dev-20231221.a2b53d3cbc61`
running with manager `3.2.5-0.20231206.8b378dea`, we are getting this error after 2.5h
```
Dec 24 10:36:47 perf-latency-nemesis-ubuntu-monitor-node-f00ac702-1 scylla-manager[31189]: {"L":"INFO","T":"2023-12-24T10:36:47.971Z","N":"scheduler.64e45ef8","M":"Run ended with ERROR","task":"repair/7a181846-a5af-4417-8767-40fca3718474","status":"ERROR","cause":"repair 1 token ranges out of 7680","duration":"2h30m18.219742422s","_trace_id":"A4jRVqRxSB2aLM4ZY3wL2g"}
```
no error or failure is spotted on the db nodes logs, 

we do see multiple of those errors in the log, but can't there related to the final error or not:
```
Dec 24 08:10:42 perf-latency-nemesis-ubuntu-monitor-node-f00ac702-1 scylla-manager[31189]: {"L":"INFO","T":"2023-12-24T08:10:42.526Z","N":"cluster.client","M":"HTTP retry backoff","operation":"StorageServiceRepairStatus","wait":"962.974517ms","error":"after 30s: context deadline exceeded","_trace_id":"A4jRVqRxSB2aLM4ZY3wL2g"}
```

## Impact

User can't understand why repair is failing 

## How frequently does it reproduce?

it's the first time we are seeing it on those tests

## Installation details

Kernel Version: 5.15.0-1051-aws
Scylla version (or git commit hash): `2024.2.0~dev-20231221.a2b53d3cbc61` with build-id `b8da1ee6e42f25d61816f807e3981ac2f9b25b71`

Cluster size: 3 nodes (i3en.2xlarge)

Scylla Nodes used in this run:
- perf-latency-nemesis-ubuntu-db-node-f00ac702-7 (3.223.195.60 | 10.12.0.99) (shards: 7)
- perf-latency-nemesis-ubuntu-db-node-f00ac702-6 (18.214.26.221 | 10.12.0.11) (shards: 7)
- perf-latency-nemesis-ubuntu-db-node-f00ac702-5 (44.204.249.160 | 10.12.0.60) (shards: 7)
[...]

PR #3673, opened 2023-12-28
Fix timeouts in RepairStatus and LoadSSTables

### LINKED ISSUE #3675: Remove timeouts from synchronous Scylla API calls
Some Scylla endpoints returns only when the operation has been finished. In case this operation is expected to take a long time, we should just hang on the connection and wait for response - not timeout and retry, because it is less efficient and creates confusing error logs. The 2 synchronous calls used with default timeout are:
- StorageServiceRepairStatus: ``` {"L":"INFO","T":"2023-12-01T01:31:54.398Z","N":"cluster.client","M":"HTTP retry backoff","operation":"StorageServiceRepairStatus","wait":"28.607063257s","error":"after 16m0s: context deadline exceeded","_trace_id":"uqSjtDSfRoOl1WhetiLPgA"} ```
- StorageServiceSstablesByKeyspacePost: ``` {"L":"INFO","T":"2023-11-30T23:03:07.117Z","N":"cluster.client","M":"HTTP retry backoff","operation":"StorageServiceSstablesByKeyspacePost","wait":"999.175032ms","error":"after 30s: context deadline exceeded","_trace_id":"uqSjtDSfRoOl1WhetiLPgA"} ```

Michal-Leszczynski:
Both of those calls return from Scylla only when the job is done. Because of that SM shouldn't timeout them on its side (even though it usually works when backoff retry is used).
This PR also includes a small improvement of repair logs mentioned in [here](https://github.com/scylladb/scylla-manager/issues/3671#issuecomment-1870061453).

Fixes #3675

### DIFF
--- a/pkg/scyllaclient/client_scylla.go
+++ b/pkg/scyllaclient/client_scylla.go
@@ -899,6 +899,8 @@ func (c *Client) TableDiskSizeReport(ctx context.Context, hostKeyspaceTables Hos
 	return report, err
 }
 
+const loadSSTablesTimeout = time.Hour
+
 // LoadSSTables that are already downloaded to host's table upload directory.
 // Used API endpoint has the following properties:
 // - It is synchronous - response is received only after the loading has finished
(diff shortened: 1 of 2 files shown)


pkg/scyllaclient/client_scylla.go
```diff
@@ -906,7 +907,7 @@ func (c *Client) LoadSSTables(ctx context.Context, host, keyspace, table string,
 	const WIPError = "Already loading SSTables"
 
 	_, err := c.scyllaOps.StorageServiceSstablesByKeyspacePost(&operations.StorageServiceSstablesByKeyspacePostParams{
-		Context:            forceHost(ctx, host),
+		Context:            noTimeout(forceHost(ctx, host)),
```
karol-kokoszka, line 910: Please don't use noTimeout.
Use customTimeout instead and set 30m to not hang forever.

Michal-Leszczynski:
@karol-kokoszka I added some big timeouts, PTAL.

karol-kokoszka (approved):
:+1:

PR #3674, opened 2023-12-28
Remove hint about recreating MV and SI from restore schema

### LINKED ISSUE #3676: Restore schema docs suggest to recreate MV manually
Mentioned fragment of SM [docs](https://manager.docs.scylladb.com/stable/restore/restore-schema.html#follow-up-action). This is outdated as SM does that automatically in restore tables step.

Michal-Leszczynski:
There is no need to do that as it is performed automatically.
Actually, this has been a part of SM for quite some time already.

In response to [comment](https://github.com/scylladb/scylla-enterprise/issues/3706#issuecomment-1864611316).
Fixes #3676


karol-kokoszka (approved):
:+1:

PR #3686, opened 2024-01-09
Fix restore schema for Scylla 5.4 with raft

Michal-Leszczynski:
This PR implements a workaround for restoring schema in multi-location cluster with Scylla 5.4 with schema changes based on raft.


Michal-Leszczynski:
I'm closing this PR because the chosen workaround is to restore schema into cluster without raft schema enabled, rolling-restart cluster and only after that enable raft schema if needed.

PR #3695, opened 2024-01-23
Update docker setup example

### LINKED ISSUE #3672: Unable to configure Scylladb-manager and agent [using containers]
We are trying to configure the pilot scylladb manager ... These are the configuration we are trying to get it
Scylladb with scylladb - agent - IP1
Scylladb manager - IP2
Using the Dockerized env as mentioned here - https://manager.docs.scylladb.com/stable/docker/index.html
When I am trying to add the cluster of IP1 using sctool at Scylla-manager [bash] running in IP2 ... Getting the following error as shared in the screenshot ....
I have already whitelisted and exposed these ports at agent side ...
```
     - "9042:9042"  # CQL native transport port
      - "19042:19042"  # Internode communication (non-SSL)
      - "7199:7199"  # JMX management port
      - "9180:9180"  # Prometheus metrics port
      - "10001:10001" # Scylla Manager agent port
      - "9100:9100"
```
Also generated and added TLS cert.
Also tried connecting the agent with GET req and seems it is able to respond but yet failed to be added as cluster

image
image

Let me know if you need any further info for this ... 

Thanks.

Michal-Leszczynski:
Previously I wasn't able to follow our docker setup [example](https://manager.docs.scylladb.com/stable/docker/index.html) because of two reasons:
- too little memory assigned to SM DB
- incorrect `echo` behavior when writing config files

Now it should work fine.
Fixes #3672

### DIFF
--- a/docs/source/docker/Dockerfile
+++ b/docs/source/docker/Dockerfile
@@ -2,14 +2,14 @@ FROM scylladb/scylla-manager-agent:latest as agent
 FROM scylladb/scylla:latest
 
 COPY --from=agent /usr/bin/scylla-manager-agent /usr/bin/
-RUN echo -e "[program:scylla-manager-agent]\n\
+RUN echo "[program:scylla-manager-agent]\n\
 command=/usr/bin/scylla-manager-agent\n\
 autorestart=true\n\
 stdout_logfile=/dev/stdout\n\
 stdout_logfile_maxbytes=0\n\
 stderr_logfile=/dev/stderr\n\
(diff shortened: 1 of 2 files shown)


karol-kokoszka (approved):
nice catch !

It may be that scylla-university uses same examples to setup manager in docket env.

:+1: from my side

Michal-Leszczynski:
From what I see in [example](https://github.com/scylladb/scylla-code-samples/blob/master/manager/Dockerfile) used in Scylla University, it actually doesn't use `echo -e` and it uses older Scylla version, that should work fine with less memory, so I don't think that there is a need to update that.

Issue #3662, opened 2023-12-14
Fix restore schema procedure


Michal-Leszczynski:
Restoring schema doesn't work with (or at least isn't stable) with Scylla 5.4.0 with `consistent_cluster_management: true`. This resulted in creating this [issue](https://github.com/scylladb/scylladb/issues/16349) which exposed that SM is restoring schema in a dangerous way. Restore schema procedure should be changed to:

PREREQUISITE:
 - backup
 - fresh cluster (can have different topology than the backup)
 - ALL nodes have access to ALL backup locations. This means that in multi-location backup setting, nodes need to have access to more than one location

PROCEDURE:
- add cluster to SM (user)
- stop the whole cluster - all nodes are down, but agents are up (user)
- download schema SSTables from ALL locations to ALL nodes DATA dir - not upload like before, so we can skip `nodetool refresh` (SM)
- rolling-start the cluster (user)
- ALTER cluster schema to match the new topology (e.g. if new cluster contains additional dc, add it to keyspace replication strategy)

The only potential problem is to extend agent configuration, so that it is possible for agent to have access to more than 1 location from given provider and make sure that it works well with rclone.

kostja:
https://manager.docs.scylladb.com/stable/sctool/backup.html has an option to restore a single table or a single keyspace.
How is the procedure above providing the implementation of restore for this option?

Michal-Leszczynski:
Option of restoring a single table/keyspace is only available for restoring user data. Schema restoration always targets the whole backed-up schema (because we just re-upload all system_schema sstables).

tzach:
@Michal-Leszczynski, to summarize, we have two tracks for fixing this issue in Scylla Manager
- Short term: for 5.4 and 2024.1, disable  Raft on cluster init, restore it, and then enable it.
- Long-term: using a new DESCRIBE schema

Both should be automated, tested, and documented.

Michal-Leszczynski:
> Short term: for 5.4 and 2024.1, disable Raft on cluster init, restore it, and then enable it.

Restoring schema from cluster with raft to the cluster without raft is now a part of SM gh actions  tests, but enabling raft afterwards isn't, so it should be added. 

> Long-term: using a new DESCRIBE schema

[...]

PR #3697, opened 2024-01-24
Restore schema workaround

Michal-Leszczynski:
This PR adds a validation of `consistent_cluster_management` and Scylla version before restoring schema (more info in  #3662). It also documents a workaround for newer Scylla versions with `consistent_cluster_management: true`.


Michal-Leszczynski:
This PR contains workaround documentation for bot Scylla 5.4 and 2024, even though the latter one is not released yet. Is this ok? It also adds Scylla 5.4 into restore compatibility matrix. Should Scylla 2024 be added there as well?

pkg/scyllaclient/config_client.go
```diff
@@ -209,6 +209,18 @@ func (c *ConfigClient) UUIDSStableIdentifiers(ctx context.Context) (bool, error)
 	return resp.Payload, err
 }
 
+// ConsistentClusterManagement returns true if node uses RAFT for cluster management and DDL.
+func (c *ConfigClient) ConsistentClusterManagement(ctx context.Context) (bool, error) {
+	resp, err := c.client.Config.FindConfigConsistentClusterManagement(config.NewFindConfigConsistentClusterManagementParamsWithContext(ctx))
+	if isStatusCode400(err) {
+		return false, nil
+	}
```
karol-kokoszka, line 217: What is the meaning of HTTP 400 here ? Current scylla doesn't support this config param ?

I've found this swagger doc in enterprise repo https://github.com/scylladb/scylla-enterprise/blob/enterprise/api/api-doc/config.json

There is nothing about 400 code.

karol-kokoszka (approved):
:+1: LGTM

pkg/scyllaclient/config_client.go
Michal-Leszczynski: > What is the meaning of HTTP 400 here ? Current scylla doesn't support this config param ?

I think that it is used for backward compatibility. Without that, checking Scylla config `consistent_cluster_management` returns errors in Scylla versions that don't have this field introduced. Since SM 3.2 still supports Scylla 5.1 (restore compatibility matrix), there is still a need for this small workaround.

Michal-Leszczynski:
@tzach @annastuchlik could you please take a look?

docs/source/restore/compatibility-matrix.rst
```diff
@@ -12,7 +13,7 @@ The following table shows which version of Scylla Manager restore task supports
      - ScyllaDB Open Source Version
      - ScyllaDB Enterprise Version
    * - 3.2
-     - 5.0, 5.1, 5.2
+     - 5.0, 5.1, 5.2, 5.4
      - 2022.1, 2022.2, 2023.1
```
tzach, line 17: 2024.1?

docs/source/restore/restore-schema.rst
```diff
@@ -47,4 +51,17 @@ Fortunately, the small size of schema files makes this overhead negligible.
...
-        * `nodetool refresh <https://docs.scylladb.com/stable/operating-scylla/nodetool-commands/refresh.html#nodetool-refresh>`_ on all downloaded schema tables (full parallel)
\ No newline at end of file
+        * `nodetool refresh <https://docs.scylladb.com/stable/operating-scylla/nodetool-commands/refresh.html#nodetool-refresh>`_ on all downloaded schema tables (full parallel)
+
+.. _restore-schema-workaround:
+
+Restoring schema into a cluster with ScyllaDB **5.4.X** or **2024.X** with **consistent_cluster_management**
+============================================================================================================
+
+Currently, restoring schema when using ScyllaDB **5.4.X** or **2024.X** with ``consistent_cluster_management: true`` in ``scylla.yaml``
```
tzach, line 61:
```suggestion
Restoring schema when using ScyllaDB **5.4.X** or **2024.X** with ``consistent_cluster_management: true`` in ``scylla.yaml``
```

"Currently" does not mean much in docs ...

docs/source/restore/_common/restore-raft-schema-warn.rst
```diff
@@ -0,0 +1 @@
+.. warning:: Restoring schema into a cluster with ScyllaDB **5.4.X** or **2024.X** with ``consistent_cluster_management: true`` isn't supported. Please see the following :ref:`workaround <restore-schema-workaround>`.
```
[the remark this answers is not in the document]
Michal-Leszczynski: @tzach @karol-kokoszka I changed restore schema safety validation so that:
- it always works with Scylla version < 5.4 / 2024.1
- i works when `consistent_cluster_management: false` and Scylla version 5.4 / 2024.1
- it newer works with Scylla version >= 6.0 / 2024.2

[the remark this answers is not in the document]
karol-kokoszka: > I think it should be 2024.1.x. I believe consistent_cluster_management: true will not be configurable for versions later than 5.4 and 2024.1 (it was already removed from the docs and scylla.yaml).

@annastuchlik it means that there will be no option to disable raft in new versions of Scylla, right ?

karol-kokoszka (approved):

Issue #3698, opened 2024-01-25
Manager uses TLS port even when it cannot establish a handshake


zimnx:
Manager prefers to establish CQL sessions when SSL is open, even when necessary client certificate is not registered.

Clusters exposing both CQL and CQL SSL with mTLS are failing backups and CQL healthchecks.

rzetelskik:
Related to https://github.com/scylladb/scylla-manager/issues/3679

Issue #3679, opened 2024-01-03
Manager should not assume CQL-SSL in cluster sessions


rzetelskik:
### What happened?

Currently, if you set up CQL-SSL in your ScyllaDB cluster, the manager sets up a TLS client when establishing a CQL session with a cluster, even if it wasn't provided with certs.

### What did you expect to happen?

Scylla Manager should not use TLS by default. Scylla Manager should configure a TLS client when establishing a CQL session with a cluster **only when specifically configured to do so**.

### How can we reproduce it?

1. Set up a Scylla Manager instance
2. Deploy a ScyllaDB cluster with the following config (omitting irrelevant fields)
```
...
client_encryption_options:
  certificate: /var/run/secrets/scylla-operator.scylladb.com/scylladb/serving-certs/tls.crt
  enabled: true
  keyfile: /var/run/secrets/scylla-operator.scylladb.com/scylladb/serving-certs/tls.key
  optional: false
  require_client_auth: true
  truststore: /var/run/secrets/scylla-operator.scylladb.com/scylladb/client-ca/tls.crt
native_transport_port: 9042
native_transport_port_ssl: 9142
...
```
3. Register the cluster with Scylla Manager **without providing the certs**.
4. Verify that Scylla Manager throws errors on a CQL session get attempt:
```
$ sctool restore -c scylla/scylla -L gcs:rzetelskik
Error: get resource: create restore target, units and views: create worker: get CQL cluster session: TLS/SSL key/cert is not registered: not found
Trace ID: 4nGqrQEUQU2e93kHWadCyw (grep in scylla-manager logs)
```

```
{"L":"INFO","T":"2024-01-03T13:39:45.929Z","N":"cluster","M":"Creating new Scylla REST client","cluster_id":"ae1376b9-2524-4e1d-a8f9-4a4681fc7d32","_trace_id":"4nGqrQEUQU2e93kHWadCyw"}
{"L":"INFO","T":"2024-01-03T13:39:45.933Z","N":"cluster.client","M":"Measuring datacenter latencies","dcs":["us-east-1"],"_trace_id":"4nGqrQEUQU2e93kHWadCyw"}
[...]

PR #3699, opened 2024-01-25
Require certificate registered within cluster before choosing CQL SSL

zimnx:
Previously, SSL was preferred when client_encryption_options.enabled
coming from ScyllaDB configuration was true and SSL port is open,
even when Scylla Manager did not have any client certificate registered
for particular cluster.

This caused issues when ScyllaDB cluster was exposing both CQL and CQL
SSL with mTLS, because even when Manager was not registered with
certificates, it still insisted to establish sessions using SSL port.
CQL healthchecks was also affected.

It can be further improved by implementing #3679 which will add a switch deciding whether user wants to use either SSL or non SSL port.

Fixes https://github.com/scylladb/scylla-manager/issues/3698

### DIFF
--- a/pkg/service/cluster/service.go
+++ b/pkg/service/cluster/service.go
@@ -8,6 +8,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"sort"
+	"strconv"
 
 	"github.com/gocql/gocql"
 	"github.com/pkg/errors"
@@ -549,6 +550,7 @@ func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID) (session 
 	}
 
 	scyllaCluster := gocql.NewCluster(sessionHosts...)
+	cqlPort := ni.CQLPort()
 
 	if ni.CqlPasswordProtected {
 		credentials := secrets.CQLCreds{
@@ -568,20 +570,41 @@ func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID) (session 
 		}
 	}
 
-	if ni.ClientEncryptionEnabled {
+	keyPair, err := s.loadTLSIdentity(clusterID)
+	if err != nil && !errors.Is(err, service.ErrNotFound) {
+		return session, err
+	}
+
+	if ni.ClientEncryptionEnabled && !ni.ClientEncryptionRequireAuth {
+		cqlPort = ni.CQLSSLPort()
 		scyllaCluster.SslOpts = &gocql.SslOptions{
 			Config: &tls.Config{
 				InsecureSkipVerify: true,
 			},
 		}
-		if ni.ClientEncryptionRequireAuth {
-			keyPair, err := s.loadTLSIdentity(clusterID)
-			if err != nil {
-				return session, err
-			}
-			scyllaCluster.SslOpts.Config.Certificates = []tls.Certificate{keyPair}
-		}
 	}
+
+	if ni.ClientEncryptionEnabled && ni.ClientEncryptionRequireAuth && !errors.Is(err, service.ErrNotFound) {
+		cqlPort = ni.CQLSSLPort()
+		scyllaCluster.SslOpts = &gocql.SslOptions{
+			Config: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		}
+		scyllaCluster.SslOpts.Config.Certificates = []tls.Certificate{keyPair}
+	}
+
+	if ni.ClientEncryptionEnabled && ni.ClientEncryptionRequireAuth && errors.Is(err, service.ErrNotFound) {
+		s.logger.Info(ctx, "Client encryption is enabled, but Cluster wasn't registered with certificate in Scylla Manager, falling back to nonSSL port.",
+			"cluster_id", clusterID,
+		)
+	}
+
+	p, err := strconv.Atoi(cqlPort)
+	if err != nil {
+		return session, errors.Wrap(err, "parse cql port")
+	}
+	scyllaCluster.Port = p
 
 	return gocqlx.WrapSession(scyllaCluster.CreateSession())
 }
--- a/pkg/service/healthcheck/service.go
+++ b/pkg/service/healthcheck/service.go
@@ -354,6 +354,7 @@ func (s *Service) pingCQL(ctx context.Context, clusterID uuid.UUID, host string,
 
 	tlsConfig := ni.tlsConfig(cqlPing)
 	if tlsConfig != nil {
+		config.Addr = ni.CQLSSLAddr(host)
 		config.TLSConfig = tlsConfig.Clone()
 	}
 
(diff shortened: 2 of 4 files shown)


zimnx:
Seems like failures of CI tests, are also present on master. Is this a known issue?

karol-kokoszka:
@zimnx restore schema in 5.4 is a know issue.
Ill review the PR tomorrow.

pkg/service/healthcheck/service.go
```diff
@@ -418,10 +419,12 @@ func (s *Service) nodeInfo(ctx context.Context, clusterID uuid.UUID, host string
...
 			tlsConfig, err := s.tlsConfig(clusterID, clientCertAuth)
-			if err != nil {
+			if err != nil && !errors.Is(err, service.ErrNotFound) {
 				return ni, errors.Wrap(err, "fetch TLS config")
 			}
-			ni.TLSConfig[p] = tlsConfig
+			if !errors.Is(err, service.ErrNotFound) {
+				ni.TLSConfig[p] = tlsConfig
+			}
 		}
```
karol-kokoszka, line 428: The logic becomes bit fuzzy now.
It checks `nodeInfo` response if the TLS is enabled, 
if it is and if it supposed to be protected with the certificate, try to get the certificate
if the certificate is not there (we don't store it), then finally fallback to non TLS connection.

Previously service.ErrNotFound was wrapped with 
```
			return nil, errors.Wrap(err, "get SSL user cert from secrets store")
```
, now this information will be ignored.

I can imagine the situation where the TLS is enabled, there is no way to connect without TLS, but someone forgot to add the cert. SM will create the session, but the end-user will have some information like "cannot connect to the cluster" or whatever. No information about missing SSL cert.

Please don't ignore ErrNotFound, log it with Info() level adding that it will fallback to TLS disabled connection.

karol-kokoszka: It may be that just logging is not a good idea.
Healtcheck is executed often, so it means that there is chance to SPAM logs.

Is there a chance to check if nonSSL is enabled ?

pkg/service/cluster/service.go
```diff
@@ -568,20 +570,28 @@ func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID) (session
 		}
 	}
 
-	if ni.ClientEncryptionEnabled {
+	keyPair, err := s.loadTLSIdentity(clusterID)
+	if err != nil && !errors.Is(err, service.ErrNotFound) {
+		return session, err
+	}
+	if ni.ClientEncryptionEnabled && !errors.Is(err, service.ErrNotFound) {
```
Michal-Leszczynski, line 577: I think it would be clearer if you used:
```go
if ni.ClientEncryptionEnabled && err == nil {
```

pkg/service/healthcheck/service.go
zimnx: As discussed offline, it's not possible to determine whether given port is open from Scylla API (https://github.com/scylladb/scylladb/issues/7206#issuecomment-689598343).
Added logging for case when TLS is enabled in config, but Manager is missing certificate to make it better visible to the user.

pkg/service/cluster/service.go
zimnx: negation of NotFound in my view carries clearer intention of something being found, than err equal to nil.

Michal-Leszczynski:
I'm not an expert on this topic, but the following fragment of code:
   ```go
	scyllaCluster.SslOpts = &gocql.SslOptions{
		Config: &tls.Config{
			InsecureSkipVerify: true,
		},
	}
   ```
Makes it so that SM never validates Scylla certs. In general, is this desired or should be changed in the future? (I'm aware that it was always there)

karol-kokoszka:
@Michal-Leszczynski that's a good point. But, on the other hand, manager desire is not to save some sensitive information to Scylla cluster. It just need connection to check if the node lives. We can ignore the certs.

karol-kokoszka (approved):
:+1:

zimnx:
> It just need connection to check if the node lives.

Certificate validation is also disabled for CQL session, which is serious as Manager might connect to different cluster when there's a MITM.

Michal-Leszczynski:
What about a situation when:
- SM don't have TLS certs
- ClientEncryptionEnabled = true
- ClientEncryptionRequireAuth = false

This implementation suggests to use regular CQL port, but is there anything stopping SM from using CQL SSL port without specifying certs? Wouldn't it be better/safer?

Michal-Leszczynski:
This PR also removes the following behavior:
```
```

Michal-Leszczynski:
This PR also removes the following part of CQL SSL port fallback:
```go
// Scylla API always returns non-empty NativeTransportPortSSL even when
// value is explicitly disabled in configuration file.
// This makes impossible to determine which port is being used for CQL
// frontend. To workaround it, we try to dial SSL port when
// client encryption is enabled. If any error happens, assume this port
// is not used.
// Ref: https://github.com/scylladb/scylla/issues/7206
```
Which goes in pair with the following fragment of `scylla.yaml`:
```yaml
# Enabling native transport encryption in client_encryption_options allows you to either use
# encryption for the standard port or to use a dedicated, additional port along with the unencrypted
# standard native_transport_port.
# Enabling client encryption and keeping native_transport_port_ssl disabled will use encryption
# for native_transport_port. Setting native_transport_port_ssl to a different value
# from native_transport_port will use encryption for native_transport_port_ssl while
# keeping native_transport_port unencrypted.
#native_transport_port_ssl: 9142
```
My understanding is that in some configurations `NativeTransportPortSsl` might be disabled even when `ClientEncryptionEnabled` is enabled (possibly with `ClientEncryptionRequireAuth` enabled as well).
Mentioned [issue](https://github.com/scylladb/scylladb/issues/7206) is still open, so is it safe to remove this fallback behavior?

zimnx:
> What about a situation when:
> 
>     * SM doesn't have TLS certs
> 
>     * ClientEncryptionEnabled = true
> 
>     * ClientEncryptionRequireAuth = false
> 
> 
> This implementation suggests to use regular CQL port, but is there anything stopping SM from using CQL SSL port without specifying certs? Wouldn't it be better/safer?

Yep, that would be better. Added support for it.

zimnx:
> This PR also removes the following part of CQL SSL port fallback:

This fallback is error prone and gives false negatives. Implemented probe had just 1s timeout and was sent from Manager Server which for case when given node is in different DC may return different results between runs.

Because referenced issue is not solved by Scylla, there's no reliable way to tell whether given port is open or not.
Instead on relying on unreliable probes, Manager will add switch where user will decide whether his cluster should use TLS or not. This will be solved as part of #3679.

karol-kokoszka:
@Michal-Leszczynski there is unresolved converation under your comment. Please let us know if it's ok, so that @zimnx can merge the PR.

Michal-Leszczynski (approved):

PR #3711, opened 2024-02-08
Kk/3679 cluster param indicating tls enabled

### LINKED ISSUE #3679: Manager should not assume CQL-SSL in cluster sessions

karol-kokoszka:
Closes #3679 

---

Please make sure that:


pkg/service/cluster/model.go
```diff
@@ -19,6 +19,16 @@ type Cluster struct {
...
+	// Three values are needed here where tls_enabled = Null means that it's unknown
+	// if the cluster has this flag enabled or not. It's managers responsibility to
+	// find the correct value on the first usage, basing on the already existing logic where scylla configuration
+	// is checked against `/config/client_encryption_options` API call to scylla server.
+	// Null value is introduced to seamlessly support migration period.
+	// Beginning Scylla Manager 3.2.6, it's user responsibility to explicitly set on cluster
+	// if the healtchecks should be performed on encrypted or unencrypted connection.
+	//
+	// https://github.com/scylladb/scylla-manager/issues/3679
+	TLSEnabled *bool `json:"tls_enabled"`
```
Michal-Leszczynski, line 31: I don't fully understand the need for 3 values here - why can't we just have a single flag `--disable-tls` (default false)?

Also, I don't think that this approach solves problem with choosing correct port described [here](https://github.com/scylladb/scylla-manager/pull/3699#issuecomment-1912025255) (SM can't tell whether TLS is listened on regular or ssl port).

There are basically 3 things that SM needs to know:
- which port should it use?
- should it use encryption?
- should it provide certs?

It has been shown that there is no way for SM to guess the correct port without dialing it, so this information should be passed by user (e.g. boolean `--disable-ssl-port`, default false, or maybe `--port` which allows for overwriting used port with either a number or a "native" / "native-ssl" strings.)

Even when we know the port (regular or ssl), SM doesn't always know whether this port is encrypted. So it indeed needs a flag for that (e.g. `--disable-encryption`, default false)

In terms of the third question, we are lucky that at least `ClientEncryptionRequireAuth` seems like a reliable source of truth, so SM can just take it from Scylla config.

karol-kokoszka: 3 values are added to smoothly go from the previous approach where the manager were defining if the TLS should be used or not just during the session creation. 
Other approach would be to just migrate all the clusters to update the infromation about disabled/enabled TLS in the DB on manager upgrade. Wanted to avoid it, go migration code is executed outside of the context of manager. There is no handlers to Scylla client available for example. 
https://github.com/scylladb/scylla-manager/tree/master/pkg/schema/migrate

So, if the TLSEnabled = null, then it will failback to this simple check. So the TLSEnabled will be equal scylla configuration (ClientEncryptionEnabled)
https://github.com/scylladb/scylla-manager/blob/35addef69c0eb2b2012ef4367548e05f4daac5b1/pkg/service/cluster/service.go#L578-L589

> which port should it use?

if tls_enabled -> https://github.com/scylladb/scylla-manager/blob/35addef69c0eb2b2012ef4367548e05f4daac5b1/pkg/service/cluster/service.go#L592
otherwise https://github.com/scylladb/scylla-manager/blob/35addef69c0eb2b2012ef4367548e05f4daac5b1/pkg/service/cluster/service.go#L553

> should it use encryption?

It's exactly the "tls_enabled" property of the cluster (which can be nil)

> should it provide certs?

https://github.com/scylladb/scylla-manager/blob/35addef69c0eb2b2012ef4367548e05f4daac5b1/pkg/service/cluster/service.go#L598

> It has been shown that there is no way for SM to guess the correct port without dialing it, so this information should be passed by user (e.g. boolean --disable-ssl-port, default false, or maybe --port which allows for overwriting used port with either a number or a "native" / "native-ssl" strings.)

This is part of `node_info` call response. The call is made to scylla API through agent proxy. 
We don't need to store it in DB. I don't want to store it in DB, lets the manager auto-configure itself whenever there is a change in scylla config. By autoconfigure, I mean just dynamically read it everytime when it's needed.

> Even when we know the port (regular or ssl), SM doesn't always know whether this port is encrypted. So it indeed needs a flag for that (e.g. --disable-encryption, default false)

It's part of node_info response.

There is no intention to completly remove node_info call to agent here.

karol-kokoszka: To sum up:
- tls_enabled=true explicitly set to false mean **Enable TLS no matter what scylla says about its configuration**
- tls_enabled=false explicitly set to false mean **Disable TLS no matter what scylla says about its configuration**
- tls_enabled=NIL explicitly set to false mean **I have no idea, please check scylla-configuration**

karol-kokoszka: Removed `--tls-enabled` and replaced with two flags `--force-tls-disabled` and `--force-non-ssl-port`

pkg/restapi/cluster.go
```diff
@@ -108,6 +112,42 @@ func (h clusterHandler) parseCluster(r *http.Request) (*cluster.Cluster, error)
...
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to get cluster {%v}", apiC.ID)
+		}
+		if apiC.ForceTLSDisabled == nil {
+			c.ForceTLSDisabled = oldCluster.ForceTLSDisabled
+		}
+		if apiC.ForceNonSslPort == nil {
+			c.ForceNonSSLPort = oldCluster.ForceNonSSLPort
+		}
+	}
```
Michal-Leszczynski, line 146: Could you explain why those flags are handled in a special way?
My understanding is that when we create/update cluster, we require client to send a full new cluster definition to SM, not just the changed fields. We handle it on the `sctool` side by first getting the cluster definition from SM and then applying changes to it:
```go
func (cmd *command) run() error {
	cluster, err := cmd.client.GetCluster(cmd.Context(), cmd.cluster)
	if err != nil {
		return err
	}

	ok := false
	if cmd.Flags().Changed("name") {
		cluster.Name = cmd.name
		ok = true
	}
	if cmd.Flags().Changed("host") {
		cluster.Host = cmd.host
		ok = true
	}
	if cmd.Flags().Changed("port") {
		cluster.Port = cmd.port
		ok = true
	}
```
This way we don't worry that on update not specified value could overwrite previously set value.
If that's the case, I think that those flags should follow the same convention for simplicity.

pkg/command/cluster/clusteradd/cmd.go
```diff
@@ -169,3 +184,7 @@ func clusterAddedMessage(w io.Writer, id, name string) error {
 const clusterAddNoAuthTokenWarning = `
 WARNING! Scylla data may be exposed
 Protect it by specifying auth_token in /etc/scylla-manager-agent/scylla-manager-agent.yaml on Scylla nodes`
+
+const clusterForceTLSDisabledWarning = `WARNING! TLS is explicitly disabled on cluster sessions (--force-tls-disabled=false).`
```
Michal-Leszczynski, line 188: Nice warnings! But I think that you meant to write `(--force-tls-disabled=true)`.

pkg/command/cluster/clusterupdate/cmd.go
```diff
@@ -95,6 +99,14 @@ func (cmd *command) run() error {
...
 		ok = true
 	}
+	if cmd.forceTLSDisabled.Changed {
+		cluster.ForceTLSDisabled = &cmd.forceTLSDisabled.Val
+		ok = true
+	}
+	if cmd.forceNonSSLPort.Changed {
+		cluster.ForceNonSslPort = &cmd.forceNonSSLPort.Val
+		ok = true
+	}
```
Michal-Leszczynski, line 109: This is the part when we overwrite those flags values only when they were specified.

pkg/service/cluster/service.go
```diff
@@ -385,6 +385,13 @@ func (s *Service) PutCluster(ctx context.Context, c *Cluster) (err error) {
 		)
 	}
 
+	//// Create the session and log error
```
Michal-Leszczynski, line 388: Why so many dashes????

pkg/service/cluster/service.go
```diff
@@ -385,6 +385,13 @@ func (s *Service) PutCluster(ctx context.Context, c *Cluster) (err error) {
...
+	_, err = s.GetSession(ctx, c.ID)
+	if err != nil {
+		s.logger.Info(ctx, "WARNING! Cannot create CQL session to the cluster. It will affect backup/restore/healthcheck services.",
+			"cluster_id", c.ID)
```
Michal-Leszczynski, line 392: It's a shame there is no `Warning` log level... 
Do you think that `Error` log level would fit better here? I don't have strong preferences, but it would make this more visible for sure.

Michal-Leszczynski (commented):
With those changes the cql dialing logic finally looks good! Also really nice idea with testing sctool commands.

I just left some comments about simplifying (if it is possible) the way in which the new flags are handled.

Michal-Leszczynski:
Also, the `res.yaml` files containing description of sctool flags should be updated (they are used in our docs as well).

Michal-Leszczynski:
Also, there are more failures in gh actions then expected, but they look like errors not connected to this PR.

swagger/scylla-manager.json
```diff
@@ -54,6 +54,14 @@
         },
         "without_repair": {
           "type": "boolean"
+        },
+        "force_tls_disabled": {
+          "type": "boolean",
+          "x-nullable": true
```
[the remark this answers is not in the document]
karol-kokoszka: It cannot.

We create sctool CLI that allows to update the cluster.
If particular cluster has it's own state in the DB already and `force-tls-disabled` is not nullable, then I have no possibility of NOT CHANGING force-tls-disabled flag on update.
CLI is calling manager HTTP API and creates json payload out of the request. Without nullable, I always have to define if force-tls-disabled is false or true.

swagger/scylla-manager.json
```diff
@@ -54,6 +54,14 @@
...
+        },
+        "force_non_ssl_port": {
```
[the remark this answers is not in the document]
karol-kokoszka: `port` is already confusing. Should be rather called "api_port". But let's not change it as it would basically break the contract scylla manager made by introducing this field.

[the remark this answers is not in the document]
Michal-Leszczynski: That's true, but since we can't change that, I think that we could consider renaming or describing the new flag.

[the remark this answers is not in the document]
karol-kokoszka: OK, maybe it's better to name it "force-non-ssl-session-port" ?

pkg/restapi/cluster.go
karol-kokoszka: Actually you are right. CLI update gets the cluster first and then applies all what is changed.
There is no need to even keep fields nullable in swagger.

swagger/scylla-manager.json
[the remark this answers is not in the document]
karol-kokoszka: done

swagger/scylla-manager.json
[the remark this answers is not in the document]
karol-kokoszka: changed to "force_non_ssl_session_port"

pkg/restapi/cluster.go
karol-kokoszka: done

pkg/service/cluster/service.go
karol-kokoszka: done

pkg/service/cluster/service.go
karol-kokoszka: Info log is fine.

pkg/command/cluster/clusterupdate/cmd.go
karol-kokoszka: right, changed

karol-kokoszka:
docs are updated with 274388b7f8dcce80f581943d5dc4c5e838e238b4

Michal-Leszczynski (approved):
One last thing would be to adjust healthcheck's pinging to follow the logic introduced here, but I guess it can be done in a separate PR.

PR #3718, opened 2024-02-16
Repair: describe ring per table and without cache

### LINKED ISSUE #3745: Incorrect replication strategy calculation in DescribeRing
The way in which `DescribeRing` calculates keyspace replication strategy isn't always correct.
```go
	// Detect replication strategy
	if len(ring.HostDC) == 1 {
		ring.Replication = LocalStrategy
	} else {
		ring.Replication = NetworkTopologyStrategy
		for _, tokens := range dcTokens {
			if tokens != len(resp.Payload) {
				ring.Replication = SimpleStrategy
				break
			}
		}
	}
```

E.g. this test fails when `SimpleStrategy` RF is greater than half of the nodes count: 

```go
	t.Run("keyspace replication strategy", func(t *testing.T) {
		h := newRepairTestHelper(t, session, defaultConfig())
		ctx, cancel := context.WithCancel(context.Background())
		defer cancel()

		if err := clusterSession.ExecStmt("CREATE KEYSPACE test_rf WITH replication = {'class': 'SimpleStrategy', 'replication_factor' : 4}"); err != nil {
			t.Fatal(err)
		}
		defer dropKeyspace(t, clusterSession, "test_rf")

		ring, err := h.Client.DescribeRing(ctx, "test_rf")
		if err != nil {
			t.Fatal(err)
		}

		t.Log("Replication strategy", ring.Replication)
		if ring.Replication != scyllaclient.SimpleStrategy {
			t.Fatal("Incorrect replication strategy")
		}
	})
```

Michal-Leszczynski:
Right now, repair plan contains descriptions of all repaired rings which are stored at the keyspace level. This has two main  issues:
- unnecessary memory consumption
- it makes it difficult to move to a per table ring description required for tablets

The main goal of this PR is remove ring descriptions from plan and query them when needed during an actual repair. 
In order to achieve that, generator responsibility is passed to generator and tableGenerator. Generator is responsible for querying ring description and creating tableGenerators, which take care of repairing given table.

This introduces a lot of code refactoring, so in order to make repair logic more comprehensible I added:
- MaxRingParallel
- ShouldRepairRing

Currently, determining max parallel or if a keyspace should be repaired is convoluted as it requires traversing all token ranges. The changes to DescribeRing replication strategy calculation make it possible to do it by looking at the replication strategy and replication factor, which is easier to understand, explain and test.

This PR tried not to touch repair progress manager, but it might be good to refactor it as well.

Fixes #3745

### DIFF
--- a/pkg/service/repair/plan.go
+++ b/pkg/service/repair/plan.go
@@ -320,6 +320,39 @@ func maxRepairRangesInParallel(shards uint, totalMemory int64) Intensity {
 	return NewIntensity(max)
 }
 
+// MaxRingParallel calculates max amount of repair jobs on ring limited to dcs.
+func MaxRingParallel(ring scyllaclient.Ring, dcs []string) int {
+	repairedDCs := strset.New(dcs...)
+	dcNodeCnt := make(map[string]int)
+	for _, dc := range ring.HostDC {
+		dcNodeCnt[dc]++
+	}
+
+	switch ring.Replication {
+	case scyllaclient.SimpleStrategy:
+		repaired := 0
+		for dc, cnt := range dcNodeCnt {
+			if repairedDCs.Has(dc) {
+				repaired += cnt
+			}
+		}
+		return repaired / ring.RF
+	case scyllaclient.NetworkTopologyStrategy:
+		minDC := math.MaxInt / 2
+		for dc, rf := range ring.DCrf {
+			if repairedDCs.Has(dc) {
+				minDC = min(minDC, dcNodeCnt[dc]/rf)
+			}
+		}
+		if minDC == math.MaxInt/2 {
+			minDC = 1
+		}
+		return minDC
+	default:
+		return 1
+	}
+}
+
 // keyspacePlan describes repair schedule and state for keyspace.
 type keyspacePlan struct {
 	Keyspace string
--- a/pkg/service/repair/plan.go
+++ b/pkg/service/repair/plan.go
@@ -302,6 +302,39 @@ func (p *plan) KeyspaceRangesMap() map[string]int64 {
 	return out
 }
 
+// ShouldRepairRing when all ranges are replicated (len(replicaSet) > 1) in specified dcs.
+// If host is set, it also checks if host belongs to the dcs.
+func ShouldRepairRing(ring scyllaclient.Ring, dcs []string, host string) bool {
+	repairedDCs := strset.New(dcs...)
+	if host != "" {
+		if dc, ok := ring.HostDC[host]; !ok || !repairedDCs.Has(dc) {
+			return false
+		}
+	}
+
+	switch ring.Replication {
+	case scyllaclient.SimpleStrategy:
+		// Check range consisting of excluded hosts
+		excluded := 0
+		for _, dc := range ring.HostDC {
+			if !repairedDCs.Has(dc) {
+				excluded++
+			}
+		}
+		return ring.RF > excluded+1
+	case scyllaclient.NetworkTopologyStrategy:
+		rep := 0
+		for dc, rf := range ring.DCrf {
+			if repairedDCs.Has(dc) {
+				rep += rf
+			}
+		}
+		return rep > 1
+	default:
+		return false
+	}
+}
+
 func hostMaxRanges(shards map[string]uint, memory map[string]int64) map[string]Intensity {
 	out := make(map[string]Intensity, len(shards))
 	for h, sh := range shards {
--- a/pkg/service/repair/plan.go
+++ b/pkg/service/repair/plan.go
@@ -15,13 +15,38 @@ import (
 
 // plan describes whole repair schedule and state.
 type plan struct {
-	Keyspaces []keyspacePlan
-	Idx       int // Idx of currently repaired keyspace
+	Keyspaces keyspacePlans
 
-	SkippedKeyspaces []string
+	Hosts            []string
 	MaxParallel      int
 	MaxHostIntensity map[string]Intensity
-	HostTableSize    map[scyllaclient.HostKeyspaceTable]int64
+	// Used for progress purposes
+	Stats map[scyllaclient.HostKeyspaceTable]tableStats
+}
+
+type keyspacePlans []keyspacePlan
+
+// keyspacePlan describes repair schedule and state for keyspace.
+type keyspacePlan struct {
+	Keyspace string
+	Size     int64
+	Tables   []tablePlan
+}
+
+// tablePlan describes repair schedule and state for table.
+type tablePlan struct {
+	Table         string
+	Size          int64
+	RangesCnt     int
+	ReplicaSetCnt int
+	// Optimized tables (small and fully replicated)
+	// have all ranges for replica set repaired in a single job.
+	Optimize bool
+}
+
+type tableStats struct {
+	Size   int64
+	Ranges int
 }
 
 func newPlan(ctx context.Context, target Target, client *scyllaclient.Client) (*plan, error) {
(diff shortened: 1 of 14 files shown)


Michal-Leszczynski:
On the second thought, it is also possible to keep repair plan organized per keyspace / table. It might require less changes.

Michal-Leszczynski:
We could also change progress to keep stats of only currently repaired table, but this can be a part of another PR.

Michal-Leszczynski:
[Dtests](https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-dtest/1408/) passed except for one, not connected test.

pkg/service/repair/plan.go
```diff
@@ -140,104 +154,111 @@ func (p *plan) Units() []Unit {
...
 
-// SetMaxParallel sets maximal repair parallelism.
-func (p *plan) SetMaxParallel(dcMap map[string][]string) {
-	var max int
-	for _, kp := range p.Keyspaces {
-		// Max parallel is equal to the greatest max keyspace parallel
-		if cand := kp.maxParallel(dcMap); max < cand {
-			max = cand
+func perHostTableSize(ctx context.Context, client *scyllaclient.Client, hkts []hkt) (map[hkt]int64, error) {
+	report, err := client.TableDiskSizeReport(ctx, hkts)
```
karol-kokoszka, line 158: OMG. I know this is something what was there before, but looks like `client.TableDiskSizeReport` matches reports to `hkts []hkt` basing on the index...
It's blurry...
Why not to change it, to extend the `scyllaclient.HostKeyspaceTable` with size information.
Then, instead of matching by ids, just modify the scyllaclient.HostKeyspaceTable by adding the size, or returning new struct that contains the size ?

I see no reason why this logic is sooo convulted.

This method `perHostTableSize` would not be needed then.

pkg/service/repair/plan.go
```diff
@@ -13,124 +13,138 @@ import (
...
+				hkts = append(hkts, hkt{Host: h, Keyspace: ksp.Keyspace, Table: tp.Table})
 			}
 		}
-		tabIdx = 0
 	}
+	hostSize, err := perHostTableSize(ctx, client, hkts)
+	if err != nil {
+		return nil, errors.Wrap(err, "calculate tables size")
+	}
+	fillSize(hostSize, keyspaces)
```
karol-kokoszka, line 115: WDYT about extending `[]keyspacePlan` struct with the method `.fillSize(hostSize) // or with hkt extended with size` ?
It is easier to understand 
`keyspaces.fillSize(...)` than `filleSize(a, b)`. Second example makes me wonder what size I'm gonna update and where.

pkg/service/repair/plan.go
```diff
@@ -13,124 +13,138 @@ import (
...
+	fillOptimize(keyspaces, target.SmallTableThreshold)
```
karol-kokoszka, line 116: What do you think about moving `fillOptimize` under `keyspaces` authority ? 
...and to call `keyspaces.fillOptimize` ? (as the comment on fillSize)

pkg/service/repair/plan.go
```diff
@@ -13,124 +13,138 @@ import (
...
+	Hosts            []string
 	MaxParallel      int
 	MaxHostIntensity map[string]Intensity
-	HostTableSize    map[scyllaclient.HostKeyspaceTable]int64
+	// Used for progress purposes
+	HostSize   map[hkt]int64
+	HostRanges map[hkt]int
 }
 
 func newPlan(ctx context.Context, target Target, client *scyllaclient.Client) (*plan, error) {
```
karol-kokoszka, line 30: This constructor is pretty big.

Is there a chance to either:
- update this method with comments explaining what step is it and what it's gonna do ...
- ... or (better) split this bit too big constructor to smaller pieces (functions) having self explanatory names ?

pkg/service/repair/plan.go
```diff
@@ -13,124 +13,138 @@ import (
...
-	Idx       int // Idx of currently repaired keyspace
 
-	SkippedKeyspaces []string
+	Hosts            []string
 	MaxParallel      int
 	MaxHostIntensity map[string]Intensity
-	HostTableSize    map[scyllaclient.HostKeyspaceTable]int64
+	// Used for progress purposes
+	HostSize   map[hkt]int64
+	HostRanges map[hkt]int
```
karol-kokoszka, line 27: Maybe it's worth to join Hosts, HostSize, HostRange into one struct ?

pkg/service/repair/plan.go
```diff
@@ -422,31 +319,32 @@ func setIntersection(s1, s2 *strset.Set) *strset.Set {
...
-	Table string
-	Size  int64
-	// Deleted tables are still being sent to workers,
-	// so that their progress can still be updated in a fake way,
-	// as their jobs are not actually sent to Scylla.
-	Deleted bool
+	Table         string
+	Size          int64
+	RangesCnt     int
+	ReplicaSetCnt int
```
karol-kokoszka, line 325: Maybe `ReplicationFactor` is the better name here ? Especially that it's just a counter, not the reference to nodes belonging to replica.

pkg/service/repair/plan.go
```diff
@@ -422,31 +319,32 @@ func setIntersection(s1, s2 *strset.Set) *strset.Set {
 
 // tablePlan describes repair schedule and state for table.
 type tablePlan struct {
```
karol-kokoszka, line 321: Please sort this source code file.

You can follow the "rules":

> Structs at the top: It's a common practice to define structs at the top of the file, especially if they are widely used within the file or package. Placing them at the top makes it easier for developers to locate them quickly when reading the code.

> Methods near the struct definition: While it's not mandatory, organizing methods near the struct definition can improve code readability. When methods are placed close to the struct they operate on, it's easier for developers to understand the relationship between the two and to find related methods quickly.

> Group related methods: If a struct has many methods, it's a good practice to group related methods together. For example, constructors or initialization methods can be grouped at the top, followed by getters and setters, then other utility methods.

> Order by visibility: When grouping methods within a struct, consider ordering them by visibility (public before private). Public methods are those that should be accessible from outside the package, while private methods are internal to the package.

> Use comments for clarity: Regardless of where you place structs or methods, use comments effectively to document their purpose, behavior, and any important details. Clear comments can help other developers understand your code more easily.

Especially, please move tablePlan, keyspacePlan to the top. plan and its methods let be next to each other.

pkg/service/repair/service.go
```diff
@@ -149,37 +149,25 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
...
-	t.plan, err = newPlan(ctx, t, client)
+	p, err := newPlan(ctx, t, client)
 	if err != nil {
 		return t, errors.Wrap(err, "create repair plan")
 	}
 	// Sort plan
-	t.plan.SizeSort()
-	t.plan.PrioritySort(NewInternalTablePreference())
+	p.SizeSort()
+	p.PrioritySort(NewInternalTablePreference())
```
karol-kokoszka, line 158: This is just for me to have better understanding.

Why do you sort the plan here, on the `GetTarget`, but not on the actual repair `service.Repair` ?

pkg/service/repair/progress_integration_test.go
```diff
@@ -25,6 +26,10 @@ import (
 	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
 )
 
+func fhkt(host, keyspace, table string) hkt {
```
karol-kokoszka, line 29: What `fhkt` means ? 
How should I interpret this name seeing it at the first time ?

pkg/service/repair/progress.go
```diff
@@ -55,14 +48,14 @@ type dbProgressManager struct {
 	total          atomic.Float64   // Total weighted repair progress
 	tableSize      map[string]int64 // Maps table to its size
 	totalTableSize int64            // Sum over tableSize
-	keyspaceRanges map[string]int64 // Maps keyspace to its range count
+	tableRanges    map[string]int   // Maps keyspace to its range count
```
karol-kokoszka, line 51: Is the comment still valid ? If it is, then why `tableRanges` ?

pkg/service/repair/progress.go
```diff
@@ -31,19 +31,12 @@ type ProgressManager interface {
 	OnJobStart(ctx context.Context, job job)
 	// OnJobEnd must be called when single repair job is finished.
 	OnJobEnd(ctx context.Context, job jobResult)
-	// UpdatePlan marks plan's ranges which were already successfully
-	// repaired in the previous run.
-	UpdatePlan(plan *plan)
+	// GetDoneRanges returns ranges already successfully repaired in the previous runs.
+	GetDoneRanges(keyspace, table string) []scyllaclient.TokenRange
```
karol-kokoszka, line 35: Maybe `GetCompletedRanges` ?

karol-kokoszka (commented):
My brain compiler is too weak to verify that all changes made to generator won't break the repair process.
The code itself (of generator.go) looks good.
I believe that integration-tests of the repair should catch pottential bugs. Do you agree with me ? Maybe it's worth to execute the SCT sanity against this branch too ?

pkg/service/repair/service.go
Michal-Leszczynski: So that the `--dry-run` shows an actual repair order.

pkg/service/repair/progress_integration_test.go
Michal-Leszczynski: It's just a 1-line wrapper to reduce test code length, but it can be removed/renamed in favor of more readability.

pkg/service/repair/progress.go
Michal-Leszczynski: It should be: `Maps table to its range count`.

pkg/service/repair/plan.go
```diff
@@ -13,124 +13,138 @@ import (
 	"github.com/scylladb/scylla-manager/v3/pkg/util/slice"
 )
 
+type hkt = scyllaclient.HostKeyspaceTable
```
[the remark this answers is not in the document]
Michal-Leszczynski: >  The only reason I see is to have shorter ... code lines.

That's exactly the reason I used that :p. It's annoying to use that long name for a simple container for 3 strings, but it's true that it reduces readability.

Michal-Leszczynski:
@karol-kokoszka It seems like additional fields in Ring added in  #3747 would be useful in this PR. Having the dc -> rf mapping would make keyspace filtering and max parallel calculation way cleaner. I will refactor and rebase this PR on top of master when #3747 gets merged.

pkg/service/repair/plan.go
Michal-Leszczynski: Actually, it counts the amount of different replica sets, not the replica set size.

Michal-Leszczynski:
@karol-kokoszka I decided to fix #3745 in this PR as it is used in refactored code. I also tried to address all comments above. Sorry for a large PR, but could you please take a look at it again?

I have some problems with building and running tests on [jenkins](https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-build/646/). I will retry them tomorrow.

pkg/service/repair/plan.go
```diff
@@ -74,7 +74,9 @@ func newPlan(ctx context.Context, target Target, client *scyllaclient.Client) (*
 		}
 
 		// Update max parallel
-		maxP = max(maxP, MaxRingParallel(ring, target.DC))
+		if ringP := MaxRingParallel(ring, target.DC); maxP < ringP {
+			maxP = ringP
+		}
```
karol-kokoszka, line 79: I understand it's due to the failing jenkins job. Please revert. Releng will bump go to the correct version (unless we are really in a hurry)

Michal-Leszczynski: done

karol-kokoszka:
> @karol-kokoszka It seems like additional fields in Ring added in https://github.com/scylladb/scylla-manager/pull/3747 would be useful in this PR. Having the dc -> rf mapping would make keyspace filtering and max parallel calculation way cleaner. I will refactor and rebase this PR on top of master when https://github.com/scylladb/scylla-manager/pull/3747 gets merged.

But you closed the PR.

Michal-Leszczynski:
> But you closed the PR.

Yes, I mentioned above that I decided to have those fixes here as they are important for this PR.

karol-kokoszka (approved):
:+1:

Michal-Leszczynski:
Test status:
- dtests - [passed](https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-dtest/1422/)
- centos-satnity-test - failed due to instance termination - [rekicked](https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/centos-sanity-test/1119/)
- ubuntu22-sanity-test - failed due to instance termination - [rekicked](https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/ubuntu22-sanity-test/412/)

Michal-Leszczynski:
@karol-kokoszka are we ok to merge or do we want to investigate the `ubuntu22-sanity-test` failure first?

karol-kokoszka:
I checked it and looks that it failed on your run. Previous builds failed due to spot instance being terminated.
Unfortunately I don't see a clear reason of why this SCT run failed ....
Did you try with other sanity tests ? Is it only `ubuntu-22-sanity-test` that fails ?

Michal-Leszczynski:
> Did you try with other sanity tests ? Is it only ubuntu-22-sanity-test that fails ?

`centos-satnity-test` passes

karol-kokoszka:
Merge it, then please check the output of this job execution after merging to master. If it will appear once again without the exact reason in logs (as it is now), please fill the ticket in https://github.com/scylladb/scylla-cluster-tests

PR #3720, opened 2024-02-19
Deprecate interval and mix start_date with cron

### LINKED ISSUE #3700: sctool claims `start-date` is deprecated in favor of `cron`, yet `cron` does not cover its functionality
As described in the title, [`sctool` flags claim that `start-date` is deprecated in favour of `cron`](https://github.com/scylladb/scylla-manager/blob/d96e5d8338dd0afa5e130ee8cbb88b9cb11bc94b/pkg/command/flag/task.yaml#L39), yet `cron` does not fully cover its functionality. For instance, it is not possible to run a one-off task specified with `cron` (in its standard implementation).

cc @Michal-Leszczynski

### LINKED ISSUE #3701: Confusing `start-date` deprecation
[sctool claims that `start-date` is deprecated](https://github.com/scylladb/scylla-manager/blob/d96e5d8338dd0afa5e130ee8cbb88b9cb11bc94b/pkg/command/flag/task.yaml#L39), yet there is no mention of it in its [swagger specification](https://github.com/scylladb/scylla-manager/blob/d96e5d8338dd0afa5e130ee8cbb88b9cb11bc94b/swagger/scylla-manager.json#L276), nor the [generated model](https://github.com/scylladb/scylla-manager/blob/d96e5d8338dd0afa5e130ee8cbb88b9cb11bc94b/swagger/gen/scylla-manager/models/schedule.go#L34).
Is this only a deprecated flag in sctool, or should a user refrain from using it with the REST API as well? If the latter is the case, it should be documented.

cc @Michal-Leszczynski

### LINKED ISSUE #3702: `now+duration` start-date is broken by design
sctool's `start-date` supports `now+duration` syntax, which is broken by design.
While `now` is handled by sending a zero-value time to the manager server, `now+duration` is resolved on the CLI's side by taking current time and adding duration to it: https://github.com/scylladb/scylla-manager/blob/d96e5d8338dd0afa5e130ee8cbb88b9cb11bc94b/pkg/command/flag/type.go#L121 

With an offset shorter than the network trip, this results in an error:
```
scylla@scylla-manager-f9fd9cc87-rvhzm:/$ sctool backup --cluster=scylla/scylla --location=gcs:rzetelskik-test --start-date='now+1s'
Flag --start-date has been deprecated, use cron instead
Error: create task: start date of scheduled task cannot be in the past
Trace ID: GDVIyJ0wTCiJU-e_6BxFIw (grep in scylla-manager logs)
```

I believe this is broken by design, since it should be handled on the server's side, but the API doesn't allow it.
Perhaps the most confusing part is that with the current implementation, scheduling two tasks with 'now' and 'now+duration' might result in only the latter failing.

Whether the `start date of scheduled task cannot be in the past` error is reasonable is up for debate - maybe the server could run the task immediately if the start-date is in the past according to it. However, such behaviour would force the server to take an opinionated decision regarding the `interval` flag then.

Since `start-date` is deprecated in sctool (see https://github.com/scylladb/scylla-manager/issues/3701 and https://github.com/scylladb/scylla-manager/issues/3700), this bug should at least be documented as a known issue. Ideally, there should be a replacement for this functionality.

cc @Michal-Leszczynski @karol-kokoszka

karol-kokoszka:
Fixes #3701 
Fixes #3700 

This PR deprecates the `interval` param in swagger, `interval` flag in SM codebase.

Besides that, this PR extends Cron to include start date.
`start_date` mixed with `cron` is interpreted as "Run the task frequently according to cron spec, but the first execution must happen after `start_date`".

Fixes #3702 

This PR removes check on start_date when the interval flag is not introduced. More details https://github.com/scylladb/scylla-manager/issues/3702#issuecomment-1952938525

**How it was tested ?**

Integration tests (part of CI) (expected 4 to fail, schema restore in 5.4 and its enterprise equivalent)
Added unit tests to cover new functionality.

---

Please make sure that:

### DIFF
--- a/docs/source/sctool/partials/sctool_backup.yaml
+++ b/docs/source/sctool/partials/sctool_backup.yaml
@@ -126,8 +126,6 @@ options:
 - name: start-date
   shorthand: s
   usage: |
-    --start-date is deprecated, please use `--cron` instead
-
     The date can be expressed relatively to now or as a RFC3339 formatted string.
     To run the task in 2 hours use 'now+2h'. The supported units are:
 
--- a/docs/source/sctool/partials/sctool_backup_update.yaml
+++ b/docs/source/sctool/partials/sctool_backup_update.yaml
@@ -127,8 +127,6 @@ options:
 - name: start-date
   shorthand: s
   usage: |
-    --start-date is deprecated, please use `--cron` instead
-
     The date can be expressed relatively to now or as a RFC3339 formatted string.
     To run the task in 2 hours use 'now+2h'. The supported units are:
 
--- a/docs/source/sctool/partials/sctool_backup_validate.yaml
+++ b/docs/source/sctool/partials/sctool_backup_validate.yaml
@@ -72,8 +72,6 @@ options:
 - name: start-date
   shorthand: s
   usage: |
-    --start-date is deprecated, please use `--cron` instead
-
     The date can be expressed relatively to now or as a RFC3339 formatted string.
     To run the task in 2 hours use 'now+2h'. The supported units are:
 
--- a/docs/source/sctool/partials/sctool_backup_validate_update.yaml
+++ b/docs/source/sctool/partials/sctool_backup_validate_update.yaml
@@ -69,8 +69,6 @@ options:
 - name: start-date
   shorthand: s
   usage: |
-    --start-date is deprecated, please use `--cron` instead
-
     The date can be expressed relatively to now or as a RFC3339 formatted string.
     To run the task in 2 hours use 'now+2h'. The supported units are:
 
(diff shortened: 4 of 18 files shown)


Michal-Leszczynski:
Rolling back flag (`--start-date`) deprecation feels bad, but that's true that it covers some schedules that are not possible just with the `--cron` flag. The most annoying part of this flag is the extended `now+X` syntax (present only in sctool and not in swagger). I think that we should keep this syntax deprecated and eventually remove it.

Also, the extended `--cron` syntax is also problematic as `@every` is not parsed into cron expression (e.g. like `@hourly`), but is handled in the same way as `--interval` and from what I can see in the code, it can introduce drifts in task activations, because the next activation is calculated at the end of task run - not sure if it's expected/documented. 

Because of that, I think that all:  `--start-date 'now+X'`, `--interval`, `--cron '@every X'` should be deprecated.
Also, the last 2 flags could/should be parsed into cron expressions so that there are less ways in which SM calculates next task activation.

Michal-Leszczynski (approved):
LGTM, but please take a look at my comment and consider deprecating some parts of the problematic flags.

PR #3725, opened 2024-02-22
Tablet repair

Michal-Leszczynski:
WIP - have yet to test it against Scylla with tablets enabled.

### DIFF
--- a/pkg/scyllaclient/client_scylla.go
+++ b/pkg/scyllaclient/client_scylla.go
@@ -20,14 +20,13 @@ import (
 	"github.com/pkg/errors"
 	"github.com/scylladb/go-set/strset"
 	"github.com/scylladb/scylla-manager/v3/pkg/dht"
-	"github.com/scylladb/scylla-manager/v3/pkg/util/slice"
-	"go.uber.org/multierr"
-
 	"github.com/scylladb/scylla-manager/v3/pkg/util/parallel"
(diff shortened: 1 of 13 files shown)


Michal-Leszczynski:
I encountered:
```
agent [HTTP 500] std::runtime_error (The ranges option is not supported for tablet repair)"
```
when running repair against `scylla-nightly:latest`.

asias:
> I encountered:
> 
> ```
> agent [HTTP 500] std::runtime_error (The ranges option is not supported for tablet repair)"
> ```
> 
> when running repair against `scylla-nightly:latest`.

https://github.com/scylladb/scylladb/pull/17436
https://github.com/scylladb/scylladb/pull/17571

asias:
@Michal-Leszczynski I saw you closed the PR. Are the relevant patches merged already? Btw, you can start to test SM with the latest branch https://github.com/scylladb/scylladb/pull/17767. It contains all the necessary options support for SM.

Michal-Leszczynski:
This PR was closed because it was rebased on other repair related PR that was laying ground work for tablet repair and it got heavily refactored. I will get what's useful from this PR and test SM repair on tablets table today.

Michal-Leszczynski:
@asias I see that the host param is not suported. SM repair uses the host param in order to filter dcs and down nodes. It can be changed to use the dc param for dc filtering, but what about filtering out down nodes?

PR #3731, opened 2024-02-26
Backup: don't ignore uploadHost error

### LINKED ISSUE #3729: SM ignores backup `uploadHost` error
https://github.com/scylladb/scylla-enterprise/issues/3927
In the following fragment of SM backup code:
https://github.com/scylladb/scylla-manager/blob/69de42c491c577a5871861f7fd2af7d0e363d8fb/pkg/service/backup/worker_upload.go#L26-L32

we can see that `uploadHost` error is actually only used for logging and not returned from `f` function. This means that even when uploading host snapshot fails (a lot of `ERROR` logs), backup proceeds as nothing happened. This has a few consequences for a backup run that encountered an error (but didn't report it):
- SM reports task as `DONE` even though progress indicates otherwise (see [comment](https://github.com/scylladb/scylla-enterprise/issues/3927#issuecomment-1951133196))
- SM uploads manifest (even though not all files listed in manifest were uploaded) - restoring such manifest ends with a failure 
- SM removes snapshot files from Scylla nodes which makes it impossible to retry this backup (a fresh backup is required)

### LINKED ISSUE #3733: Backup: don't fail when there are no snapshot dirs
Consider the following scenario: backup stage upload failed on 1 host and fully succeeded on remaining hosts. Backup is retried, but it fails on all previously succeeded hosts because of this part of the code:
https://github.com/scylladb/scylla-manager/blob/69de42c491c577a5871861f7fd2af7d0e363d8fb/pkg/service/backup/worker_index.go#L160-L162
It's perfectly legal to have no snapshot dirs on resumed backup (assuming they were fully uploaded and deleted during previous run).

I'm hesitating whether just removing this check is fine or does it require more planned action. Note that this check was added because of #2426, but the description of this issue is lacking context.

Michal-Leszczynski:
Fixes #3729
Fixes #3733

### DIFF
--- a/pkg/service/backup/worker_snapshot.go
+++ b/pkg/service/backup/worker_snapshot.go
@@ -4,23 +4,12 @@ package backup
 
 import (
 	"context"
-	"time"
 
 	"github.com/pkg/errors"
 	. "github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
-	"github.com/scylladb/scylla-manager/v3/pkg/util/timeutc"
 )
 
 func (w *worker) Snapshot(ctx context.Context, hosts []hostInfo, limits []DCLimit) (err error) {
-	w.Logger.Info(ctx, "Taking snapshots...")
-	defer func(start time.Time) {
-		if err != nil {
-			w.Logger.Error(ctx, "Taking snapshots failed see exact errors above", "duration", timeutc.Since(start))
-		} else {
-			w.Logger.Info(ctx, "Done taking snapshots", "duration", timeutc.Since(start))
-		}
-	}(timeutc.Now())
-
 	f := func(h hostInfo) error {
 		w.Logger.Info(ctx, "Taking snapshots on host", "host", h.IP)
 		err := w.snapshotHost(ctx, h)
--- a/pkg/service/backup/service.go
+++ b/pkg/service/backup/service.go
@@ -810,8 +810,10 @@ func (s *Service) Backup(ctx context.Context, clusterID, taskID, runID uuid.UUID
 			}
 			defer clusterSession.Close()
 
-			w.AwaitSchemaAgreement(ctx, clusterSession)
-
+			if err := w.AwaitSchemaAgreement(ctx, clusterSession); err != nil {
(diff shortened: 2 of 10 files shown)


Michal-Leszczynski:
I will try to add tests later today.

Michal-Leszczynski:
This should be included in the 3.2.6 release.

karol-kokoszka:
Please add the test and change the PR from draft to regular one. 
I'll include it into 3.2.6 then.

Michal-Leszczynski:
@karol-kokoszka I added a test and it turns out that it detected another, not connected issue #3733.
When commenting out validation mentioned in this issue, this test failed before fa7dccf3b6f55ff357538044b4e46ebc09d0fddc and passes after it (but it always fails when the validation is left as it is).

pkg/service/backup/worker_snapshot.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/scylladb/scylla-manager/v3/pkg/util/timeutc"
 )
 
-func (w *worker) Snapshot(ctx context.Context, hosts []hostInfo, limits []DCLimit) (err error) {
+func (w *worker) Snapshot(ctx context.Context, hosts []hostInfo, limits []DCLimit) (err error) { // nolint: dupl
```
karol-kokoszka, line 14: you lazy bastard...
But frankly it's much better than customizing the insides

karol-kokoszka (approved):
Just one nit (it's about test code).
Other than that :+1:

pkg/service/backup/worker_snapshot.go
Michal-Leszczynski: I just wanted to include a quick fix so that it can be included in the release!

Michal-Leszczynski: I fixed it in 3e9a9de3.

Michal-Leszczynski:
@karol-kokoszka I decided to add a fix for failing tests in this PR. I also adressed the `// nolint: dupl` comment. Could you take a look at this PR one more time?

karol-kokoszka (approved):
:+1:

PR #3737, opened 2024-02-28
Repair: fix repair --host filtering

### LINKED ISSUE #3736: Incorrect repair --host node filtering
SM doesn't allow to repair with `--host` flag when table is not fully replicated (all tokens are on all hosts).
I was introduced when refactoring SM repair in `3.2.1`.
Ref https://github.com/scylladb/scylla-enterprise/issues/3948

Michal-Leszczynski:
Fixes #3736

### DIFF
--- a/pkg/service/repair/plan.go
+++ b/pkg/service/repair/plan.go
@@ -49,17 +49,23 @@ func newPlan(ctx context.Context, target Target, client *scyllaclient.Client) (*
 		}
 
 		skip := false
-		for _, rep := range ring.ReplicaTokens {
-			rtr := scyllaclient.ReplicaTokenRanges{
-				ReplicaSet: filteredReplicaSet(rep.ReplicaSet, filtered, target.Host),
-				Ranges:     rep.Ranges,
+		for _, rtr := range ring.ReplicaTokens {
+			// Skip the whole keyspace based on repaired dcs only
+			// (unless it's a single node cluster).
+			replicas := 0
+			for _, h := range rtr.ReplicaSet {
+				if slice.ContainsString(target.DC, ring.HostDC[h]) {
+					replicas++
+				}
 			}
-
-			// Don't add keyspace with some ranges not replicated in filtered hosts,
-			// unless it's a single node cluster.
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
Have yet to add a test for that.

Michal-Leszczynski:
The following test was failing after changing repair test setup and passed after applying the fix:
```
=== RUN   TestServiceRepairIntegration/repair_host
--- When: run repair
--- Then: repair is running
13:04:25.232	INFO	Start generating target	{"task": "87f5bbe7-4483-4304-998c-ca04d95f3baf", "run": "acf1dfb4-dbb1-11ee-9625-0892040e83bb"}
13:04:25.279	INFO	End generating target	{"task": "87f5bbe7-4483-4304-998c-ca04d95f3baf", "run": "acf1dfb4-dbb1-11ee-9625-0892040e83bb", "error": "create repair plan: no replicas to repair"}
    service_repair_integration_test.go:1242: timeout 1m0s
    --- FAIL: TestServiceRepairIntegration/repair_host (60.01s)
```

Michal-Leszczynski:
@karol-kokoszka why do we only test `2024.1.0.rc3` on PRs? I know that other versions are tested on push to master, but I think that it would be useful to see results of those tests before merging PRs to master.

karol-kokoszka:
> @karol-kokoszka why do we only test 2024.1.0.rc3 on PRs? I know that other versions are tested on push to master, but I think that it would be useful to see results of those tests before merging PRs to master

There were too many of these checks on a single PR. They are part of merging to master branch now.
You can still execute these tests locally.

karol-kokoszka:
Let's keep it as a check for the latest commercial release of Scylla that includes all variants.
Other version let be checked on master branch.

Michal-Leszczynski:
Ok, btw this PR is ready for review

Michal-Leszczynski:
@karol-kokoszka, actually, just pushed a small fix for a single node cluster, now it should be good to go.

karol-kokoszka (approved):
:+1:

PR #3748, opened 2024-03-11
Tablet backup and restore

### LINKED ISSUE #3703: Execute scylla-manager BACKUP & RESTORE integration tests against tablets enabled environment
To validate what additional work may be needed on the Scylla Manager side to provide full support for BACKUP & RESTORE procedures with Scylla having tablet support enabled, we need to create the possibility of establishing a tablet-enabled environment in the docker-compose environment we use for testing.

Here is a brief description of what happens during the backup process (https://github.com/scylladb/scylla-manager/blob/master/pkg/service/backup/README.md): Scylla Manager (SM) calls POST /storage_service/snapshot to prepare SSTables that are going to be backed up. An open question is whether the endpoint can be called at any given moment in a tablet-enabled environment or if there are any prerequisites for this action (cc: @bhalevy).

Here is a brief description of what happens during the restore process (https://github.com/scylladb/scylla-manager/blob/master/pkg/service/restore/README.md): POST /storage_service/sstables/{keyspace} is called with the load&stream flag enabled after copying files to the expected destination.

Backup suites will test if backup is possible in a tablet-enabled environment. Restore suites will answer whether we can back up the data in a tablet-enabled environment and bring it back to the new, tablet-enabled cluster.

This is just a supplement to the SCT tests that are supposed to check the backup and restore with tablets too.

Reference to scylla.yaml https://github.com/scylladb/scylladb/blob/0cbf8f75f0163898cd6359b843f8c9d3237177f3/conf/scylla.yaml#L277-L284 (cc: @Michal-Leszczynski )

Michal-Leszczynski:
Fixes #3703

### DIFF
--- a/.github/actions/test-setup/action.yml
+++ b/.github/actions/test-setup/action.yml
@@ -11,6 +11,9 @@ inputs:
   raft-enabled:
     description: "Specifies if cluster should use consistent_cluster_management option"
     required: false
+  tablets:
+    description: "Specifies if cluster should use tablets replication"
+    required: false
   start-dev-env:
     description: "Should this action run 'make start-dev-env'"
     required: false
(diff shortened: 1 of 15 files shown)


Michal-Leszczynski:
Closing in favor of #3753.

PR #3775, opened 2024-03-28
fix(session): close client after session creation

### LINKED ISSUE #3769: Number of connections established by Scylla Manager outgrows system capabilities
In one of QA suites for Scylla Operator test scheduled 20 ad-hoc repairs for single cluster and then Scylla Operator tried to update most of them. 
Manager wasn't able to validate the tasks, as it couldn't connect to healthy Scylla nodes due to lack of sockets.

```
2024-03-25T16:52:35.149228320Z {"L":"INFO","T":"2024-03-25T16:52:35.148Z","N":"cluster.client","M":"HTTP retry backoff","operation":"ColumnFamilyMetricsTotalDiskSpaceUsedByNameGet","wait":"1s","error":"dial tcp 172.20.163.164:10001: connect: cannot assign requested address","_trace_id":"rUqZzZC4RhWCQVPNkEdIHg"}
```

There're a lot of logs suggesting that new http client is created.
```
2024-03-25T17:07:31.364346000Z {"L":"INFO","T":"2024-03-25T17:07:31.364Z","N":"cluster","M":"Creating new Scylla HTTP client","cluster_id":"b4f10047-8d7f-490c-b829-5f60b5b56a1f","_trace_id":"cHtrJ-8rQlWJl-Rt9CMYJg"}
```

Looks like a http client/connection leak or unbound http client.

Run: https://jenkins.scylladb.com/job/scylla-operator/job/operator-1.12/job/eks/job/longevity-scylla-operator-3h-eks-repair-test/3/
Logs: https://cloudius-jenkins-test.s3.amazonaws.com/b1c45503-3f4f-4197-ae76-359835e2c0b2/20240325_174610/kubernetes-must-gather-b1c45503.tar.gz
Scylla Manager logs are under `must-gather/namespaces/scylla-manager/pods/scylla-manager-6f6556fd64-jcrrx`

karol-kokoszka:
Fixes #3769 

---

Please make sure that:

### DIFF
--- a/pkg/service/cluster/service.go
+++ b/pkg/service/cluster/service.go
@@ -578,6 +578,11 @@ func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID) (session 
 	if err != nil {
 		return session, errors.Wrap(err, "get client")
 	}
+	defer func() {
+		if err := client.Close(); err != nil {
+			s.logger.Error(ctx, "Couldn't close scylla client", "error", err)
+		}
+	}()
 
 	ni, err := client.AnyNodeInfo(ctx)
 	if err != nil {


Michal-Leszczynski:
I think that there are more places in the code which don't close client/session. Take a look at:
- client: `func (s *Service) ListNodes(ctx context.Context, clusterID uuid.UUID) ([]Node, error)`
- session: `func (s *Service) Restore(ctx context.Context, clusterID, taskID, runID uuid.UUID, properties json.RawMessage) error` (creation of restore worker)
- session: `func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties json.RawMessage) (Target, error)` (in repair pkg)

I think that closing client/session is really easy in those places. The bigger problem could be reliably closing cached clients, as cache provider can't just close expired client because it still might be used somewhere. This requires more planning and refactoring, but it's also less problematic since the clients are not recreated multiple times. I guess we can leave it for the future.

karol-kokoszka:
> I think that there are more places in the code which don't close client/session. Take a look at:

> client: func (s *Service) ListNodes(ctx context.Context, clusterID uuid.UUID) ([]Node, error)
session: func (s *Service) Restore(ctx context.Context, clusterID, taskID, runID uuid.UUID, properties json.RawMessage) error (creation of restore worker)
session: func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties json.RawMessage) (Target, error) (in repair pkg)
I think that closing client/session is really easy in those places. 

Yes, but this PR is not the general solution. These clients are taken from cache.
It closes the client just on session creation.

> The bigger problem could be reliably closing cached clients, as cache provider can't just close expired client because it still might be used somewhere. This requires more planning and refactoring, but it's also less problematic since the clients are not recreated multiple times. I guess we can leave it for the future.

Exactly, let's leave it for future. The situation with cached client and closing connection on them is something what we had already and we need more general solution.

Michal-Leszczynski:
> Yes, but this PR is not the general solution. These clients are taken from cache.
It closes the client just on session creation.

But the examples I mentioned above are not taken from cache and can be closed right after they are used, so I would include them in this PR but I would not touch cached clients here.

PR #3784, opened 2024-04-03
Switch from docker-compose to docker compose

Michal-Leszczynski:
This PR switches the usage of `docker-compose` to `docker compose`.
Outdated `docker-compose` has been the reason why some gh actions jobs were failing.


Michal-Leszczynski:
@karol-kokoszka this PR should fix fails on master gh actions.

karol-kokoszka (approved):
:+1:

PR #3785, opened 2024-04-03
Disable tablet migration on repair

### LINKED ISSUE #3773: Stop tablet load balancing during repair
Repairing should happen only when tablet load balancing is disabled. Without that, it would be possible for some tablets (mapped to ranges) to escape from being repaired by migrating from not repaired range to an already repaired one.

But what should happen when repair is paused? If SM re-enables tablet load balancing, then it would need to repair interrupted table from scratch (note that SM repairs table by table). If SM leaves tablet load balancing as disabled, then it might cause problems when paused repair task is never resumed.

I think that re-enabling it and starting repair of interrupted table from scratch is safer, but it would be nice to have other opinions on that.

Ref https://github.com/scylladb/scylladb/issues/17435

cc: @karol-kokoszka @tzach @bhalevy

Michal-Leszczynski:
Fixes #3773

### DIFF
--- a/pkg/scyllaclient/ring_describer.go
+++ b/pkg/scyllaclient/ring_describer.go
@@ -104,13 +104,14 @@ func getTabletKs(ctx context.Context, client *Client) *strset.Set {
 	out := strset.New()
 	// Assume that errors indicate that endpoints rejected 'replication' param,
 	// which means that given Scylla version does not support tablet API.
-	// Other errors will be handled on other API calls.
 	tablets, err := client.ReplicationKeyspaces(ctx, ReplicationTablet)
 	if err != nil {
+		client.logger.Info(ctx, "Couldn't list tablet keyspaces", "error", err)
 		return out
 	}
 	vnodes, err := client.ReplicationKeyspaces(ctx, ReplicationVnode)
 	if err != nil {
+		client.logger.Info(ctx, "Couldn't list vnode keyspaces", "error", err)
 		return out
 	}
 	// Even when both API calls succeeded, we need to validate
--- a/pkg/scyllaclient/ring_describer.go
+++ b/pkg/scyllaclient/ring_describer.go
@@ -18,6 +18,9 @@ type RingDescriber interface {
 	Reset(ctx context.Context)
 	// IsTabletKeyspace returns true if keyspace has tablet replication.
 	IsTabletKeyspace(keyspace string) bool
+	// ControlTabletLoadBalancing enables or disables tablet load balancing
+	// only when there are any tablet tables detected.
+	ControlTabletLoadBalancing(ctx context.Context, enabled bool) error
 }
 
 type ringDescriber struct {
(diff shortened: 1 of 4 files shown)


karol-kokoszka:
@Michal-Leszczynski I know I overlooked in some previous review, but this part (not introduced in this PR directly)
```
// getTabletKs returns set of tablet replicated keyspaces.
func getTabletKs(ctx context.Context, client *Client) *strset.Set {
	out := strset.New()
	// Assume that errors indicate that endpoints rejected 'replication' param,
	// which means that given Scylla version does not support tablet API.
	// Other errors will be handled on other API calls.
	tablets, err := client.ReplicationKeyspaces(ctx, ReplicationTablet)
	if err != nil {
		return out
	}
	vnodes, err := client.ReplicationKeyspaces(ctx, ReplicationVnode)
	if err != nil {
		return out
	}
	// Even when both API calls succeeded, we need to validate
	// that the 'replication' param wasn't silently ignored.
	out.Add(tablets...)
	if out.HasAny(vnodes...) {
		return strset.New()
	}
	return out
}
```
Is problematic...

Errors are hidden.
Can you log these errors ?

pkg/service/repair/progress.go
```diff
@@ -261,12 +262,12 @@ func (pm *dbProgressManager) emptyState() bool {
 	return true
 }
 
-func (pm *dbProgressManager) GetCompletedRanges(keyspace, table string) []scyllaclient.TokenRange {
+func (pm *dbProgressManager) GetCompletedRanges(keyspace, table string) (doneRanges []scyllaclient.TokenRange, allRangesCnt int) {
```
karol-kokoszka, line 265: What you need named return parameters for ?
They are not used.

karol-kokoszka (commented):
Please address the general comment about logging error and it's good to go.

pkg/service/repair/progress.go
Michal-Leszczynski: It was a tip from linter. It suggests to name returned values when their type doesn't make it obvious.

Michal-Leszczynski:
@karol-kokoszka the logging comment was addressed. Could you take a look?

Michal-Leszczynski:
@karol-kokoszka could you take a look?

pkg/scyllaclient/ring_describer.go
```diff
@@ -18,6 +18,9 @@ type RingDescriber interface {
 	Reset(ctx context.Context)
 	// IsTabletKeyspace returns true if keyspace has tablet replication.
 	IsTabletKeyspace(keyspace string) bool
+	// ControlTabletLoadBalancing runs scyllaclient ControlTabletLoadBalancing
+	// only when there are any tablet tables detected.
```
karol-kokoszka, line 22: Please change the comment. To inform that is about enabling/disabling tablets load balancing.

karol-kokoszka (approved):
LGTM !

Issue #3786, opened 2024-04-04
Warning/error does not make sense on 'sctool task list'


karol-kokoszka:
> [sa_107723664434226916601@escylla-1pmonitor-prod-eu ~]$ sctool task list
Command "list" is deprecated, use sctool instead.

Should be 
`Command "list" is deprecated, use 'sctool tasks' instead.`

PR #3787, opened 2024-04-04
fix(msg): update deprecate message on sctool task list

karol-kokoszka:
Fixes https://github.com/scylladb/scylla-manager/issues/3786

---

Please make sure that:

### DIFF
--- a/pkg/command/legacy/task/tasklist/res.yaml
+++ b/pkg/command/legacy/task/tasklist/res.yaml
@@ -1,7 +1,7 @@
 use: list
 
 deprecated: |
-  use sctool instead.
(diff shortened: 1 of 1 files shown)


Michal-Leszczynski:
@karol-kokoszka could you exclude non related auto generated code from this PR? Especially that it looks like it just rearranges some fragments of the code, but it doesn't change them in a meaningful way. Also, why does the commit have `fix(msg)`? Shouldn't it be something like `fix(docs)` or `fix(legacy)` or `fix(tasklist)`?

karol-kokoszka:
generated code is removed

Issue #3767, opened 2024-03-26
Cache NodeInfo outside of healthcheck service


karol-kokoszka:
The purpose of the health check service is simple: to report whether:
- The agent's API (REST ping) is reachable.
- The alternator session (alternator ping) is reachable.
- The CQL session (CQL ping) is reachable.

Scylla Manager may report false positives for the CQL ping and alternator ping if the agent's API is unresponsive or overloaded, and does not respond within the expected time. This occurs because both the CQL and alternator pings involve retrieving basic information about the nodes using the agent's node_info endpoint, which concatenates configuration-related responses from the Scylla API and returns them to the caller.

NodeInfo is necessary to properly build the client and contains information about encryption and ports configured to establish the session. However, this setup causes the health check service not only to verify the CQL session but also assumes that the agent's API is fully responsive. If the API is not responsive at the expected level, the health check produces false positives regarding the ability to create the CQL session and query the data.

```
          /--- agent's API to get Scylla's config (and cache it for short time)
Manager   ---- create CQL session with a single node and query simple data

set the status of healthcheck_cql_state basing on these two calls
```

The logic must be changed, and the health check service MUST be decoupled from the agent completely. To achieve this, the agent is expected to start a background goroutine that periodically checks Scylla's config and updates the cached config. The health check service is expected to maintain a reference to the cache and retrieve the latest Scylla config from there without directly interfering with the agent's API. If the API is unresponsive, then the cache may be outdated for some time, but situations where the configuration of a particular node changes are very rare.

**The goal:**

The health check should not call the agent's API at all.
Another service working in a separate goroutine is responsible for updating the cache.
The cache never expires; it is periodically updated by the ConfigCacheUpdater service.
The health check simply accesses the cache every time it needs Scylla config.

cc: @d-helios @gmizrahi @mykaul

Michal-Leszczynski:
> To achieve this, the agent is expected to start a background goroutine that periodically checks Scylla's config and updates the cached config.

So then SM would still need to query agent for node info (that would already be cached and wouldn't require interaction with Scylla), or did you mean that the background task and caching happens on SM side?

> The logic must be changed, and the health check service MUST be decoupled from the agent completely.

Simpler solution would be to just change healthchecks `nodeInfo` to return "expired" info when newer cannot be queried and try to update it the next time it's needed. On the other hand, current node info caching works only for healthcheck service, so perhaps decoupling it from healthcheck could mean that other services can use this cache as well.

[...]

PR #3803, opened 2024-04-15
feat(config-cache): initial stub for cluster config cache service

### LINKED ISSUE #3802: Initial stub
Part of https://github.com/scylladb/scylla-manager/issues/3767

karol-kokoszka:
Fixes #3802 

This is just an initial stub for new cluster-config-cache service.

General description of the goal is here https://github.com/scylladb/scylla-manager/issues/3767#issue-2208639491

---

Please make sure that:

### DIFF
--- a/pkg/service/configcache/service.go
+++ b/pkg/service/configcache/service.go
@@ -1,0 +1,223 @@
+// Copyright (C) 2024 ScyllaDB
+
+package configcache
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
+	"github.com/scylladb/scylla-manager/v3/pkg/secrets"
+	"github.com/scylladb/scylla-manager/v3/pkg/service"
+	"github.com/scylladb/scylla-manager/v3/pkg/service/cluster"
+	"github.com/scylladb/scylla-manager/v3/pkg/store"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
+)
+
+// ConnectionType defines an enum for different types of configuration.
+type ConnectionType int
+
+const (
+	// CQL defines cql connection type.
+	CQL ConnectionType = iota
+	// Alternator defines alternator connection type.
+	Alternator
+
+	updateFrequency = 5 * time.Minute
+)
+
+type tlsConfigWithAddress struct {
+	*tls.Config
+	Address string
+}
+
+// NodeConfig keeps the node current node configuration together with the TLS details per different type of connection.
+type NodeConfig struct {
+	*scyllaclient.NodeInfo
+	TLSConfig map[ConnectionType]*tlsConfigWithAddress
+}
+
+// ConfigCacher is the interface defining the cache behavior.
+type ConfigCacher interface {
+	Read(clusterID uuid.UUID, host string) (NodeConfig, error)
+}
+
+// Service is responsible for handling all cluster configuration cache related operations.
+// Use svc.Read(clusterID, host) to read the configuration of particular host in given cluster.
+// Use svc.Run() to let the cache update itself periodically with the current configuration.
+//
+// Current implementation doesn't care about changing IP, what means that the configuration for given cluster and host
+// will stay there forever.
+type Service struct {
+	clusterSvc   cluster.Servicer
+	scyllaClient scyllaclient.ProviderFunc
+	secretsStore store.Store
+
+	configs sync.Map
+}
+
+// Read returns either the host configuration that is currently stored in the cache,
+// ErrNoClusterConfig if config for the whole cluster doesn't exist,
+// or ErrNoHostConfig if config of the particular host doesn't exist.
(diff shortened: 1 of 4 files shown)


pkg/service/configcache/service.go
```diff
@@ -0,0 +1,223 @@
...
+type ConnectionType int
+
+const (
+	// CQL defines cql connection type.
+	CQL ConnectionType = iota
+	// Alternator defines alternator connection type.
+	Alternator
+
+	updateFrequency = 30 * time.Minute
+)
```
Michal-Leszczynski, line 31: I think it would be nicer to separate enums and other consts by placing them in separate blocks.
What do you think?

pkg/service/configcache/service.go
```diff
@@ -0,0 +1,223 @@
...
+			}
+			if clientCertAuth && errors.Is(err, service.ErrNotFound) {
+				return config, errors.Wrap(err, "client encryption is enabled, but certificate is missing")
+			}
+			config.TLSConfig[p] = &tlsConfigWithAddress{
+				Config:  tlsConfig,
+				Address: address,
+			}
+		}
+	}
```
Michal-Leszczynski, line 198: I know that's how it was done before, but I don't see the reasons for keeping a map consisting of only two entries. Also this for loop which fills this map is cloudy as CQL and ALTERNATOR are still if-ed in the first half of the loop. Also getting configs from map is less clear (are the configs always there? Should we validate that every time we want to get them?)

Wouldn't it be better to directly store 2 TLS configs in Node config?

Michal-Leszczynski (commented):
Looks good, just some minor comments.

pkg/service/configcache/service.go
```diff
@@ -0,0 +1,223 @@
...
+// ConnectionType defines an enum for different types of configuration.
+type ConnectionType int
+
+const (
+	// CQL defines cql connection type.
+	CQL ConnectionType = iota
+	// Alternator defines alternator connection type.
+	Alternator
+
+	updateFrequency = 30 * time.Minute
```
Michal-Leszczynski, line 30: Why do we update so rarely? Shouldn't it be something like 5min?

pkg/service/configcache/service.go
```diff
@@ -0,0 +1,223 @@
...
+
+		go func() {
+			client, err := svc.scyllaClient(ctx, c.ID)
+			if err != nil {
+				fmt.Println(err)
+			}
+
+			// Hosts that are going to be asked about the configuration are exactly the same as
+			// the ones used by the scylla client.
+			hostsWg := sync.WaitGroup{}
```
Michal-Leszczynski, line 141: I think it would be safer to wait for all updates before returning from this function. Otherwise it's possible that a single cluster will be updated in parallel by different calls to update function.

pkg/service/configcache/service.go
```diff
@@ -0,0 +1,223 @@
...
+
+	for _, c := range clusters {
+		c := c
+		currentClusterConfig, err := svc.readClusterConfig(c.ID)
+		if err != nil && errors.Is(err, ErrNoClusterConfig) {
+			fmt.Println(err)
+		}
+
+		go func() {
+			client, err := svc.scyllaClient(ctx, c.ID)
```
Michal-Leszczynski, line 134: Just a note that this opens a client without closing it. We probably can't close it because of the cache, but then maybe it makes sense to create client from scratch so that it can be closed right after?

pkg/service/configcache/service.go
karol-kokoszka: It actually can, I have no strong opinion though.
If the cache is not updated, but some configuration changed in meantime, then manager may misbehave for some time (if we spread the cache access to all services, not the healtcheck only).

But this is something what we would met with the current way of caching the nodeinfo as well.
Let me bring the default from previous approach. So the 5min you mentioned.

pkg/service/configcache/service.go
karol-kokoszka: And what is the benefit of doing so over keeping them in one block split with empty line ?
There is actually no enum in Go, the ConnectionType is just a work around for this lack of enum. 

Compiler would still accept 
```
	config.TLSConfig[5] = &tlsConfigWithAddress{}
```

pkg/service/configcache/service.go
karol-kokoszka: It actually doesn't lead to TCP connections leak. Here is why.
Keep-alive is enabled by default on HTTP level in go, what means that the client will reuse TCP connection for multiple requests to the same server. That is good.
Scylla-manager explicitly sets the keep-alive on TCP level too https://github.com/scylladb/scylla-manager/blob/701b8cb300a7f677f3524a80629bd411a67952c7/pkg/scyllaclient/client.go#L52 (and it's enabled by default in go btw but the value is 15 seconds).
Setting keep-alive on TCP level means that inactive connections are eventually detected and closed by the operating system.

The scylla client cache is OK as is, we don't need to explicitly close the client.

pkg/service/configcache/service.go
karol-kokoszka: You are right.
Created an issue to not jump into the chaos with the PRs.
It will be addressed as a part of the epic.

karol-kokoszka:
@Michal-Leszczynski I'm merging it to feature branch.

PR #3811, opened 2024-04-17
docs: explain --all-clusters in backup list

### LINKED ISSUE #3780: 'sctool backup list --all-clusters` requires cluster id
Sctool supports command `sctool backup list --all-clusters -L s3:<location>` which is expected to list all snapshots that exist in the given location.
The command `sctool backup` requires to include cluster-id, what makes no sense for listing all snapshots from given location.

Example:
```
➜  scylla-manager git:(branch-3.2) ./sctool backup list --all-clusters -L s3:backuptest-smoke 
Error: required flag(s) "cluster" not set
```

, but works fine when cluster is specified and lists snapshots from other clusters too
```
➜  scylla-manager git:(branch-3.2) ./sctool backup list -c 2db3cfc5-e0bd-4832-ab13-652bff9a7bec --all-clusters -L s3:backuptest-smoke
Cluster: 2db3cfc5-e0bd-4832-ab13-652bff9a7bec

backup/aa327c5e-c12e-4340-a47d-70d0f2dcdf03
Snapshots:
  - sm_20240403092454UTC (3.900M, 6 nodes)
Keyspaces:
  - system_auth (4 tables)
  - system_traces (5 tables)
  - system_distributed (4 tables)
  - system_distributed_everywhere (1 table)
  - system_schema (15 tables)

Cluster: ab667845-996e-46a5-9699-9c3b3c792443

backup/eb28b979-c128-453d-b15c-88c0783f46f2
Snapshots:
  - sm_20240321143455UTC (3.595M, 3 nodes)
  - sm_20240321143454UTC (3.595M, 3 nodes)
Keyspaces:
  - backuptest_data (1 table)
```

**Goal**
Do not require `-c` flag on `sctool backup list --all-clusters -L s3:<location>` if `--all-clusters` is used.

karol-kokoszka:
Fixes #3780 

---

Please make sure that:


karol-kokoszka:
@annastuchlik Could you please take a look at this PR ? There were a lot of question asking if customer can somehow remove the backup of non-existing cluster from the backup location.
They actually can do it with the manager by listing the backups first and then removing the one that is not needed.
`sctool backup list` requires to pass --cluster-id though, what may not intuitive at the first glance. There is an explanation for that and I included this explanation into the CLI description.

karol-kokoszka:
> If you're using the --all-clusters option, you still need to provide the --cluster-id of one of the clusters (and that cluster will work as a coordinator for listing backups of other clusters).
You need to provide --cluster-id of an existing cluster managed by Scylla Manager. If the cluster is deleted, the command will fail.

That's correct.

> I'm guessing because the description explains how it works under the hood, but it's unclear to me what the user should do.

That's very good point, and I agree with you that the explanation should be added to the general backup description.
Will do the followup commit just right after.

> This PR updates the description of the backup list command. How is that related to removing backups? Should the description of backup delete be updated, too?

Makes sense, let me update `backup delete` too.

> Finally, perhaps we should add instructions on how to remove backups to the [Backup page](https://manager.docs.scylladb.com/stable/backup/index.html).

Yes, will add the commit.

@annastuchlik Thanks for the follow up questions !

karol-kokoszka:
@annastuchlik PR updated.
Please have a look.

karol-kokoszka:
@annastuchlik ping. I appreciate your view on the documentation and don't want to merge changes without it.

PR #3817, opened 2024-04-22
Publish 3.2 docs

Michal-Leszczynski:
In order to display fixes/changes in SM 3.2 docs, commits need to be merged into the 3.2 branch first.


Michal-Leszczynski:
@karol-kokoszka let's update the docs with relevant changes before releasing new SM version.

karol-kokoszka (approved):
LGTM

PR #3824, opened 2024-04-24
Use small_table_optimization in repair

### LINKED ISSUE #3642: Manager: Support small_table_optimization feature
See https://github.com/scylladb/scylladb/pull/15974

Michal-Leszczynski:
This PR utilizes the small_table_optimization param in repair API which allows SM to repair the whole table with just one API call. It should bring significant improvements when repairing smaller tables.

As we can't check supported API parameters on Scylla side, we need to rely on checking Scylla version.

Fixes #3642

### DIFF
--- a/pkg/scyllaclient/client_agent.go
+++ b/pkg/scyllaclient/client_agent.go
@@ -158,6 +158,28 @@ func (ni NodeInfo) AlternatorTLSEnabled() (tlsEnabled, certAuth bool) {
 	return ni.AlternatorEncryptionEnabled(), certAuth
 }
 
+// SupportsRepairSmallTableOptimization returns true if /storage_service/repair_async/{keyspace} supports small_table_optimization param.
+func (ni *NodeInfo) SupportsRepairSmallTableOptimization() (bool, error) {
+	// Detect master builds
+	if scyllaversion.MasterVersion(ni.ScyllaVersion) {
+		return true, nil
+	}
+	// Check OSS
+	supports, err := scyllaversion.CheckConstraint(ni.ScyllaVersion, ">= 5.5, < 2000")
+	if err != nil {
+		return false, errors.Errorf("Unsupported Scylla version: %s", ni.ScyllaVersion)
+	}
+	if supports {
+		return true, nil
+	}
+	// Check ENT
+	supports, err = scyllaversion.CheckConstraint(ni.ScyllaVersion, ">= 2024.1.5")
+	if err != nil {
+		return false, errors.Errorf("Unsupported Scylla version: %s", ni.ScyllaVersion)
+	}
+	return supports, nil
+}
+
 // FreeOSMemory calls debug.FreeOSMemory on the agent to return memory to OS.
 func (c *Client) FreeOSMemory(ctx context.Context, host string) error {
 	p := operations.FreeOSMemoryParams{
--- a/pkg/service/repair/generator.go
+++ b/pkg/service/repair/generator.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/scylladb/go-log"
-	"github.com/scylladb/scylla-manager/v3/pkg/dht"
 	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
 )
 
@@ -33,8 +32,7 @@ type tableGenerator struct {
 	Ring         scyllaclient.Ring
 	TodoRanges   map[scyllaclient.TokenRange]struct{}
 	DoneReplicas map[uint64]struct{}
-	Optimize     bool
-	Deleted      bool
+	JobType      jobType
 	Err          error
 }
 
@@ -55,16 +53,22 @@ type submitter[T, R any] interface {
 	Close()
 }
 
+// jobType describes how worker should handle given repair job.
+type jobType int
+
+const (
+	normalJobType jobType = iota
+	skipJobType
+	mergeRangesJobType
+)
+
 type job struct {
 	keyspace   string
 	table      string
 	master     string
 	replicaSet []string
 	ranges     []scyllaclient.TokenRange
-	// jobs of optimized tables should merge all token ranges into one.
-	optimize bool
-	// jobs of deleted tables are sent only for progress updates.
-	deleted bool
+	jobType    jobType
 }
 
 type jobResult struct {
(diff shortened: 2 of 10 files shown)


Michal-Leszczynski:
Note: `small_table_optimization` doesn't work with tablets.

Michal-Leszczynski:
@karol-kokoszka PR is ready for review.

pkg/service/repair/plan.go
```diff
@@ -229,12 +235,11 @@ func (p keyspacePlans) fillSize(ctx context.Context, client *scyllaclient.Client
 	return sizeReport, nil
 }
 
-func (p keyspacePlans) fillOptimize(smallTableThreshold int64) {
+func (p keyspacePlans) fillSmall(smallTableThreshold int64) {
 	for _, ksp := range p {
 		for j, tp := range ksp.Tables {
-			// Return merged ranges for small, fully replicated table (#3128)
-			if tp.Size < smallTableThreshold && tp.ReplicaSetCnt == 1 {
```
karol-kokoszka, line 236: Is it the answer to my prev quesiton:
> I don't follow. Could you please elaborate why len(ring.ReplicaTokens) == 1 ?
Isn't tp.Small just enough ?
?
So, the logic is moved from here ?
But this is the number of replicasets,.

pkg/service/repair/plan.go
```diff
@@ -371,3 +376,28 @@ func newHostKsTable(host, ks, table string) scyllaclient.HostKeyspaceTable {
...
+		}
+		res, err := ni.SupportsRepairSmallTableOptimization()
+		if err != nil {
+			return parallel.Abort(err)
+		}
+		if !res {
+			out.Store(false)
+		}
+		return nil
+	}, parallel.NopNotify)
```
karol-kokoszka, line 398: It doesn't have much sens to use inhouse parallel.Run here IMHO
This https://pkg.go.dev/golang.org/x/sync/errgroup serves exactly the same purpose.

pkg/service/repair/service_repair_integration_test.go
```diff
@@ -1784,9 +1788,36 @@ func TestServiceRepairIntegration(t *testing.T) {
...
 		defer cancel()
 
+		// Check small_table_optimization support
+		ni, err := h.Client.AnyNodeInfo(ctx)
+		if err != nil {
+			t.Fatal(err)
+		}
+		support, err := ni.SupportsRepairSmallTableOptimization()
+		if err != nil {
+			t.Fatal(err)
```
karol-kokoszka, line 1798: Why ? Isn't it hidden assuption that all the envrionments that we use to test against supports smallTableOptimization ?
I suggest to use t.Skip("explanation") here

pkg/scyllaclient/client_agent.go
```diff
@@ -158,6 +158,28 @@ func (ni NodeInfo) AlternatorTLSEnabled() (tlsEnabled, certAuth bool) {
...
 }
 
+// SupportsRepairSmallTableOptimization returns true if /storage_service/repair_async/{keyspace} supports small_table_optimization param.
+func (ni *NodeInfo) SupportsRepairSmallTableOptimization() (bool, error) {
+	// Detect master builds
+	if scyllaversion.MasterVersion(ni.ScyllaVersion) {
+		return true, nil
+	}
+	// Check OSS
+	supports, err := scyllaversion.CheckConstraint(ni.ScyllaVersion, ">= 5.5, < 2000")
```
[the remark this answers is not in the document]
Michal-Leszczynski: This is just a convention (also used in tests) to distinguish OSS from ENT Scylla versions. Constraint  `">= 5.5"` is satisfied for all ENT versions, but this check is suppose to target OSS versions only.

pkg/service/repair/generator.go
```diff
@@ -158,20 +163,31 @@ func (g *generator) newTableGenerator(keyspace string, tp tablePlan, ring scylla
...
 		for _, r := range done {
 			delete(todoRanges, r)
 		}
 	}
 
+	var jt jobType
+	switch {
+	case g.plan.SmallTableOptSupport && tp.Small && !tabletKs:
+		jt = optimizeJobType
+	case len(ring.ReplicaTokens) == 1 && tp.Small:
```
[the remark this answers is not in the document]
Michal-Leszczynski: This is the requirement for using the old small table optimization - we send `min_token:max_token` only for fully replicated and small tables. I'm not sure about the reasoning behind table being fully replicated. I suspect that this is because otherwise SM would be sending `min_token:max_token` for each replica set and this could result in repairing ranges many times instead of just once. On the other hand we could just merge the ranges in a normal way (just concatenate them), but it was never done before.

pkg/service/repair/plan.go
Michal-Leszczynski: `tp.ReplicaSetCnt` and `len(ring.ReplicaTokens)` should be equal. I tried to answer the question in the reply to previous comment.

pkg/service/repair/service_repair_integration_test.go
Michal-Leszczynski: This function returns error only when there is a problem with parsing Scylla version (so it should never happen). When `small_table_optimization` is not supported, this function returns `false, nil`. We can't skip such a case, because SM should work fine with Scylla versions that does not support  `small_table_optimization`.

Michal-Leszczynski:
Ref for Scylla versions constraints: https://github.com/scylladb/scylla-enterprise/issues/4140#issuecomment-2095497858.

pkg/service/repair/plan.go
karol-kokoszka: Right, I found the logic in plan.go and the comment
```
// ReplicaTokenRanges describes all token ranges belonging to given replica set.
```

```
			for _, rep := range ring.ReplicaTokens {
				filtered := filterReplicaSet(rep.ReplicaSet, ring.HostDC, target)
				if len(filtered) == 0 {
					continue
				}

				replicaSetCnt++
				allHosts.Add(filtered...)

				for _, h := range filtered {
					ranges[newHostKsTable(h, u.Keyspace, t)] += len(rep.Ranges)
				}
				rangesCnt += len(rep.Ranges)
			}
```

Looks that ReplicaToken name is bit misleading, could be ReplicaTokensPerReplicaSet.

karol-kokoszka (approved):
:+1:

Issue #3219, opened 2022-08-23
Give possibility for labeling the clusters added to SM


Choraden:
Scylla-manager omits fields such as Username, Password, AuthToken in response to /clusters request.
For security reasons this is desired, but when it comes to Scylla-operator, clusters cannot be compared and updated when the only field that changes is e.g. password. 
  
That's how the cluster model looks like right now: 
https://github.com/scylladb/scylla-manager/blob/master/swagger/gen/scylla-manager/models/cluster.go

I would suggest adding a hash field for client to set, so that it will be possible to distinguish different clusters without any security risks.

zimnx:
The issue is there's no way to tell whether cluster you PUT into API is in the desired state, because GET doesn't return all the info - which is fine.
This leads to situation where from automation we can't tell whether credentials are updated or not.

Proposed solution is to add a `labels` field to the clsuters. User defined key-values which might be useful for filtering, tagging or putting arbitrary values like hash of entire cluster we put in API. 
This would allow on our side to compute hash of required object - with up-to-date credentials - and compare it with object in API. If hash mismatches, then we need to update the cluster.

rzetelskik:
This would be useful for us in tasks as well, as it would allow for a more robust assessment of the tasks being up to date, i.e. we wouldn't have to compare the properties of the desired task specification and the actual state in the manager. Ref: https://github.com/scylladb/scylla-operator/issues/1827.

The approach proposed by @zimnx, i.e. a map of labels/annotations, would work in both cases.

@karol-kokoszka can we have it triaged?

karol-kokoszka:
@rzetelskik @zimnx 
API of manager is returning what you want, have a look
```
➜  scylla-manager git:(master) curl 127.0.0.1:5080/api/v1/cluster/294e86f8-6133-4930-9d88-94ac83bcf9cf
{"id":"294e86f8-6133-4930-9d88-94ac83bcf9cf","name":"","host":"192.168.200.11","auth_token":"token","force_tls_disabled":false,"force_non_ssl_session_port":false}
➜  scylla-manager git:(master) curl 127.0.0.1:5080/api/v1/clusters                                    
[{"id":"294e86f8-6133-4930-9d88-94ac83bcf9cf","name":"","host":"192.168.200.11","auth_token":"token","force_tls_disabled":false,"force_non_ssl_session_port":false,"username":"set","password":"set"}]
```

What is missing exactly ?

rzetelskik:
[...]

PR #3828, opened 2024-04-26
Add `label` parameter to cluster and tasks

karol-kokoszka:
Fixes https://github.com/scylladb/scylla-manager/issues/3219

cc: @d-helios @zimnx 

---

Please make sure that:

### DIFF
--- a/docs/source/sctool/partials/sctool_backup.yaml
+++ b/docs/source/sctool/partials/sctool_backup.yaml
@@ -9,6 +9,9 @@ options:
   shorthand: c
   usage: |
     The target cluster `name or ID` (envvar SCYLLA_MANAGER_CLUSTER).
+- name: label
+  usage: |
+    Free form text field to put additional comments.
 - name: cron
   usage: |
     Task schedule as a cron `expression`.
--- a/docs/source/sctool/partials/sctool_backup_update.yaml
+++ b/docs/source/sctool/partials/sctool_backup_update.yaml
@@ -9,6 +9,9 @@ options:
   shorthand: c
   usage: |
     The target cluster `name or ID` (envvar SCYLLA_MANAGER_CLUSTER).
+- name: label
+  usage: |
+    Free form text field to put additional comments.
 - name: cron
   usage: |
     Task schedule as a cron `expression`.
--- a/docs/source/sctool/partials/sctool_cluster_add.yaml
+++ b/docs/source/sctool/partials/sctool_cluster_add.yaml
@@ -37,6 +37,9 @@ options:
     When a cluster is added, it is assigned a unique identifier.
     Use this parameter to identify the cluster by an alias name which is more meaningful.
     This `alias` name can be used with all commands that accept --cluster parameter.
+- name: label
+  usage: |
+    Free form text field to put additional comments.
 - name: password
   shorthand: p
   usage: |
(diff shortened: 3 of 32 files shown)


Michal-Leszczynski:
@karol-kokoszka this PR introduces additional string label for cluster/tasks, but I think that operator team requested it to be a map:

> Proposed solution is to add a labels field to the clsuters. User defined key-values which might be useful for filtering, tagging or putting arbitrary values like hash of entire cluster we put in API.

> The only thing we're requesting here is enriching the cluster/task structs with user-defined key-value pairs, allowing the users to annotate the objects - akin to https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/.

karol-kokoszka:
We won't introduce any map, the field is expected to serve general purpose. DevOps wants to use it as well to add their labels.
If operator wants to have the map, then can use JSON.

zimnx:
What's the problem with having a map for DevOps?
Single text fields is not extensible, doesn't allow for advanced querying or filtering. 
While with maps, users could search for clusters having particular key or set of keys.

rzetelskik:
> We won't introduce any map, the field is expected to serve general purpose. DevOps wants to use it as well to add their labels.

They can use a map for a single key, we can't use a label for a map - we'd collide with a user whenever they'd try to amend it.

d-helios:
> What's the problem with having a map for DevOps? Single text fields is not extensible, doesn't allow for advanced querying or filtering. While with maps, users could search for clusters having particular key or set of keys.

just to let you know that map for us is also ok.

karol-kokoszka:
@zimnx @rzetelskik I don't understand you guys.
Here is a comment from the issue that is expected to be closed with this PR https://github.com/scylladb/scylla-manager/issues/3219#issuecomment-2031727386
It's talking about label text field.

Why you cannot use JSON ?
Single text field is extensible, you can put whatever you want.

> While with maps, users could search for clusters having particular key or set of keys.

There is no functionality in manager to search for clusters using particular key or set of keys basing on this new label.
Nothing prevents you from building this kind of functionality on top of what SM returns.
Besides, the label field is not part of the key in the backing DB, you need to scan to get the data and just filter.

It's still manager client app that must implement the search.

rzetelskik:
> Here is a comment from the issue that is expected to be closed with this PR https://github.com/scylladb/scylla-manager/issues/3219#issuecomment-2031727386

It's your comment, not ours. We specifically ask for a key-value store there.

> Why you cannot use JSON ?
> Single text field is extensible, you can put whatever you want.

As long as the user doesn't fiddle with the label directly, sure. If they do, we lose all of the metadata and no longer know if this cluster/task should be controlled by the operator. And this is the only use case I care about personally here. 

I don't really understand the pushback - it seem like a pretty standard approach? Grafana, prometheus, kubernetes all seem to use maps for labels.

karol-kokoszka:
> As long as the user doesn't fiddle with the label directly, sure. If they do, we lose all of the metadata and no longer know if this cluster/task should be controlled by the operator. And this is the only use case I care about personally here.

Makes sense, but the current implementation (in this PR), just allows to put all or nothing. 
If there is a map of label fields, than it implies that there is completely new CLI and API in sctool to manage these labels.
It's achievable, but it won't be a part of 3.2.8.
It's because it's manager who will not only need to update the label field, but just update the diff.

The plan for 3.2.8 is to release it by 6th of May and the release is driven by https://github.com/scylladb/scylla-manager/issues/3767

Guys, if the label as a single text field is no go, then we must postpone it for a couple of weeks.

zimnx:
It's a no go for us. We can wait for it until next release.

Issue #3666, opened 2023-12-18
Allow setting restore and repair settings on restore command


pdbossman:
Two common restore scenarios:
1. Restore into target that is unavailable until restore completes (max out restore and repair to restore as fast as possible)
2. Restore into target that is live (may need to throttle)

It would be nice if a dry-run would output the restore and repair intensity and range of settings that are possible so I could see what the max options are, what SM has planned (particularly with defaults), and set it based on the type of restore scenarios I'm performing.

Additionally, please provide the ability to set the repair settings that will be used during restore on the restore command.  I don't want to have to monitor the (very long running) process to wait until the repair part begins to be able to adjust the speed (parallel and intensity).

I would like the ability to set "full speed restore" or "friendly neighbor" restore right up front.

PR #3832, opened 2024-05-02
fix(restore): post-restore repair to use max intensity and parallel

### LINKED ISSUE #3542: Run post-restore repair with max intensity
I think that SM should run post-restore repair with max intensity.
There are two reasons for that:
- restore procedure is usually the only thing putting pressure on the cluster
- recent changes to repair (SM 3.2) made it run slower (although with greater stability)

Right now SM runs the repair task with default intensity (1) and this change should speed-up the process.

### LINKED ISSUE #3823: Default parallelism and intensity for the repair after restore to MAX
This is the ghost issue, the real one is here https://github.com/scylladb/scylla-enterprise/issues/4118#issuecomment-2070890386

karol-kokoszka:
Fixes #3823 
Fixes #3542 

---

Please make sure that:

### DIFF
--- a/pkg/service/restore/tables_worker.go
+++ b/pkg/service/restore/tables_worker.go
@@ -246,7 +246,9 @@ func (w *tablesWorker) stageRepair(ctx context.Context) error {
 		}
 	}
 	repairProps, err := json.Marshal(map[string]any{
-		"keyspace": keyspace,
+		"keyspace":  keyspace,
+		"intensity": 0,
+		"parallel":  0,
 	})
 	if err != nil {
 		return errors.Wrap(err, "parse repair properties")


Michal-Leszczynski:
The reason for repair test failure: #3834.

Michal-Leszczynski:
I'm fine with this change and I actually created identical PR in the past that was rejected because this [comment](https://github.com/scylladb/scylla-manager/pull/3541#issuecomment-1698659134). Did we validate that `sctool repair control` works on restore repair as intended? Or maybe this change should be introduced only alongside #3666?

karol-kokoszka:
Right, would be good to have the `repair control` available for this post-restore.
Let's discuss on grooming.

karol-kokoszka:
@Michal-Leszczynski did you have a chance to check if repair control works when the post-restore repair is on-going ?

PR #3836, opened 2024-05-06
Skip know restore schema failures

### LINKED ISSUE #3835: Skip failing restore schema tests
Since introduction of Scylla 5.4, the regular restore schema tests have been failing when `consistent_cluster_management` is set to `true.` It is a know issue with a known [workaround](https://manager.docs.scylladb.com/stable/restore/restore-schema.html#restore-schema-workaround) that will be fixed only for Scylla 6.0. We should skip the restore schema tests that are executed against Scylla versions which are known to fail so that the red gh actions check actually means something useful.

cc: @karol-kokoszka

Michal-Leszczynski:
Fixes #3835


Michal-Leszczynski:
This will be a part of #3839.

Issue #3833, opened 2024-05-06
Include Scylla Enterprise nightly latest image to CI tests


karol-kokoszka:
Scylla Nightly may not be fully stable.
Example of failing repair against Scylla Nightly https://github.com/scylladb/scylla-manager/actions/runs/8925245333/job/24517270900?pr=3832

We must test against fully releases or release candidates only.

Michal-Leszczynski:
Mentioned test fails because of #3834, not Scylla Nightly on its own - but I agree that it would be good to have some more stable Scylla version in gh test matrix.

mykaul:
I disagree, we need to test against master. We can pick master that passed CI of course. @yaronkaikov  can help with that.

PR #3837, opened 2024-05-07
[GitHub nightly] run against Scylla-enterprise-nightly:latest

yaronkaikov:
Setting the nightly tests to run against the latest Scylla-enterprise nightly container rather than a specific version from OSS

Fixes: https://github.com/scylladb/scylla-manager/issues/3833

### DIFF
--- a/.github/workflows/integration-tests-nightly-ipv4-raftschema-rafttopology-tablets.yaml
+++ b/.github/workflows/integration-tests-nightly-ipv4-raftschema-rafttopology-tablets.yaml
@@ -12,7 +12,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  scylla-version: 'scylla-nightly:5.5.0-dev-0.20240321.900b56b117a1'
+  scylla-version: 'scylla-enterprise-nightly:latest-enterprise'
   ip-family: IPV4
   raft-schema: true
   raft-topology: true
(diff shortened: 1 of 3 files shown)


yaronkaikov:
@karol-kokoszka @Michal-Leszczynski I see that the GitHub action checks a specific release, any reason not to check the latest point release for each branch?  also, I don't see it running on 2024.1. is there a reason for that?

Maybe we should adjust the tests to run latest Scylla and latest Enterprise release, does it make sense?

Michal-Leszczynski:
@yaronkaikov how often are changes from OSS master moved to ENT? Is it often enough that we don't need to test SM against `scylla-nightly`, or is it done just before releasing ENT version?

yaronkaikov:
> @yaronkaikov how often are changes from OSS master moved to ENT? Is it often enough that we don't need to test SM against `scylla-nightly`, or is it done just before releasing ENT version?

OSS master is merged to Enterprise once a week.

Michal-Leszczynski:
So it looks like testing nightly on ENT only is good enough. I will work on test failures introduced by this change and will probably create a separate PR for them.

yaronkaikov:
@Michal-Leszczynski What do you want to do with this PR?

Michal-Leszczynski:
@yaronkaikov I will close it and assign myself to the fixed issue. Anyway, thanks for help!

PR #3866, opened 2024-05-24
Repair: use ranges parallelism

### LINKED ISSUE #3789: Coalesce token ranges sent in repair jobs
As mentioned in the https://github.com/scylladb/scylla-enterprise/issues/4006, it might be beneficial to always send more token ranges per job than `max_repair_ranges_in_parallel` (e.g. the greater of `10 * max_repair_ranges_in_parallel` and `5%-10%` of all ranges owned by repaired replica set). It allows for better concurrency (the whole job won't be blocked from returning when only 1 token range is still being repaired). Note that this should be safe to do with non-max intensity as well. It would require SM to use the `ranges_parallelism` repair param that limits used parallelism even when SM sends less ranges than `max_repair_ranges_in_parallel`.

This behavior could be controlled by an additional flag or repair config option in `scylla-manager.yaml`.
A side benefit would be to decrease the amount of clutter in SM logs.
In terms of testing, it would be good to see performance improvement on a cluster like: 2ds, 5 nodes each, keyspace with RF 3 in each dc, setup in which the repair indeed has to do some work (missing rows on some nodes).

### LINKED ISSUE #3792: Add separate parallel/intensity control for tablets
As per [tablet repair design](https://docs.google.com/document/d/1ATh5i4EMTJYnWERZ-BDH7lpgqwHe0HE-zxPWycW70So/edit#bookmark=id.ghsrwz1fdn4s), SM should introduce separate flags for controlling repair parallelism/intensity for tablet tables.
Those new flags should also be supported by the `sctool repair control` command.

As per mentioned doc, the default for `--tablet-intensity` should be the amount of shards (even despite `max_repair_ranges_in_parallel`). I'm not aware of any reasonable limit for this value, so I suppose that SM won't cap it at any point.

I'm wondering whether it makes sense to introduce `--tablet-parallel` flag (@tgrabiec please advise). If there is a need for this flag, then I believe that it should behave in the same way as `--parallel` flag. This means that it has the same default (1) and the same cap.

cc: @karol-kokoszka @tzach @tgrabiec @asias

Michal-Leszczynski:
This PR introduces token ranges batching in repair jobs.
Previously, with `intensity=X`, SM would simply send X token ranges per job. As described in #3789, this approach could lead to cluster being underutilized. Batching also solves #3792, as it increases the odds of repair job consisting of similar amount of token ranges owned by any shard. 

Fixes #3789
Fixes #3792

### DIFF
--- a/pkg/scyllaclient/client_scylla.go
+++ b/pkg/scyllaclient/client_scylla.go
@@ -553,7 +553,7 @@ func dumpRanges(ranges []TokenRange) string {
 			_ = buf.WriteByte(',')
 		}
 		if ttr.StartToken > ttr.EndToken {
-			_, _ = fmt.Fprintf(&buf, "%d:%d,%d:%d", dht.Murmur3MinToken, ttr.EndToken, ttr.StartToken, dht.Murmur3MaxToken)
+			_, _ = fmt.Fprintf(&buf, "%d:%d,%d:%d", ttr.StartToken, dht.Murmur3MaxToken, dht.Murmur3MinToken, ttr.EndToken)
 		} else {
 			_, _ = fmt.Fprintf(&buf, "%d:%d", ttr.StartToken, ttr.EndToken)
 		}
--- a/pkg/scyllaclient/client_scylla.go
+++ b/pkg/scyllaclient/client_scylla.go
@@ -520,7 +520,7 @@ func ReplicaHash(replicaSet []string) uint64 {
 }
 
 // Repair invokes async repair and returns the repair command ID.
-func (c *Client) Repair(ctx context.Context, keyspace, table, master string, replicaSet []string, ranges []TokenRange, smallTableOpt bool) (int32, error) {
+func (c *Client) Repair(ctx context.Context, keyspace, table, master string, replicaSet []string, ranges []TokenRange, intensity int, smallTableOpt bool) (int32, error) {
 	dr := dumpRanges(ranges)
 	p := operations.StorageServiceRepairAsyncByKeyspacePostParams{
 		Context:        forceHost(ctx, master),
@@ -530,6 +530,8 @@ func (c *Client) Repair(ctx context.Context, keyspace, table, master string, rep
 	}
 	if smallTableOpt {
 		p.SmallTableOptimization = pointer.StringPtr("true")
+	} else {
+		p.RangesParallelism = pointer.StringPtr(fmt.Sprint(intensity))
 	}
 	// Single node cluster repair fails with hosts param
 	if len(replicaSet) > 1 {
--- a/pkg/service/repair/intensity_handler.go
+++ b/pkg/service/repair/intensity_handler.go
@@ -1,0 +1,81 @@
+// Copyright (C) 2024 ScyllaDB
+
+package repair
+
+import (
+	"context"
+	"math"
+
+	"github.com/scylladb/go-log"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
+	"go.uber.org/atomic"
+)
+
+type sizeSetter interface {
+	SetSize(size int)
+}
+
+type intensityHandler struct {
+	taskID           uuid.UUID
+	runID            uuid.UUID
+	logger           log.Logger
+	maxHostIntensity map[string]Intensity
+	intensity        *atomic.Int64
+	maxParallel      int
+	parallel         *atomic.Int64
+	poolController   sizeSetter
+}
+
+const (
+	maxIntensity     Intensity = 0
+	defaultIntensity Intensity = 1
+	defaultParallel            = 0
+	chanSize                   = 10000
+)
+
+// SetIntensity sets the value of '--intensity' flag.
+func (i *intensityHandler) SetIntensity(ctx context.Context, intensity Intensity) {
+	i.logger.Info(ctx, "Setting repair intensity", "value", intensity, "previous", i.intensity.Load())
+	i.intensity.Store(int64(intensity))
+}
+
+// SetParallel sets the value of '--parallel' flag.
+func (i *intensityHandler) SetParallel(ctx context.Context, parallel int) {
+	i.logger.Info(ctx, "Setting repair parallel", "value", parallel, "previous", i.parallel.Load())
+	i.parallel.Store(int64(parallel))
+	if parallel == defaultParallel {
+		i.poolController.SetSize(i.maxParallel)
+	} else {
+		i.poolController.SetSize(parallel)
+	}
+}
+
+func (i *intensityHandler) ReplicaSetMaxIntensity(replicaSet []string) Intensity {
+	out := NewIntensity(math.MaxInt)
+	for _, rep := range replicaSet {
+		if ranges := i.maxHostIntensity[rep]; ranges < out {
+			out = ranges
+		}
+	}
+	return out
+}
+
+// MaxHostIntensity returns max_token_ranges_in_parallel per host.
+func (i *intensityHandler) MaxHostIntensity() map[string]Intensity {
+	return i.maxHostIntensity
+}
+
+// Intensity returns stored value for intensity.
+func (i *intensityHandler) Intensity() Intensity {
+	return NewIntensity(int(i.intensity.Load()))
+}
+
+// MaxParallel returns maximal achievable parallelism.
+func (i *intensityHandler) MaxParallel() int {
+	return i.maxParallel
+}
+
+// Parallel returns stored value for parallel.
+func (i *intensityHandler) Parallel() int {
+	return int(i.parallel.Load())
+}
--- a/pkg/service/repair/intensity_handler.go
+++ b/pkg/service/repair/intensity_handler.go
@@ -15,7 +15,7 @@ type sizeSetter interface {
 	SetSize(size int)
 }
 
-type intensityHandler struct {
+type intensityParallelHandler struct {
 	taskID           uuid.UUID
 	runID            uuid.UUID
 	logger           log.Logger
@@ -33,14 +33,17 @@ const (
 	chanSize                   = 10000
 )
 
-// SetIntensity sets the value of '--intensity' flag.
-func (i *intensityHandler) SetIntensity(ctx context.Context, intensity Intensity) {
+// SetIntensity sets the effective '--intensity' value used by newly created repair jobs.
+// It can be used to control currently running repair without stopping it and changing '--intensity' flag.
+// The change is ephemeral, so it does not change the value of '--intensity' flag in SM DB, meaning that without
+// additional actions, it won't be visible in next task runs.
+func (i *intensityParallelHandler) SetIntensity(ctx context.Context, intensity Intensity) {
 	i.logger.Info(ctx, "Setting repair intensity", "value", intensity, "previous", i.intensity.Load())
 	i.intensity.Store(int64(intensity))
 }
 
-// SetParallel sets the value of '--parallel' flag.
-func (i *intensityHandler) SetParallel(ctx context.Context, parallel int) {
+// SetParallel works in the same way as SetIntensity, but for '--parallel' value.
+func (i *intensityParallelHandler) SetParallel(ctx context.Context, parallel int) {
 	i.logger.Info(ctx, "Setting repair parallel", "value", parallel, "previous", i.parallel.Load())
 	i.parallel.Store(int64(parallel))
 	if parallel == defaultParallel {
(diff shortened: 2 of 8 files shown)


Michal-Leszczynski:

karol-kokoszka:
Let's execute repair SCT to see the impact https://github.com/scylladb/scylla-manager/issues/3867

pkg/service/repair/intensity_handler.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/scylladb/go-log"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
+	"go.uber.org/atomic"
+)
+
+type sizeSetter interface {
+	SetSize(size int)
+}
+
+type intensityHandler struct {
```
karol-kokoszka, line 18: nit: it's more intensityAndParallelHandler

pkg/service/repair/intensity_handler.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+const (
+	maxIntensity     Intensity = 0
+	defaultIntensity Intensity = 1
+	defaultParallel            = 0
+	chanSize                   = 10000
+)
+
+// SetIntensity sets the value of '--intensity' flag.
```
karol-kokoszka, line 36: nit: I think it doesn't set the value of `--intensity` flag, but rather updates ongoing repairs with given value for intensity (link https://manager.docs.scylladb.com/stable/sctool/repair.html#intensity)

pkg/service/repair/intensity_handler.go
```diff
@@ -0,0 +1,81 @@
...
+func (i *intensityHandler) SetIntensity(ctx context.Context, intensity Intensity) {
+	i.logger.Info(ctx, "Setting repair intensity", "value", intensity, "previous", i.intensity.Load())
+	i.intensity.Store(int64(intensity))
+}
+
+// SetParallel sets the value of '--parallel' flag.
```
karol-kokoszka, line 42: nit: same as for intensity, updates ongoing repairs with new value defining parallel https://manager.docs.scylladb.com/stable/sctool/repair.html#parallel

pkg/service/repair/intensity_handler.go
```diff
@@ -0,0 +1,81 @@
...
+	i.logger.Info(ctx, "Setting repair parallel", "value", parallel, "previous", i.parallel.Load())
+	i.parallel.Store(int64(parallel))
+	if parallel == defaultParallel {
+		i.poolController.SetSize(i.maxParallel)
+	} else {
+		i.poolController.SetSize(parallel)
+	}
+}
+
+func (i *intensityHandler) ReplicaSetMaxIntensity(replicaSet []string) Intensity {
```
karol-kokoszka, line 53: Public method without the comment. (Although linters says it's ok -> it must be due to the fact that intensityHandler is unexported)

pkg/service/repair/service.go
```diff
@@ -454,71 +462,73 @@ func (s *Service) SetParallel(ctx context.Context, clusterID uuid.UUID, parallel
...
+			"type":       "repair",
+			"task_id":    taskID,
+			"id":         id,
+		}).Scan(&status)
+		if err != nil {
+			return false, errors.Wrap(err, "get prev run status")
+		}
+		// Fall back to no-batching when some of the previous runs:
+		// - finished with error
+		// - got out of scheduler window
```
karol-kokoszka, line 497: Can you elaborate on the second statement ? 
What do you mean by "got out of scheduler window" ? Is it connected to maintenance window ? If so, then why is it the indicator to not batch ?

pkg/service/repair/service_repair_integration_test.go
```diff
@@ -432,6 +435,19 @@ func createKeyspace(t *testing.T, session gocqlx.Session, keyspace string, rf1,
...
 
+func tryCreateTabletKeyspace(t *testing.T, session gocqlx.Session, keyspace string, rf1, rf2, tablets int) {
+	createKeyspaceStmt := "CREATE KEYSPACE %s WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': %d, 'dc2': %d}"
+	tabletStmt := " AND tablets = {'enabled': true, 'initial': %d}"
+	if tablets > 0 {
+		err := session.ExecStmt(fmt.Sprintf(createKeyspaceStmt+tabletStmt, keyspace, rf1, rf2, tablets))
+		if err == nil {
+			return
+		}
+		// Fallback as we don't know if tablets are enabled
```
karol-kokoszka, line 446: I'm not sure if I understand the fallback.
What happens with the executed statement above if tablets are not enabled ? Will the ExecStmt fail and return error ? 
If it's true, then you return from the method.

Or the next ExecStmt (without tablets part) will fail if the tablets are enabled ? But if the latter will fail, then the test will fail.

pkg/scyllaclient/client_scylla.go
```diff
@@ -551,7 +553,7 @@ func dumpRanges(ranges []TokenRange) string {
 			_ = buf.WriteByte(',')
 		}
 		if ttr.StartToken > ttr.EndToken {
-			_, _ = fmt.Fprintf(&buf, "%d:%d,%d:%d", dht.Murmur3MinToken, ttr.EndToken, ttr.StartToken, dht.Murmur3MaxToken)
+			_, _ = fmt.Fprintf(&buf, "%d:%d,%d:%d", ttr.StartToken, dht.Murmur3MaxToken, dht.Murmur3MinToken, ttr.EndToken)
```
karol-kokoszka, line 556: Why the order is changed here ? What is the meaning of this change ?

karol-kokoszka (commented):
few nits, let's wait for comparison results from SCT (batching and no batching with VNodes)

pkg/scyllaclient/client_scylla.go
Michal-Leszczynski: As described in 6624e060:
```
refactor(scyllaclient): make split range easier to detect

It's easier to look for a substring than regex match when checking if given token range string was created by splitting overlapping range.
```
This change just makes it easier to detect when a split token range is included in the repair job - which is important when writing tests validating that SM sends correct amount of token ranges per job.

pkg/service/repair/service_repair_integration_test.go
Michal-Leszczynski: > What happens with the executed statement above if tablets are not enabled ? Will the ExecStmt fail and return error ?

yes.

>  If it's true, then you return from the method.

yes, because method managed to create tablet keyspace.

> Or the next ExecStmt (without tablets part) will fail if the tablets are enabled ? But if the latter will fail, then the test will fail.

The second statement is correct for both vnode and tablets - it basically creates a keyspace with default type of replication. So if it fails, SM failed at creating regular keyspace and the test cannot proceed. 



In general it was annoying to check if tablets are enabled as it required to query experimental.tablets field from Scylla config. In Syclla 6.0 it was moved from experimental to regular fields. So it was annoying to check for that, and that's why I decided to use this simple approach for testing.

pkg/service/repair/service.go
Michal-Leszczynski: As described in https://github.com/scylladb/scylla-manager/issues/3792#issuecomment-2112151831, I was worried that batching combined with a maintenance window set to many, small time windows (e.g. repair can run for between `X:45-X:59` where X can be any hour of any day), could result in repair not making any progress as repair would be interrupted by going out of maintenance window before finishing any repair job. This could be especially true for tables with only one replica set (e.g. fully replicated table), where batching could send 10% of all ranges (so usually` #node_cnt * 256 / 10` token ranges) in a single repair job.

Having said that, after playing with SCT performance tests, I plan to change batching from `max(intensity, 10% of replica set ranges)` to `min(100% of replica set ranges, 10 * intensity)`, so the maintenance window argument will be weaker.

@karol-kokoszka do you think that we shouldn't worry about maintenance window and SM shouldn't fallback to non-batching approach when repair runs are interrupted by it?

Michal-Leszczynski: On the other hand, I think I will change batching to `min(100% of replica set ranges, 1000)`.

Sending all ranges owned by replica set is what asias suggested initially and this number might be big only for fully replicated tables, which won't happen in bigger (more than 9 nodes) clusters. Hard limit of 1000 makes more sense than 10 * intensity when running with low intensity setting (batching helps with non max intensity as well).
In real life, this limit won't really be hit anyway.

karol-kokoszka: Ok, let's leave the fallback on maintenance window.

pkg/service/repair/service_repair_integration_test.go
karol-kokoszka: fine

karol-kokoszka (approved):
Please update https://docs.google.com/document/d/1mBaufbomXU6hDO_25KkhC7CO65AbFgKpHVpB5Mei8xA/edit

..and attach SCT run results on VNodes (and on tablets if possible)

Michal-Leszczynski:
SCT setup:
- 5 `i4i.2xlarge` node cluster
- inserted total of `292968720` rows for vnode scenarios and **twice** as much for tablet scenarios
- inserted rows while one of the nodes were down
-  stress_command_template: 
``` "cassandra-stress write cl=QUORUM n={} -schema 'keyspace={}" \
                                  " replication(strategy=NetworkTopologyStrategy,replication_factor=3)'" \
                                  " -col 'size=FIXED(1024) n=FIXED(1)' -pop seq={}..{} -mode cql3" \
                                  " native -rate threads=200 -log interval=5"
```
SCT results:
```
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/michal-leszczynski/job/repair-intensity-multiple-repaired-nodes-test/23/
master & tablets: 18:57:27 -> 19:34:25 ~ 36 min 58 sec

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/michal-leszczynski/job/repair-intensity-multiple-repaired-nodes-test/22/
master & vnode: 04:28:19 -> 04:36:35 ~ 8 min 16 sec

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/michal-leszczynski/job/repair-intensity-multiple-repaired-nodes-test/24/
batching & tablets: 00:43:36 -> 01:02:21 ~ 18 min 45 sec

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/michal-leszczynski/job/repair-intensity-multiple-repaired-nodes-test/25/
batching & vnode: 02:45:18 -> 02:53:07 ~ 7 min 49 sec
```

The problem here is that I could only set the amount of rows, not data, inserted to the cluster. If I understand it correctly, the load for vnode was around 280GB and around 560GB for tablets, but someone would need to verify my C-S math.
I first run the tablet scenarios and observed, that the SCT runs for about 4h, so that's why I decided to decrease the load so that it can finish faster.

@karol-kokoszka is this enough to merge this PR?

karol-kokoszka:
> @karol-kokoszka is this enough to merge this PR?
 
Yes it is. 
It's a bit strange that repair for tablets lasts longe than for vnodes though. 

Please merge. Thanks !

mykaul:
> SCT setup:
> 
> * 5 `i4i.2xlarge` node cluster
> * inserted total of `292968720` rows for vnode scenarios and **twice** as much for tablet scenarios
> * inserted rows while one of the nodes were down
> * stress_command_template:
> 
> ```
>                                   " replication(strategy=NetworkTopologyStrategy,replication_factor=3)'" \
>                                   " -col 'size=FIXED(1024) n=FIXED(1)' -pop seq={}..{} -mode cql3" \
>                                   " native -rate threads=200 -log interval=5"
> ```
> 
> SCT results:
> 
> ```
> https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/michal-leszczynski/job/repair-intensity-multiple-repaired-nodes-test/23/
> master & tablets: 18:57:27 -> 19:34:25 ~ 36 min 58 sec
> 
> https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/michal-leszczynski/job/repair-intensity-multiple-repaired-nodes-test/22/
> master & vnode: 04:28:19 -> 04:36:35 ~ 8 min 16 sec
> 
> https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/michal-leszczynski/job/repair-intensity-multiple-repaired-nodes-test/24/
> batching & tablets: 00:43:36 -> 01:02:21 ~ 18 min 45 sec
> 
> https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/michal-leszczynski/job/repair-intensity-multiple-repaired-nodes-test/25/
> batching & vnode: 02:45:18 -> 02:53:07 ~ 7 min 49 sec
> ```
> 
> The problem here is that I could only set the amount of rows, not data, inserted to the cluster. If I understand it correctly, the load for vnode was around 280GB and around 560GB for tablets, but someone would need to verify my C-S math. I first run the tablet scenarios and observed, that the SCT runs for about 4h, so that's why I decided to decrease the load so that it can finish faster.
> 
> @karol-kokoszka is this enough to merge this PR?

@asias - thoughts on the above?

Michal-Leszczynski:
> inserted total of 292968720 rows for vnode scenarios and twice as much for tablet scenarios

Also, this setup initially resulted in creating only 64 tablets. This is way too little for testing batching (too little token ranges owned by each replica set), so I created the tablet keyspace with 8192 initial tablets.

PR #3876, opened 2024-06-06
Restore schema with internals

### LINKED ISSUE #3662: Fix restore schema procedure

### LINKED ISSUE #3868: Use DESC SCHEMA for backup in Scylla 6.0
Context:
As the workaround described [here](https://manager.docs.scylladb.com/stable/restore/restore-schema.html#restore-schema-workaround) won't work with Scylla 6.0, we need to implement a new way for schema backup/restore.
It was decided to move back to the DESCRIBE SCHEMA WITH INTERNALS approach after it was changed so that it works well with restore procedure (https://github.com/scylladb/scylladb/issues/16482).

Previous approach:
As SM does not always have CQL credentials to managed cluster, the first approach was to use `maintenance_socket` from agent (https://github.com/scylladb/scylla-manager/issues/3831). In the end it was aborted for Scylla 6.0, as it wouldn't fit the release deadline (perhaps it will be done as a part of integration with Scylla 6.1).

Chosen approach:
It was [decided](https://github.com/scylladb/scylladb/issues/18668#issuecomment-2132151808) that SM should have cluster CQL credentials when doing a backup. This is a regression when compared with the current backup which does not need the credentials to backup the schema, but this approach is better for the Scylla 6.0 release deadline.

Implementation details:
DESCRIBE SCHEMA WITH INTERNALS returns 4 columns: object keyspace, object type, object name, CQL stmt.
SM should just save the whole output in a file on backup. It's important to preserve the order of rows (https://github.com/scylladb/scylladb/issues/18539) and not to loose the granularity of mentioned columns (even though right now we only perform full cluster schema restore, there is a plan to support it with nicer granularity in the future). The restore would simply download mentioned file and apply all the CQL stmts in order - no cluster restart needed.

Michal-Leszczynski:
This PR makes backup/restore procedure use DESCRIBE SCHEMA WITH INTERNALS.

Backup:
The `schema.tar.gz` (describe schema from driver method) is no longer backed up, as it was never used anyway. Instead of that, new `schema_with_internals.json.gz` (describe schema from cql query) is backed up and (when present) used for restore.
This requires cluster credentials, so schema is backed up only when the credentials are set. If credentials are not there, backup proceed without DESCRIBE SCHEMA output. As a safety measure, `system_schema` is still always backed up so that it could be used to restore schema when DESCRIBE SCHEMA output is missing from the backup (this would require restoring `system_schema` into a single node cluster and then describing schema on it).

Restore schema:
It always tries to use `schema_with_internals.json.gz` if present. Otherwise it checks Scylla in order to validate restore schema from sstables safety.

Fixes #3868
Fixes #3662


pkg/service/backup/worker_schema.go
```diff
@@ -100,3 +125,51 @@ func (w *worker) UploadSchema(ctx context.Context, hosts []hostInfo) (stepError
...
+
+// isDescribeSchemaSafe checks if restoring schema from DESCRIBE SCHEMA WITH INTERNALS is safe.
+func isDescribeSchemaSafe(ctx context.Context, client *scyllaclient.Client, hosts []string) (bool, error) {
+	const (
+		ConstraintOSS = ">= 6.0, < 2000"
+		ConstraintENT = ">= 2024.2, > 1000"
+	)
+
+	for _, h := range hosts {
+		ni, err := client.NodeInfo(ctx, h)
```
karol-kokoszka, line 155: NodeInfo on agent is not a single call to Scylla's API. The result is the concatenation of many different calls made on scylla server by manager-agent (see `pkg/cmd/agent/nodeinfo.go`).
It can be replaced with direct call to ScyllaAPI (through agent's proxy) -> "/storage_service/scylla_release_version"

pkg/service/backup/worker_manifest.go
```diff
@@ -71,8 +71,8 @@ func (w *worker) createTemporaryManifest(h hostInfo, tokens []int64) ManifestInf
 		},
 		Index: make([]FilesMeta, len(dirs)),
 	}
-	if w.Schema != nil {
-		c.Schema = RemoteSchemaFile(w.ClusterID, w.TaskID, w.SnapshotTag)
+	if w.Schema != nil && w.SchemaFilePath != "" {
```
karol-kokoszka, line 74: Is this added check necessary ?
`c.Schema` is pre-initalized somehow and cannot be updated to "" ?

pkg/service/restore/worker.go
```diff
@@ -224,6 +231,45 @@ func IsRestoreSchemaSupported(ctx context.Context, client *scyllaclient.Client)
...
+		ossConstraint = ">= 6.0, < 2000"
+		entConstraint = ">= 2024.2, > 1000"
+	)
+
+	status, err := client.Status(ctx)
+	if err != nil {
+		return errors.Wrap(err, "get status")
+	}
+	for _, n := range status {
+		ni, err := client.NodeInfo(ctx, n.Addr)
```
karol-kokoszka, line 263: NodeInfo is a concatenation of different API calls to Scylla. Can be simplified with direct call for scylla_version.

pkg/service/restore/worker.go
```diff
@@ -224,6 +231,45 @@ func IsRestoreSchemaSupported(ctx context.Context, client *scyllaclient.Client)
...
+		oss, err := version.CheckConstraint(ni.ScyllaVersion, ossConstraint)
+		if err != nil {
+			return errors.Wrapf(err, "check version constraint for %s", n.Addr)
+		}
+		ent, err := version.CheckConstraint(ni.ScyllaVersion, entConstraint)
+		if err != nil {
+			return errors.Wrapf(err, "check version constraint for %s", n.Addr)
+		}
+
+		if oss || ent {
```
karol-kokoszka, line 277: oss and ent could be renamed to something more meaningful.
I.e. ossNotSupported, entNotSupported

pkg/service/restore/schema_worker.go
```diff
@@ -123,6 +132,47 @@ func (w *schemaWorker) stageRestoreData(ctx context.Context) error {
...
+	for _, row := range *w.describedSchema {
+		if row.Keyspace == "system_replicated_keys" {
+			// See https://github.com/scylladb/scylla-enterprise/issues/4168
+			continue
+		}
+		// Sometimes a single object might require multiple CQL statements (e.g. table with dropped and added column)
+		for _, stmt := range parseCQLStatement(row.CQLStmt) {
+			if err := w.clusterSession.ExecStmt(stmt); err != nil {
+				return errors.Wrapf(err, "create %s (%s) with %s", row.Name, row.Keyspace, stmt)
+			}
```
karol-kokoszka, line 147: I think that it may lead to inconsistent behavior when some of the statements succeeds, but it fail on the next one.
First run will end on first incorrect statement, the second one will error on the first statement as the table may exist already.
WDYT about dropping created keyspaces on failed run ? Kind of rollback.

pkg/service/restore/schema_worker.go
```diff
@@ -273,3 +323,93 @@ func (w *schemaWorker) getFileNamesMapping(sstables []string, sstableUUIDFormat
 	}
 	return sstable.RenameToIDs(sstables, &w.generationCnt)
 }
+
+func getDescribedSchema(ctx context.Context, client *scyllaclient.Client, snapshotTag string, locHost map[Location][]string) (*query.DescribedSchema, error) {
+	baseDir := path.Join("backup", string(SchemaDirKind))
+	// It's enough to get a single schema file, but it's important to validate
+	// that each location contains exactly one or none of them.
```
karol-kokoszka, line 330: Should they be compared against the equal content or hash ? Is the consistency guaranteed by awaiting schema agreement ?

pkg/service/restore/schema_worker.go
```diff
@@ -273,3 +323,93 @@ func (w *schemaWorker) getFileNamesMapping(sstables []string, sstableUUIDFormat
...
+		}
+	}
+
+	if foundCnt == 0 {
+		return nil, nil // nolint: nilnil
+	} else if foundCnt < len(locHost) {
+		return nil, errors.New("only a subset of provided locations has schema files")
+	}
+
+	r, err := client.RcloneOpen(ctx, host, *schemaPath)
```
karol-kokoszka, line 354: I think it's caller responsibility to close the Reader here.
```
// RcloneOpen streams remote file content. The stream is an HTTP body.
// Callers must close the body after use.
```
Please defer the r.Close and log error if appears.
Closing twice should not cause errors.

pkg/service/restore/worker.go
```diff
@@ -166,19 +161,36 @@ func (w *worker) initTarget(ctx context.Context, properties json.RawMessage) err
...
 
+	if t.RestoreSchema {
+		w.logger.Info(ctx, "Look for schema file")
+		w.describedSchema, err = getDescribedSchema(ctx, w.client, t.SnapshotTag, locationHosts)
+		if err != nil {
+			return errors.Wrap(err, "look for schema file")
+		}
+
+		if w.describedSchema == nil {
+			w.logger.Info(ctx, "Couldn't find schema file")
```
karol-kokoszka, line 173: `Couldn't find schema file` -> `Couldn't find schema file. Proceeding with schema restoration using sstables.` ?

pkg/service/restore/service_restore_integration_test.go
```diff
@@ -300,21 +301,45 @@ func TestRestoreGetTargetUnitsViewsIntegration(t *testing.T) {
...
 			}
 
+			var ignoreUnits []string
+			if checkAnyConstraint(t, h.Client, "< 1000") {
+				ignoreUnits = append(ignoreUnits, "system_replicated_keys")
+			}
+			if checkAnyConstraint(t, h.Client, "< 2024.2") {
+				ignoreUnits = append(ignoreUnits, "dicts")
+			}
+			if checkAnyConstraint(t, h.Client, ">= 5.5, < 2000", ">= 2024.2, > 1000") {
```
karol-kokoszka, line 328: As per my knowledge, 5.5 never went out (please correct me if I'm wrong). It can be >= 6.0

karol-kokoszka (commented):
Few comments. Overall :+1:

pkg/service/restore/service_restore_integration_test.go
Michal-Leszczynski: I never did, but scylla-nightly had version 5.5 for some time and this test constraint was written so that it works fine with that. I will change it to 6.0 and see how the tests behave.

pkg/service/restore/schema_worker.go
Michal-Leszczynski: Nice catch!

pkg/service/restore/schema_worker.go
Michal-Leszczynski: Too be honest, this check could be removed.

This check is about verifying if user specified correct locations - they either all have schema file with given snapshot tag, or none of them have it. It could be extended to checking file hash, but IMO this would be an overkill (perhaps this check is already a one). 

In general, when performing backup, we ask only a single node for the DESC SCHEMA and we backup it to all specified backup locations. This means, that we don't store per node DESC SCHEMA output, as it would be difficult to use it during restore. It also means, that all locations have the same file content.

In terms of schema agreement, it turns out that we don't even need to have it when querying DESC SCHEMA (see https://github.com/scylladb/scylladb/issues/19213 for more details), but I have yet to change that in another commit.

Michal-Leszczynski: So I'm fine with removing this check or just leaving it as it is. What is your opinion on this matter?

pkg/service/restore/schema_worker.go
Michal-Leszczynski: We can do that!
First I wanted to say that it would require us to parse CQL statements (bad idea), so that we would know what should be dropped, but thanks to the new schema file format, we save information about all keyspace names present in restored schema, so it makes it simple and safe to do!

pkg/service/backup/worker_schema.go
```diff
@@ -100,3 +125,51 @@ func (w *worker) UploadSchema(ctx context.Context, hosts []hostInfo) (stepError
...
 	return hostsInParallel(hostPerLocation, parallel.NoLimit, f, notify)
 }
+
+func marshalAndCompressSchema(schema query.DescribedSchema) (*bytes.Buffer, error) {
+	rawSchema, err := json.Marshal(schema)
+	if err != nil {
+		return nil, errors.Wrap(err, "marshal schema")
+	}
+
+	b := &bytes.Buffer{}
```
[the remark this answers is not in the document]
Michal-Leszczynski: I think that it depends on the context.
Here, we return `*bytes.Buffer{}` from this function, so it can't be initialized on this function stack, as it needs to be available from outside of this function body. So I doubt that changing it to `var b bytes.Buffer` would accomplish anything (unless  I'm missing something).

This could work if both reading and uploading schema happened in the same function body. This could be done by combining the `StageAwaitSchema` and `StageSchema` stages - I don't see why they need to be separate. Nevertheless, this seems more like a refactoring task that is outside of the scope of this PR and can be included in another one.

pkg/service/backup/worker_manifest.go
Michal-Leszczynski: It shouldn't be, those two values should be always both initialized of uninitialized.
On the other hand, this is some form of another safety measure that doesn't harm anything, so I would be fine with leaving it as it is.

pkg/service/restore/worker.go
Michal-Leszczynski: That's true. A similar situation happens in the `IsRestoreSchemaFromSSTablesSupported` method, but there we also need the `consistent_cluster_management` config field.

What would you say if we made `configcache` svc reachable from `restore` svc (the same as in `healthcheck` svc) and use `nodeInfo` already fetched by `configcache` svc instead of doing it on our own? This would also help in other places that fetch `nodeInfo`.

I could do it as a part of a refactoring PR, but I would leave this as it is in the current one.

pkg/service/backup/worker_schema.go
Michal-Leszczynski: The same idea as for restore described in: https://github.com/scylladb/scylla-manager/pull/3876#discussion_r1641978810.

Michal-Leszczynski:
@karol-kokoszka I addressed the comments (either by new commits or responses with questions).

zimnx:
We (Operator team) would appreciate prioritizing this PR and preparing a release right after it's merged. This blocks our adoption of ScyllaDB 6.0.0, as our Manager rejects restore attempt when 6.0.0 is used. Lack of support for it breaks our e2e tests.

pkg/service/restore/worker.go
karol-kokoszka: I would prefer to not spread the configcache among all the services yet.
We must be sure that everything is fine with it and it's battle-tested for some time.

But, it makes sense to leave it as is for the moment and refactor in future.

pkg/service/restore/schema_worker.go
karol-kokoszka: I'm fine with leaving it as is.

pkg/service/backup/worker_schema.go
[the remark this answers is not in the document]
karol-kokoszka: > Here, we return *bytes.Buffer{} from this function, so it can't be initialized on this function stack, as it needs to be available from outside of this function body. So I doubt that changing it to var b bytes.Buffer would accomplish anything (unless I'm missing something).

Makes sense, but please be explicit with the type that the function returns.
Use bytes.Buffer as a return type instead of the pointer.

karol-kokoszka:
@Michal-Leszczynski Sanity-check is failing (unit-test).

Please check/fix.

karol-kokoszka (approved):
:+1:

PR #3880, opened 2024-06-11
healthcheck: runner to return error

karol-kokoszka:
Return error from healtcheck runner

---

Please make sure that:


Michal-Leszczynski:
I understand that this change is fixing the logging from `run.Parallel`. While we are at it, could we also change:
```go
_ = parallel.Run(len(addresses), parallel.NoLimit, f, func(i int, err error) { // nolint: errcheck
		r.logger.Error(ctx, "Parallel hosts check failed", "", addresses[i], "error", err)
	})
```
so that we don't include an empty string in logging? It should be something like "host" instead.

karol-kokoszka:
updated

Michal-Leszczynski (approved):

PR #3883, opened 2024-06-12
testing: enable sstable uuid identifiers in Scylla config

karol-kokoszka:
This PR enables UUID in SSTables names feature in Scylla Cluster against which all tests are being executed.

---

Please make sure that:


Michal-Leszczynski:
It won't hurt to specify it manually, but I believe that sstable UUID identifiers are enabled by default.

karol-kokoszka:
Actually you're right :)
There is no need to set it explicitly. 
5.4.1 (the oldest we use to execute tests against) keeps it as a default.

Issue #3827, opened 2024-04-25
Add explicit deduplicate stage to SM (remove basing on .crc32 or UUID as generation ID)


karol-kokoszka:
Scylla Enterprise 2024.1 introduced as a default support for UUIDs in SSTable names.
https://github.com/scylladb/scylladb/pull/13932

Scylla manager no longer needs to compare checksums when copying files from local to storage, as the SSTable uniqueness is guaranteed by the UUID and the situation where SSTable have the same name and size, but different content is mitigated.

The above is valid for backups made on >= 2024.1 only.
We still need to support old approach for older Scyllas.
 
Issue on QA side https://github.com/scylladb/qa-tasks/issues/1397

More details of why we have to add this support is here https://github.com/scylladb/scylla-enterprise/issues/4126#issuecomment-2075049181

karol-kokoszka:
Update:

SSTable specification includes .crt32 file which is actually the checksum created by scylla server.
It's enough to compare the content of .crt32 inlcuded into created snapshot for every SSTable to the content of its remote equivalent.
If they are equal, then just remove the SSTable from snapshot dir.

karol-kokoszka:
We need to test the scenario described in the previous comment against the snapshot that contains hundreds of SSTables to see the performance of this operation.
It includes reading content of every crc32 file that is part of the current snapshot.

karol-kokoszka:
RFC for the new deduplication stage:
https://docs.google.com/document/d/1EtGlF6UGNy34D_7QsnCheaukp3UwVObZU56PBdd0CQ8/edit?usp=sharing

PR #3885, opened 2024-06-18
Cheksum comparison

karol-kokoszka:
Fixes https://github.com/scylladb/scylla-manager/issues/3827

This PR adds explicit stage to the backup process - deduplication.
It follows https://docs.google.com/document/d/1EtGlF6UGNy34D_7QsnCheaukp3UwVObZU56PBdd0CQ8/edit#heading=h.jl2qbpcarwp9

It is done to mitigate the problem with current deduplication/checksum comparison which causes SM to create a checksum of every SSTable file and to compare it with the checksum from remote.

This PR changes restore of versioned snapshot as well to corrupt CRC32 files.

---

Please make sure that:

### DIFF
--- a/pkg/service/backup/worker_deduplicate.go
+++ b/pkg/service/backup/worker_deduplicate.go
@@ -5,6 +5,7 @@ package backup
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"path"
 	"slices"
 	"strconv"
@@ -75,13 +76,17 @@ func (w *worker) deduplicateHost(ctx context.Context, h hostInfo) error {
 			ssTablesGroupByID[id] = append(ssTablesGroupByID[id], file.Name)
 		}
 
-		if err := w.basedOnUUIDGenerationAvailability(ctx, d, h, remoteFilesSSTableIDSet, ssTablesGroupByID); err != nil {
+		deduplicatedByUUID, err := w.basedOnUUIDGenerationAvailability(ctx, d, h, remoteFilesSSTableIDSet, ssTablesGroupByID)
+		if err != nil {
 			return errors.Wrap(err, "deduplication based on UUID as generation id content")
 		}
+		d.Progress.Skipped += deduplicatedByUUID
 
-		if err := w.basedOnCrc32Content(ctx, d, h, dataDst, remoteFilesSSTableIDSet, ssTablesGroupByID); err != nil {
+		deduplicatedByCrc32, err := w.basedOnCrc32Content(ctx, d, h, dataDst, remoteFilesSSTableIDSet, ssTablesGroupByID)
+		if err != nil {
 			return errors.Wrap(err, "deduplication based on .crc32 content")
 		}
+		d.Progress.Skipped += deduplicatedByCrc32
 
 		return nil
 	}
@@ -101,7 +106,7 @@ func (w *worker) deduplicateHost(ctx context.Context, h hostInfo) error {
 
 func (w *worker) basedOnUUIDGenerationAvailability(ctx context.Context, d snapshotDir, h hostInfo,
 	remoteSSTables map[string]struct{}, ssTablesGroupByID map[string][]string,
-) error {
+) (deduplicated int64, err error) {
 	// Per every SSTable files group, check if the name uses UUID to identify the SSTable generation.
 	// If the above is true, then check if files exists in the remote storage and remove it locally if exists in rmeote.
 	for id, ssTableContent := range ssTablesGroupByID {
@@ -120,17 +125,24 @@ func (w *worker) basedOnUUIDGenerationAvailability(ctx context.Context, d snapsh
 			w.Logger.Debug(ctx, "Removing local snapshot file (deduplication based on generation UUID)",
 				"host", h.IP, "file", localSSTableFileNameWithPath)
 			if err := w.Client.RcloneDeleteFile(ctx, h.IP, localSSTableFileNameWithPath); err != nil {
-				return errors.Wrapf(err, "cannot delete local snapshot's SSTable file %s", localSSTableFileNameWithPath)
+				return deduplicated, errors.Wrapf(err, "cannot delete local snapshot's SSTable file %s", localSSTableFileNameWithPath)
 			}
+			idx := slices.IndexFunc(d.Progress.files, func(info fileInfo) bool {
+				return info.Name == file
+			})
+			if idx == -1 {
+				return deduplicated, fmt.Errorf("no file %s in indexed files", file)
+			}
+			deduplicated += d.Progress.files[idx].Size
 		}
 	}
 
-	return nil
+	return deduplicated, nil
 }
 
 func (w *worker) basedOnCrc32Content(ctx context.Context, d snapshotDir, h hostInfo, dataDst string,
 	remoteSSTables map[string]struct{}, ssTablesGroupByID map[string][]string,
-) error {
+) (deduplicated int64, err error) {
 	// Reference to SSTables 3.0 Data File Format
 	// https://opensource.docs.scylladb.com/stable/architecture/sstable/sstable3/sstables-3-data-file-format.html
 
@@ -151,30 +163,54 @@ func (w *worker) basedOnCrc32Content(ctx context.Context, d snapshotDir, h hostI
 		remoteCRC32FileNameWithPath := path.Join(dataDst, ssTableContent[crc32Idx])
 		remoteCRC32, err := w.Client.RcloneCat(ctx, h.IP, remoteCRC32FileNameWithPath)
 		if err != nil {
-			if strings.Contains(err.Error(), "object not found") {
+			if scyllaclient.StatusCodeOf(err) == 404 {
 				continue
 			}
-			return errors.Wrapf(err, "cannot get content of remote CRC32 %s", remoteCRC32FileNameWithPath)
+			return deduplicated, errors.Wrapf(err, "cannot get content of remote CRC32 %s", remoteCRC32FileNameWithPath)
 		}
 
 		localCRC32FileNameWithPath := path.Join(d.Path, ssTableContent[crc32Idx])
 		localCRC32, err := w.Client.RcloneCat(ctx, h.IP, localCRC32FileNameWithPath)
 		if err != nil {
-			return errors.Wrapf(err, "cannot get content of local CRC32 %s", localCRC32FileNameWithPath)
+			return deduplicated, errors.Wrapf(err, "cannot get content of local CRC32 %s", localCRC32FileNameWithPath)
 		}
 
 		// If checksums are equal, then it means that SSTable can be deduplicated as there is no need in
 		// transferring it again to the remote storage.
 		// Deduplication here means to remove SSTable files from local storage.
 		if bytes.Equal(localCRC32, remoteCRC32) {
+			// validate sizes first
+			for _, file := range ssTableContent {
+				remoteFilePath := path.Join(dataDst, file)
+				remoteInfo, err := w.Client.RcloneFileInfo(ctx, h.IP, remoteFilePath)
+				if err != nil {
+					return deduplicated, errors.Wrapf(err, "get file info %s", remoteFilePath)
+				}
+				idx := slices.IndexFunc(d.Progress.files, func(info fileInfo) bool {
+					return info.Name == file
+				})
+				if idx == -1 {
+					return deduplicated, fmt.Errorf("no file %s in indexed files", file)
+				}
+				if remoteInfo.Size != d.Progress.files[idx].Size {
+					w.Logger.Debug(ctx, "Equal CRC32 hash but different sizes, skipping", "remote", remoteCRC32FileNameWithPath)
+				}
+			}
 			for _, fileToBeRemoved := range ssTableContent {
 				localSSTableFileNameWithPath := path.Join(d.Path, fileToBeRemoved)
 				w.Logger.Debug(ctx, "Removing local snapshot file (deduplication based on .crc32)", "host", h.IP, "file", localSSTableFileNameWithPath)
 				if err := w.Client.RcloneDeleteFile(ctx, h.IP, localSSTableFileNameWithPath); err != nil {
-					return errors.Wrapf(err, "cannot delete local snapshot's SSTable file %s", localSSTableFileNameWithPath)
+					return deduplicated, errors.Wrapf(err, "cannot delete local snapshot's SSTable file %s", localSSTableFileNameWithPath)
 				}
+				idx := slices.IndexFunc(d.Progress.files, func(info fileInfo) bool {
+					return info.Name == fileToBeRemoved
+				})
+				if idx == -1 {
+					return deduplicated, fmt.Errorf("no file %s in indexed files", fileToBeRemoved)
+				}
+				deduplicated += d.Progress.files[idx].Size
 			}
 		}
 	}
-	return nil
+	return deduplicated, nil
 }
--- a/pkg/service/backup/worker_deduplicate.go
+++ b/pkg/service/backup/worker_deduplicate.go
@@ -17,6 +17,11 @@ import (
 	"github.com/scylladb/scylla-manager/v3/pkg/util/parallel"
 )
 
+type deduplicateTestHooks interface {
+	beforeDeduplicateHost()
+	afterDeduplicateHost(skipped, uploaded, size int64)
+}
+
 // Deduplicate handles the deduplicate stage of the backup process.
 // The implementation is expected to follow RFC document
 // https://docs.google.com/document/d/1EtGlF6UGNy34D_7QsnCheaukp3UwVObZU56PBdd0CQ8/edit#heading=h.jl2qbpcarwp9
@@ -38,6 +43,19 @@ func (w *worker) Deduplicate(ctx context.Context, hosts []hostInfo, limits []DCL
 }
 
 func (w *worker) deduplicateHost(ctx context.Context, h hostInfo) error {
+	if w.dth != nil {
+		w.dth.beforeDeduplicateHost()
+		defer func(sd []snapshotDir) {
+			var skipped, uploaded, size int64
+			for _, v := range sd {
+				skipped += v.Progress.Skipped
+				uploaded += v.Progress.Uploaded
+				size += v.Progress.Size
+			}
+			w.dth.afterDeduplicateHost(skipped, uploaded, size)
+		}(w.hostSnapshotDirs(h))
+	}
+
 	if err := w.setRateLimit(ctx, h); err != nil {
 		return errors.Wrap(err, "set rate limit")
 	}
@@ -108,7 +126,7 @@ func (w *worker) basedOnUUIDGenerationAvailability(ctx context.Context, d snapsh
 	remoteSSTables map[string]struct{}, ssTablesGroupByID map[string][]string,
 ) (deduplicated int64, err error) {
 	// Per every SSTable files group, check if the name uses UUID to identify the SSTable generation.
-	// If the above is true, then check if files exists in the remote storage and remove it locally if exists in rmeote.
+	// If the above is true, then c4heck if files exists in the remote storage and remove it locally if exists in rmeote.
 	for id, ssTableContent := range ssTablesGroupByID {
 		// Check if id is an integer (no UUID)
 		if _, err := strconv.Atoi(id); err == nil {
--- a/pkg/service/backup/worker_deduplicate.go
+++ b/pkg/service/backup/worker_deduplicate.go
@@ -1,0 +1,170 @@
+// Copyright (C) 2024 ScyllaDB
+
+package backup
+
+import (
+	"bytes"
+	"context"
+	"path"
+	"slices"
+	"strconv"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
+	"github.com/scylladb/scylla-manager/v3/pkg/sstable"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/parallel"
+)
+
+// Deduplicate handles the deduplicate stage of the backup process.
+// The implementation is expected to follow RFC document
+// https://docs.google.com/document/d/1EtGlF6UGNy34D_7QsnCheaukp3UwVObZU56PBdd0CQ8/edit#heading=h.jl2qbpcarwp9
+func (w *worker) Deduplicate(ctx context.Context, hosts []hostInfo, limits []DCLimit) (err error) {
+	f := func(h hostInfo) error {
+		w.Logger.Info(ctx, "Removing duplicated files from local snapshot", "host", h.IP)
+		err := w.deduplicateHost(ctx, h)
+		if err == nil {
+			w.Logger.Info(ctx, "Done deduplication", "host", h.IP)
+		}
+		return err
+	}
+
+	notify := func(h hostInfo, err error) {
+		w.Logger.Error(ctx, "Removing duplicated files failed on host", "host", h.IP, "error", err)
+	}
+
+	return inParallelWithLimits(hosts, limits, f, notify)
+}
+
+func (w *worker) deduplicateHost(ctx context.Context, h hostInfo) error {
+	if err := w.setRateLimit(ctx, h); err != nil {
+		return errors.Wrap(err, "set rate limit")
+	}
+
+	dirs := w.hostSnapshotDirs(h)
+	f := func(i int) (err error) {
+		d := dirs[i]
+		dataDst := h.Location.RemotePath(w.remoteSSTableDir(h, d))
+
+		remoteFilesSSTableIDSet := make(map[string]struct{})
+		listOpts := &scyllaclient.RcloneListDirOpts{
+			FilesOnly: true,
+			Recurse:   true,
+		}
+		if err := w.Client.RcloneListDirIter(ctx, h.IP, dataDst, listOpts, func(f *scyllaclient.RcloneListDirItem) {
+			id := sstable.ExtractID(f.Name)
+			remoteFilesSSTableIDSet[id] = struct{}{}
+		}); err != nil {
+			return errors.Wrapf(err, "host %s: listing all files from %s", h.IP, dataDst)
+		}
+
+		// Iterate over all SSTable IDs and group files per ID.
+		ssTablesGroupByID := make(map[string][]string)
+		for _, file := range d.Progress.files {
+			id := sstable.ExtractID(file.Name)
+			ssTablesGroupByID[id] = append(ssTablesGroupByID[id], file.Name)
+		}
+
+		if err := w.basedOnUUIDGenerationAvailability(ctx, d, h, remoteFilesSSTableIDSet, ssTablesGroupByID); err != nil {
+			return errors.Wrap(err, "deduplication based on UUID as generation id content")
+		}
+
+		if err := w.basedOnCrc32Content(ctx, d, h, dataDst, remoteFilesSSTableIDSet, ssTablesGroupByID); err != nil {
+			return errors.Wrap(err, "deduplication based on .crc32 content")
+		}
+
+		return nil
+	}
+
+	notify := func(i int, err error) {
+		d := dirs[i]
+		w.Logger.Error(ctx, "Failed to deduplicate host",
+			"host", d.Host,
+			"keyspace", d.Keyspace,
+			"table", d.Table,
+			"error", err,
+		)
+	}
+
+	return parallel.Run(len(dirs), 1, f, notify)
+}
+
+func (w *worker) basedOnUUIDGenerationAvailability(ctx context.Context, d snapshotDir, h hostInfo,
+	remoteSSTables map[string]struct{}, ssTablesGroupByID map[string][]string,
+) error {
+	// Per every SSTable files group, check if the name uses UUID to identify the SSTable generation.
+	// If the above is true, then check if files exists in the remote storage and remove it locally if exists in rmeote.
+	for id, ssTableContent := range ssTablesGroupByID {
+		// Check if id is an integer (no UUID)
+		if _, err := strconv.Atoi(id); err == nil {
+			continue
+		}
+		// Check if SSTable is available on remote already
+		if _, ok := remoteSSTables[id]; !ok {
+			continue
+		}
+		// Remove duplicated SSTable from local snapshot
+		for _, file := range ssTableContent {
+			// remove if exists in remote
+			localSSTableFileNameWithPath := path.Join(d.Path, file)
+			w.Logger.Debug(ctx, "Removing local snapshot file (deduplication based on generation UUID)",
+				"host", h.IP, "file", localSSTableFileNameWithPath)
+			if err := w.Client.RcloneDeleteFile(ctx, h.IP, localSSTableFileNameWithPath); err != nil {
+				return errors.Wrapf(err, "cannot delete local snapshot's SSTable file %s", localSSTableFileNameWithPath)
+			}
+		}
+	}
+
+	return nil
+}
+
+func (w *worker) basedOnCrc32Content(ctx context.Context, d snapshotDir, h hostInfo, dataDst string,
+	remoteSSTables map[string]struct{}, ssTablesGroupByID map[string][]string,
+) error {
+	// Reference to SSTables 3.0 Data File Format
+	// https://opensource.docs.scylladb.com/stable/architecture/sstable/sstable3/sstables-3-data-file-format.html
+
+	// Per every SSTable files group, compare local <ID>-Digest.crc32 content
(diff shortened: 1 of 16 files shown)


karol-kokoszka:
Integration tests are failing. To be checked.

karol-kokoszka:
This jenkins build is expected to trigger DTests and SCT.
https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-build/718/

Let's check if we are asserting the backup progress.

karol-kokoszka:
DTest -> https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-dtest/1595/ FAILED
SCT Sanity Debian 11 -> https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/debian11-sanity-test/

pkg/service/restore/helper_integration_test.go
```diff
@@ -356,7 +356,7 @@ func dropNonSuperUsers(t *testing.T, s gocqlx.Session) {
 
 func createUser(t *testing.T, s gocqlx.Session, user, pass string) {
 	if err := s.ExecStmt(fmt.Sprintf("CREATE USER '%s' WITH PASSWORD '%s'", user, pass)); err != nil {
-		t.Fatal(errors.Wrapf(err, "create user %s with pass %s", user, pass))
+		t.Logf("create user %s with pass %s, err = {%v}", user, pass, err)
```
Michal-Leszczynski, line 359: Why shouldn't it fail the test? This user should be used for restore, so if it fails, the restore should fail as well.

pkg/sstable/sstable_naming.go
```diff
@@ -42,6 +42,21 @@ func ExtractID(sstable string) string {
...
+
+	if len(parts) < 2 {
+		return sstable + "-Digest.crc32"
+	}
+	// Replace the last part with "Digest.crc32"
+	parts[len(parts)-1] = "Digest.crc32"
+	// Join the parts back together with dashes
+	return strings.Join(parts, "-")
+}
+
```
Michal-Leszczynski, line 59: Wouldn't it be safer to run it through sstable format regex?

pkg/service/restore/service_restore_integration_test.go
```diff
@@ -1001,21 +1002,49 @@ func restoreWithVersions(t *testing.T, target Target, keyspace string, loadCnt,
...
+		bothCorruptCrc32[sstable.Crc32FileNameFromGivenSSTableFile(f)] = struct{}{}
+	}
+	secondCorruptCrc32 := make(map[string]struct{})
+	for _, f := range secondCorrupt {
+		secondCorruptCrc32[sstable.Crc32FileNameFromGivenSSTableFile(f)] = struct{}{}
+	}
+	for f := range bothCorruptCrc32 {
+		firstCorruptCrc32[f] = struct{}{}
+		secondCorruptCrc32[f] = struct{}{}
+	}
```
Michal-Leszczynski, line 1020: Can't we simply append .crc32 files above in:
```go
switch {
			case len(firstCorrupt) < corruptCnt:
				firstCorrupt = append(firstCorrupt, item.Path, sstable.Crc32FileNameFromGivenSSTableFile(item.Path))
			case len(bothCorrupt) < corruptCnt:
				bothCorrupt = append(bothCorrupt, item.Path, ...)
			case len(secondCorrupt) < corruptCnt:
				secondCorrupt = append(secondCorrupt, item.Path, ...)
			}
```
without creating new maps/slices that are appended afterwards?

pkg/service/restore/service_restore_integration_test.go
```diff
@@ -1071,6 +1100,9 @@ func restoreWithVersions(t *testing.T, target Target, keyspace string, loadCnt,
 	for _, tc := range secondCorrupt {
 		swapWithNewest(tc, tag5)
 	}
+	for _, tc := range keys(secondCorruptCrc32) {
+		swapWithNewest(tc, tag5)
+	}
```
Michal-Leszczynski, line 1105: This also wouldn't be needed, if .crc32 files were a part of first/second/bothCorrupt.

Michal-Leszczynski (commented):
Nice work!

In the past there was an idea of making it possible to delete files in batches (I don't think that it is currently supported in our rclone fork). This seems like a simple change on rclone side that would allow us to save maaaany API calls (both in backup purger, and now in backup deduplication). 

@karol-kokoszka maybe we should include it in the scope of this effort (perhaps in a separate series of rclone/SM PRs)?

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,168 @@
...
+		d := dirs[i]
+		w.Logger.Error(ctx, "Failed to deduplicate host",
+			"host", d.Host,
+			"keyspace", d.Keyspace,
+			"table", d.Table,
+			"error", err,
+		)
+	}
+
+	return parallel.Run(len(dirs), 1, f, notify)
```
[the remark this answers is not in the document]
karol-kokoszka: Actually, we could change the limit to something higher... Chunks are partitioned by different dir already.
We set the rate limit on RClone as well.

I don't know what (and if) it would have impact on the deduplication process.  
Let's stay with this low value for now (it's actually the same parallelism as for uploadHost used in upload part), but it could be fine-tuned in the future if needed.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,168 @@
...
+			continue
+		}
+
+		// Check if SSTable content is already available in remote
+		for _, file := range ssTableContent {
+			if slices.Contains(remoteFiles, file) {
+				// remove if exists in remote
+				localSSTableFileNameWithPath := path.Join(d.Path, file)
+				w.Logger.Info(ctx, "Removing local snapshot file (deduplication based on generation UUID)",
+					"host", h.IP, "file", localSSTableFileNameWithPath)
```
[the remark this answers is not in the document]
karol-kokoszka: Right, changed to Debug.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,168 @@
...
+	// If the above is true, then check if files exists in the remote storage and remove it locally if exists in rmeote.
+	for id, ssTableContent := range ssTablesGroupByID {
+		// Check if id is an integer (no UUID)
+		if _, err := strconv.Atoi(id); err == nil {
+			continue
+		}
+
+		// Check if SSTable content is already available in remote
+		for _, file := range ssTableContent {
+			if slices.Contains(remoteFiles, file) {
```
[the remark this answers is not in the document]
karol-kokoszka: It makes sense. Remote files must be grupped per SSTable id.
Let me fix it.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,168 @@
...
+			Recurse:   true,
+		}
+		if err := w.Client.RcloneListDirIter(ctx, h.IP, dataDst, listOpts, func(f *scyllaclient.RcloneListDirItem) {
+			remoteFiles = append(remoteFiles, f.Name)
+		}); err != nil {
+			return errors.Wrapf(err, "host %s: listing all files from %s", h.IP, dataDst)
+		}
+
+		// Iterate over all SSTable IDs and group files per ID.
+		ssTablesGroupByID := make(map[string][]string)
```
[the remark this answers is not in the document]
karol-kokoszka: In Go, a common naming convention for variables and functions is CamelCase, where each word in the name is capitalized, except for the first word. 
When using abbreviations, such as "SS" for "SSTable," it's common practice to capitalize each letter of the abbreviation, even within CamelCase. This helps to preserve the abbreviation's distinctiveness and readability.

Here, "SS" is an abbreviation for "SSTable" (Sorted String Table).
By writing it as ssTablesGroupByID, we are maintaining the readability by clearly distinguishing the abbreviation "SS" from the rest of the words.
This approach makes it easier to recognize the abbreviation within the variable name.

By using ssTablesGroupByID, we maintain the readability and clarity of the variable name. It follows common naming conventions and makes the code easier to understand at a glance, especially for those familiar with the term "SSTable".

In summary, while sstablesGroupByID is not incorrect, ssTablesGroupByID is a better choice for readability and clarity, making it easier for others (and future you) to understand the code quickly.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,168 @@
...
+		if crc32Idx == -1 {
+			continue
+		}
+
+		remoteCRC32FileNameWithPath := path.Join(dataDst, ssTableContent[crc32Idx])
+		remoteCRC32, err := w.Client.RcloneCat(ctx, h.IP, remoteCRC32FileNameWithPath)
+		if err != nil {
+			if strings.Contains(err.Error(), "object not found") {
+				continue
+			}
```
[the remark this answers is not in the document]
karol-kokoszka: I changed it a bit to not send map of ID->list of files for remote, but just set of SSTable ID on remote.
Let me keep this "object not found" check, but just validate if SSTable of given generation ID exists on remote.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,168 @@
...
+			return errors.Wrapf(err, "cannot get content of remote CRC32 %s", remoteCRC32FileNameWithPath)
+		}
+
+		localCRC32FileNameWithPath := path.Join(d.Path, ssTableContent[crc32Idx])
+		localCRC32, err := w.Client.RcloneCat(ctx, h.IP, localCRC32FileNameWithPath)
+		if err != nil {
+			if strings.Contains(err.Error(), "object not found") {
+				continue
+			}
```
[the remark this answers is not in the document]
karol-kokoszka: yes, that is actually very valid

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,168 @@
...
+			return errors.Wrapf(err, "cannot get content of local CRC32 %s", localCRC32FileNameWithPath)
+		}
+
+		// If checksums are equal, then it means that SSTable can be deduplicated as there is no need in
+		// transferring it again to the remote storage.
+		// Deduplication here means to remove SSTable files from local storage.
+		if bytes.Equal(localCRC32, remoteCRC32) {
+			for _, fileToBeRemoved := range ssTableContent {
+				localSSTableFileNameWithPath := path.Join(d.Path, fileToBeRemoved)
+				w.Logger.Info(ctx, "Removing local snapshot file (deduplication based on .crc32)", "host", h.IP, "file", localSSTableFileNameWithPath)
```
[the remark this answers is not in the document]
karol-kokoszka: I thought about moving it to Debug actually. Let's move it to debug.

pkg/service/restore/helper_integration_test.go
karol-kokoszka: it should, revert

pkg/service/restore/service_restore_integration_test.go
karol-kokoszka: But you have counters there. Didn't want to change the exisitng logic.

pkg/service/restore/service_restore_integration_test.go
karol-kokoszka: let's leave it

pkg/sstable/sstable_naming.go
karol-kokoszka: to check if the sstable follows name pattern ?

karol-kokoszka: I don't care if it follows the name pattern for sstable, just want to change the postfix. But maybe I should ?

karol-kokoszka:
@Michal-Leszczynski please re-review.

pkg/service/backup/worker_deduplicate.go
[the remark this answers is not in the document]
Michal-Leszczynski: I'm fine with that for now as #3928 is going to take care for that.

pkg/service/backup/worker_deduplicate.go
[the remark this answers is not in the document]
Michal-Leszczynski: But still, this should never happen and if it does, it means that something is wrong - I think it's safer to return error in such case.
Also, string error checking is problematic in general - it would be better to check for http status code.

So what is the motivation behind this check?
Are you worried that there might be some really old sstables which don't have the .crc32 file?
But if that's the case, we should just change the map key from sstable ID to a pair of sstable ID and format.

pkg/sstable/sstable_naming.go
Michal-Leszczynski: Not sure if that was the case before, but after the changes, it looks like this function is only used in tests.
Could we move it to the test pkg as well?
Then I'm fine with not using the regex - although it would still be safer.

Michal-Leszczynski (commented):
This PR adds the deduplication stage, but it does not change the rclone options of the upload stage - it still has checksum. Is this expected?

In theory it shouldn't be a big deal, but we could disable the checksum comparison in rclone, so that when we checked that the crc32 doesn't match, we don't compare the hashes for nothing.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,194 @@
...
+		localCRC32FileNameWithPath := path.Join(d.Path, ssTableContent[crc32Idx])
+		localCRC32, err := w.Client.RcloneCat(ctx, h.IP, localCRC32FileNameWithPath)
+		if err != nil {
+			return deduplicated, errors.Wrapf(err, "cannot get content of local CRC32 %s", localCRC32FileNameWithPath)
+		}
+
+		// If checksums are equal, then it means that SSTable can be deduplicated as there is no need in
+		// transferring it again to the remote storage.
+		// Deduplication here means to remove SSTable files from local storage.
+		if bytes.Equal(localCRC32, remoteCRC32) {
```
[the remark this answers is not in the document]
karol-kokoszka: AI explained the probability of crc32 hash collision of two files (two sstables data) this way:
```
For two files, the probability of a CRC32 hash collision is 1/2^32
 , which is approximately 2.3 × 10^-10 or 1 in 4.3 billion. This makes CRC32 quite reliable for small-scale use, but for very large datasets or critical applications, stronger hash functions (like SHA-256) are recommended to minimize collision risks.
```

Some external references https://arstechnica.com/civis/threads/crc32-hash-collision-probability.149670/
```
It's simple to figure out. No matter what your keys are the hash will be one of 2^32 different values. Assuming random input, the probability of any of these values appearing is equal. So, the probability of collision between the hashes of two given files is 1 / 2^32. The probability of collisions between any of N given files is (N - 1) / 2^32. <BR><BR>This is of course assuming that CRC-32 is perfectly unbiased, which it isn't, but the approximation isn't a big deal unless you have good knowledge of the key distribution.
```

We shouldn't bother.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,194 @@
...
+			for _, fileToBeRemoved := range ssTableContent {
+				localSSTableFileNameWithPath := path.Join(d.Path, fileToBeRemoved)
+				w.Logger.Debug(ctx, "Removing local snapshot file (deduplication based on .crc32)", "host", h.IP, "file", localSSTableFileNameWithPath)
+				info, err := w.Client.RcloneFileInfo(ctx, h.IP, localSSTableFileNameWithPath)
+				if err != nil {
+					return deduplicated, errors.Wrapf(err, "get file info %s", localSSTableFileNameWithPath)
+				}
```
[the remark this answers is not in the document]
karol-kokoszka: we can
good point

pkg/service/backup/worker_deduplicate.go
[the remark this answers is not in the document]
karol-kokoszka: It's 0.00000002328%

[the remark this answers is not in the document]
Michal-Leszczynski: Yes, but those calculations are made for a single pair of files.
Single table in a big cluster could consist of millions of files.
Cluster has many tables.
SM manages many clusters.
And we would be really sad if a crc32 collision happened at any step.

The good thing is that having a name collision (result of replacing a node by ID) is a rare thing.
Also, new sstables are generated using the UUID, so that also lowers the pool of files exposed to potential crc32 collisions.
Having those in mind, it's always better to stay on the safer side and decrease the probability of the crc32 collision - especially that we already have the size information, so why shouldn't we use it?

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,194 @@
...
+		// Remove duplicated SSTable from local snapshot
+		for _, file := range ssTableContent {
+			// remove if exists in remote
+			localSSTableFileNameWithPath := path.Join(d.Path, file)
+			w.Logger.Debug(ctx, "Removing local snapshot file (deduplication based on generation UUID)",
+				"host", h.IP, "file", localSSTableFileNameWithPath)
+			info, err := w.Client.RcloneFileInfo(ctx, h.IP, localSSTableFileNameWithPath)
+			if err != nil {
+				return deduplicated, errors.Wrapf(err, "get file info %s", localSSTableFileNameWithPath)
+			}
```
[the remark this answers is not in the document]
karol-kokoszka: Very good point. Fixed.

pkg/service/backup/worker_deduplicate.go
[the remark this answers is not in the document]
karol-kokoszka: makes sense to check it against HTTP error code (404)
If you ask about the motivation, then I want to avoid hard stop on backup whenever the file just doesn't exist. Although I think it's not probable, as the code is looking for the file in sstablecontent.

pkg/service/backup/worker_deduplicate.go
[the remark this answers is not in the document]
karol-kokoszka: It is actually not valid, don't know how, but SCT tests failed on that.

pkg/service/backup/worker_deduplicate.go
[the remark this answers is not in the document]
karol-kokoszka: Checking the size makes sense. Fixing.

pkg/sstable/sstable_naming.go
karol-kokoszka: actually, you are right, moved to _test

karol-kokoszka:
Some sample logs (after changing from DEBUG to INFO)
```
Jul 18 09:04:14 manager-regression-master-monitor-node-d5777710-1 scylla-manager[10584]: {"L":"INFO","T":"2024-07-18T09:04:14.845Z","N":"backup.deduplicate","M":"Removing local snapshot file (deduplication based on generation UUID)","host":"10.15.1.78","file":"data:system_distributed_everywhere/cdc_generation_descriptions_v2-234d2227dd633d37ac5fc013e2ea9e6e/snapshots/sm_20240718090409UTC/me-3ghx_0net_4rw9s2bhesqa79ne6z-big-CompressionInfo.db","_trace_id":"LkFf76dqQ-miLcvXFxGsMQ"}
Jul 18 09:04:14 manager-regression-master-monitor-node-d5777710-1 scylla-manager[10584]: {"L":"INFO","T":"2024-07-18T09:04:14.845Z","N":"backup.deduplicate","M":"Removing local snapshot file (deduplication based on generation UUID)","host":"10.15.1.78","file":"data:system_distributed_everywhere/cdc_generation_descriptions_v2-234d2227dd633d37ac5fc013e2ea9e6e/snapshots/sm_20240718090409UTC/me-3ghx_0net_4rw9s2bhesqa79ne6z-big-Data.db","_trace_id":"LkFf76dqQ-miLcvXFxGsMQ"}
Jul 18 09:04:14 manager-regression-master-monitor-node-d5777710-1 scylla-manager[10584]: {"L":"INFO","T":"2024-07-18T09:04:14.846Z","N":"backup.deduplicate","M":"Removing local snapshot file (deduplication based on generation UUID)","host":"10.15.1.78","file":"data:system_distributed_everywhere/cdc_generation_descriptions_v2-234d2227dd633d37ac5fc013e2ea9e6e/snapshots/sm_20240718090409UTC/me-3ghx_0net_4rw9s2bhesqa79ne6z-big-Digest.crc32","_trace_id":"LkFf76dqQ-miLcvXFxGsMQ"}
Jul 18 09:04:14 manager-regression-master-monitor-node-d5777710-1 scylla-manager[10584]: {"L":"INFO","T":"2024-07-18T09:04:14.846Z","N":"backup.deduplicate","M":"Removing local snapshot file (deduplication based on generation UUID)","host":"10.15.1.78","file":"data:system_distributed_everywhere/cdc_generation_descriptions_v2-234d2227dd633d37ac5fc013e2ea9e6e/snapshots/sm_20240718090409UTC/me-3ghx_0net_4rw9s2bhesqa79ne6z-big-Filter.db","_trace_id":"LkFf76dqQ-miLcvXFxGsMQ"}
Jul 18 09:04:14 manager-regression-master-monitor-node-d5777710-1 scylla-manager[10584]: {"L":"INFO","T":"2024-07-18T09:04:14.847Z","N":"backup.deduplicate","M":"Removing local snapshot file (deduplication based on generation UUID)","host":"10.15.1.78","file":"data:system_distributed_everywhere/cdc_generation_descriptions_v2-234d2227dd633d37ac5fc013e2ea9e6e/snapshots/sm_20240718090409UTC/me-3ghx_0net_4rw9s2bhesqa79ne6z-big-Index.db","_trace_id":"LkFf76dqQ-miLcvXFxGsMQ"}
Jul 18 09:04:14 manager-regression-master-monitor-node-d5777710-1 scylla-manager[10584]: {"L":"INFO","T":"2024-07-18T09:04:14.847Z","N":"backup.deduplicate","M":"Removing local snapshot file (deduplication based on generation UUID)","host":"10.15.1.78","file":"data:system_distributed_everywhere/cdc_generation_descriptions_v2-234d2227dd633d37ac5fc013e2ea9e6e/snapshots/sm_20240718090409UTC/me-3ghx_0net_4rw9s2bhesqa79ne6z-big-Scylla.db","_trace_id":"LkFf76dqQ-miLcvXFxGsMQ"}
Jul 18 09:04:14 manager-regression-master-monitor-node-d5777710-1 scylla-manager[10584]: {"L":"INFO","T":"2024-07-18T09:04:14.847Z","N":"backup.deduplicate","M":"Removing local snapshot file (deduplication based on generation UUID)","host":"10.15.1.78","file":"data:system_distributed_everywhere/cdc_generation_descriptions_v2-234d2227dd633d37ac5fc013e2ea9e6e/snapshots/sm_20240718090409UTC/me-3ghx_0net_4rw9s2bhesqa79ne6z-big-Statistics.db","_trace_id":"LkFf76dqQ-miLcvXFxGsMQ"}
Jul 18 09:04:14 manager-regression-master-monitor-node-d5777710-1 scylla-manager[10584]: {"L":"INFO","T":"2024-07-18T09:04:14.848Z","N":"backup.deduplicate","M":"Removing local snapshot file (deduplication based on generation
```
Overall, it will be logged on debug level as there may be plenty of these entries during the backup.

pkg/service/backup/worker_deduplicate.go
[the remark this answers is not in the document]
Michal-Leszczynski: Ok, but please use `http.StatusNotFound` instead of raw 404 int.

Michal-Leszczynski (commented):
Actually, there is a small important thing missing.
AFAIU, the `d.Progress.files` is initialized only during the INDEX stage. This means that pausing and resuming backup during DEDUPLICATION stage would result in `d.Progress.files` being empty and the whole process falling to pieces.

This is the reason why some stages allow reindexing on resume:
```go
		// Skip completed stages
		if run.PrevID != uuid.Nil {
			// Allow reindexing if previous state is manifest creation or upload
			if stage == StageIndex && (prevStage == StageManifest || prevStage == StageUpload) { //nolint: revive
				// continue
			} else if prevStage.Index() > stage.Index() {
				return nil
			}
		}
```

DEDUPLICATE should be one of those stages as well.

What's funny is that it looks like checking for reindexing is bugged (by me), as the moved SCHEMA stage should also be reindexed on resume (as it is before UPLOAD, which supposedly requires reindexation - although I think that it only impacts progress recording/display there).

Michal-Leszczynski:
Perhaps it would be good to have a test which checks backup pause/resume during each stage.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,219 @@
...
+			return deduplicated, errors.Wrapf(err, "cannot get content of remote CRC32 %s", remoteCRC32FileNameWithPath)
+		}
+
+		localCRC32FileNameWithPath := path.Join(d.Path, ssTableContent[crc32Idx])
+		localCRC32, err := w.Client.RcloneCat(ctx, h.IP, localCRC32FileNameWithPath)
+		if err != nil {
+			if scyllaclient.StatusCodeOf(err) == 404 {
+				continue
+			}
```
Michal-Leszczynski, line 177: But how can this file be non-existant if we checked that it exists in the slice above?

karol-kokoszka:
> Actually, there is a small important thing missing.
AFAIU, the d.Progress.files is initialized only during the INDEX stage. This means that pausing and resuming backup during DEDUPLICATION stage would result in d.Progress.files being empty and the whole process falling to pieces.

Nice catch !
It should be reindexed indeed ! Added integration test to `service_deduplicate_integration_test` that proves it.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,219 @@
...
+		// transferring it again to the remote storage.
+		// Deduplication here means to remove SSTable files from local storage.
+		if bytes.Equal(localCRC32, remoteCRC32) {
+			// validate sizes first
+			for _, file := range ssTableContent {
+				remoteFilePath := path.Join(dataDst, file)
+				remoteInfo, err := w.Client.RcloneFileInfo(ctx, h.IP, remoteFilePath)
+				if err != nil {
+					return deduplicated, errors.Wrapf(err, "get file info %s", remoteFilePath)
+				}
```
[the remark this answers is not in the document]
karol-kokoszka: OK, please take care on that with the PR related to batches.

pkg/service/backup/worker_deduplicate.go
karol-kokoszka: yeah, good point. I'm removing this check for remote crc32

karol-kokoszka:
@Michal-Leszczynski please re-review

pkg/service/backup/worker_deduplicate.go
Michal-Leszczynski: I think that this check should be removed for local crc 32, since we check if this file exists a few lines above:
```go
		crc32Idx := slices.IndexFunc(ssTableContent, func(s string) bool {
			return strings.HasSuffix(s, "Digest.crc32")
		})
		if crc32Idx == -1 {
			continue
		}
		if _, ok := remoteSSTables[id]; !ok {
			continue
		}

		remoteCRC32FileNameWithPath := path.Join(dataDst, ssTableContent[crc32Idx])
		remoteCRC32, err := w.Client.RcloneCat(ctx, h.IP, remoteCRC32FileNameWithPath)
		if err != nil {
			return deduplicated, errors.Wrapf(err, "cannot get content of remote CRC32 %s", remoteCRC32FileNameWithPath)
		}

		localCRC32FileNameWithPath := path.Join(d.Path, ssTableContent[crc32Idx])
		localCRC32, err := w.Client.RcloneCat(ctx, h.IP, localCRC32FileNameWithPath)
		if err != nil {
			if scyllaclient.StatusCodeOf(err) == 404 {
				continue
			}
```

pkg/service/backup/worker_deduplicate.go
```diff
@@ -0,0 +1,232 @@
...
+	"github.com/scylladb/scylla-manager/v3/pkg/sstable"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/parallel"
+)
+
+var (
+	// DeduplicateHostBeforeInterceptor is the function executed at the beginning of every deduplicate hots.
+	DeduplicateHostBeforeInterceptor = func() {}
+	// DeduplicateHostAfterInterceptor is a function executed after every deduplicate host invocation.
+	DeduplicateHostAfterInterceptor = func(dirs []SnapshotDir) {}
+)
```
Michal-Leszczynski, line 25: I don't really like having some exported global variables with no concurrent access control (not to mention that the second function allows for modifying snapshot dirs passed as function argument).

Couldn't we use the already existing client interceptors and wait for (and block) calls to `"POST": "/rclone/sync/movedir"`, so that we know when the deduplication stage is finished?
Couldn't we fetch progress with the usual `GetProgress` and get the size/deduplicated/uploaded from there?

This would also allow for not exporting the internals `snapshotDir` struct.

Michal-Leszczynski:
@karol-kokoszka kudos for adding dedicated deduplication test. I re-review the changes and left some small comments.

pkg/service/backup/worker_deduplicate.go
karol-kokoszka: Aii, you are right. I removed it from remote, but should stay in remote.
Fixed.

pkg/service/backup/worker_deduplicate.go
karol-kokoszka: using already existing interceptor is not option, it won't work here.
Will think about syncing (if it's needed).

This is not the perfect solution I agree, but should be fine. It's clear that these variables are for testing only.

karol-kokoszka: If I cancel the context from `/rclone/sync/movedir` then I'm doing it after deduplicate stage, so everything is deduplicated alteady.
To achieve the same effect as I have there, I would need to intercept RcloneSetBandwidthLimit.
I see no direct equivalent of after.

> (not to mention that the second function allows for modifying snapshot dirs passed as function argument).

It allows modifying snapshot dir, but I don't see it as a problem. Why do you think it's a problem ?

Michal-Leszczynski: But `DeduplicateHostAfterInterceptor` also happens after the deduplication is done.
The `DeduplicateHostBeforeInterceptor` could be replaced by `"POST": "/rclone/operations/deletefile"` with filtering for sstable file. Then you can use mutex to allow deduplication only on a single host, like in your current approach.

>>  (not to mention that the second function allows for modifying snapshot dirs passed as function argument).
> It allows modifying snapshot dir, but I don't see it as a problem. Why do you think it's a problem ?

Because it's unsafe to modify them there  - backup process assume they are exactly the same as the indexing stage created them. I know that we know that they shouldn't be modified, but we should try to write code which is robust even when somebody new looks at it.

So in general I'm in favor of refactoring it to the client interceptors, or to some more general  approach (e.g. allow for interceptors between any given stages), but I'm fine with it in the context of this PR, as long as those interceptors are not exported (they can be exported to tests only via `export_test.go`) , are marked as "use for testing only" in the comments, and they return a copy of snapshot dirs.
On the other hand, making a copy might be a pretty expensive operation, so maybe it would be better to just refactor function to take the already aggregated info and not snapshot dirs directly, like:
```go
deduplicateHostAfterTestInterceptor = func(deduplicated, uploaded, totalSize int64) {}
```

karol-kokoszka: OK, I found better approach.
Will hide it with build tags, so it's gonna be available for integration tests only.

karol-kokoszka: OK, it's refactored. 
All the hooks are hidden and available for the builds that includes `integration` or `all` tags.
`snapshotDirs` is unexported again.
Hooks belong to service and worker now.

Michal-Leszczynski: @karol-kokoszka but now `SetDeduplicateTestHooks` is exported and can be used in the backup pkg (not only in the tests)?
I think that using `export_test.go` could solve this problem, but I'm fine with merging this PR in its current state.

Michal-Leszczynski (approved):

pkg/service/backup/worker_deduplicate.go
karol-kokoszka: `SetDeduplicateTestHooks` is not exported.
It's guarded with build tags

```
//go:build all || integration
// +build all integration
```

You can verify it by adding the following to backup/service.go:
```
	s.SetDeduplicateTestHooks(func() {}, func(skipped, uploaded, size int64) {})
```

and then try to compile `make build`

=> 
```
➜  scylla-manager git:(cheksum_comparison) ✗ make build
==> Building sctool
==> Building agent
==> Building scylla-manager
# github.com/scylladb/scylla-manager/v3/pkg/service/backup
pkg/service/backup/service.go:790:4: s.SetDeduplicateTestHooks undefined (type *Service has no field or method SetDeduplicateTestHooks)
make: *** [Makefile:201: build-server] Error 1

```

PR #3888, opened 2024-06-19
Fix await schema agreement

### LINKED ISSUE #3887: Await schema ageement procedure is broken
There are 2 main problems with the current await schema agreement stage.

1. It is bugged, as SM implementation uses 2 separate queries to get all schema versions. Those queries might be routed to different hosts in the cluster resulting in skipping schema_version from one node (the one to which the first query was routed):  
```go
	const (
		peerSchemasStmt = "SELECT schema_version FROM system.peers"
		localSchemaStmt = "SELECT schema_version FROM system.local WHERE key='local'"
	)

	return retry.WithNotify(ctx, func() error {
		var v []string
		if err := clusterSession.Query(peerSchemasStmt, nil).SelectRelease(&v); err != nil {
			return retry.Permanent(err)
		}
		var lv string
		if err := clusterSession.Query(localSchemaStmt, nil).GetRelease(&lv); err != nil {
			return retry.Permanent(err)
		}
```

2. It works by checking schema agreement from just a single node POV - so in the same way as it is implemented in `gocql`.
This is not enough in terms of ensuring that described schema is not older than the snapshoted data (see https://github.com/scylladb/scylladb/issues/19213#issue-2343849109 for more details).

In order to fix that, we could open separate, single host session to all backed up hosts and query their local system_schema version and await schema agreement on those values. The problem is that this is a passive way to await schema agreement - so if the schema changes keep on coming, it might never be achieved. 

Another idea discussed in https://github.com/scylladb/scylladb/issues/19213 is to perform read barrier. It does not ensure schema agreement, but it ensures that the schema on the node on which the barrier was performed, is not older than the data snapshoted on all other nodes in the previous stages of the backup procedure - which is enough for backup purposes. Also, it is an active approach which should not take much time or retries.

The problem with this approach is that it won't work when the cluster has less then QUORUM nodes alive, as read barrier is a raft mechanism that doesn't work under such conditions. So when read barrier fails, SM won't backup schema, but it will proceed with uploading user data. In case there was a need to restore schema, it could still be done by the snapshoted system_schema sstables in a hacky/workaround way.

Perhaps in the future there will be Scylla API to issue read barrier, or querying DESC SCHEMA will do it on its own, but for now SM will go with the read barrier approach described above.  

cc: @karol-kokoszka @tzach @mykaul

Michal-Leszczynski:
Fixes #3887

### DIFF
--- a/pkg/service/backup/service.go
+++ b/pkg/service/backup/service.go
@@ -817,23 +817,22 @@ func (s *Service) Backup(ctx context.Context, clusterID, taskID, runID uuid.UUID
 			return w.Snapshot(ctx, hi, target.SnapshotParallel)
 		},
 		StageAwaitSchema: func() error {
-			clusterSession, err := s.clusterSession(ctx, clusterID)
+			descSchemaHost := liveNodes[0].Addr
+			session, err := s.clusterSession(ctx, clusterID, cluster.SingleHostSessionConfigOption(descSchemaHost))
 			if err != nil {
-				w.Logger.Info(ctx, "No CQL cluster session, backup of schema as CQL files will be skipped", "error", err)
-				return nil
+				if errors.Is(err, cluster.ErrNoCQLCredentials) {
+					w.Logger.Error(ctx, "No CQL cluster credentials, backup of schema as CQL files will be skipped", "error", err)
+					return nil
+				}
+				return errors.Wrapf(err, "create single host (%s) session to", descSchemaHost)
 			}
-			defer clusterSession.Close()
-
-			if err := w.AwaitSchemaAgreement(ctx, clusterSession); err != nil {
-				return errors.Wrap(err, "await schema agreement")
-			}
+			defer session.Close()
 
 			var hosts []string
 			for _, h := range hi {
 				hosts = append(hosts, h.IP)
 			}
-
-			if err := w.DumpSchema(ctx, clusterSession, hosts); err != nil {
+			if err := w.DumpSchema(ctx, session, hosts); err != nil {
 				return errors.Wrap(err, "dump schema")
 			}
 
--- a/pkg/service/cluster/service.go
+++ b/pkg/service/cluster/service.go
@@ -583,12 +583,30 @@ func (s *Service) ListNodes(ctx context.Context, clusterID uuid.UUID) ([]Node, e
 	return nodes, nil
 }
 
+// SessionConfigOption defines function modifying cluster config that can be used when creating session.
+type SessionConfigOption func(ctx context.Context, clusterID uuid.UUID, client *scyllaclient.Client, cfg *gocql.ClusterConfig) error
+
+// SingleHostSessionConfigOption ensures that session will be connected only to the single, provided host.
+func SingleHostSessionConfigOption(host string) SessionConfigOption {
+	return func(ctx context.Context, clusterID uuid.UUID, client *scyllaclient.Client, cfg *gocql.ClusterConfig) error {
+		ni, err := client.NodeInfo(ctx, host)
+		if err != nil {
+			return errors.Wrapf(err, "fetch node (%s) info", host)
+		}
+		cqlAddr := ni.CQLAddr(host)
+		cfg.Hosts = []string{cqlAddr}
+		cfg.HostFilter = gocql.WhiteListHostFilter(cqlAddr)
+		cfg.DisableInitialHostLookup = true
+		return nil
+	}
+}
+
 // SessionFunc returns CQL session for given cluster ID.
-type SessionFunc func(ctx context.Context, clusterID uuid.UUID) (gocqlx.Session, error)
+type SessionFunc func(ctx context.Context, clusterID uuid.UUID, opts ...SessionConfigOption) (gocqlx.Session, error)
 
 // GetSession returns CQL session to provided cluster.
-func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID) (session gocqlx.Session, err error) {
-	s.logger.Debug(ctx, "GetSession", "cluster_id", clusterID)
+func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID, opts ...SessionConfigOption) (session gocqlx.Session, err error) {
+	s.logger.Info(ctx, "Get session", "cluster_id", clusterID)
 
 	client, err := s.CreateClientNoCache(ctx, clusterID)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


Michal-Leszczynski:
@karol-kokoszka PR is ready for review!

pkg/service/cluster/service.go
```diff
@@ -609,57 +621,102 @@ func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID) (session
...
+
+	return gocqlx.WrapSession(cfg.CreateSession())
+}
+
+// SingleHostSessionFunc returns single host CQL session for given cluster ID.
+type SingleHostSessionFunc func(ctx context.Context, clusterID uuid.UUID, host string) (gocqlx.Session, error)
+
+// GetSingleHostSession returns single host CQL session to provided cluster.
+// It's useful when we need to ensure that CQL queries are routed to given host.
+func (s *Service) GetSingleHostSession(ctx context.Context, clusterID uuid.UUID, host string) (session gocqlx.Session, err error) {
```
karol-kokoszka, line 640: There is no need for introducing code duplication.
The signature of `GetSession` can be changed to `func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID, opts ...SessionOpt) (session gocqlx.Session, err error) {`

Then,
```
type SessionOpt interface {
	apply(*sessionConfig)
}

type sessionConfig struct {
	host        string
	singleHost  bool
}

type singleHostOpt struct {
	host string
}

func (o singleHostOpt) apply(cfg *sessionConfig) {
	cfg.singleHost = true
	cfg.host = o.host
}

func WithSingleHost(host string) SessionOpt {
	return singleHostOpt{host: host}
}
```

and GetSession body:
```
func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID, opts ...SessionOpt) (session gocqlx.Session, err error) {
	s.logger.Info(ctx, "Get session", "cluster_id", clusterID)

	client, err := s.CreateClientNoCache(ctx, clusterID)
	if err != nil {
		return session, errors.Wrap(err, "get client")
	}
	defer logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")

	config := sessionConfig{}
	for _, opt := range opts {
		opt.apply(&config)
	}

	var ni *scyllaclient.NodeInfo
	if config.singleHost {
		s.logger.Info(ctx, "Get single host session", "host", config.host)
		ni, err = client.NodeInfo(ctx, config.host)
		if err != nil {
			return session, errors.Wrap(err, "fetch node info")
		}
	} else {
		ni, err = client.AnyNodeInfo(ctx)
		if err != nil {
			return session, errors.Wrap(err, "fetch node info")
		}
	}

	var sessionHosts []string
	if config.singleHost {
		sessionHosts = []string{ni.CQLAddr(config.host)}
	} else {
		sessionHosts, err = GetRPCAddresses(ctx, client, client.Config().Hosts)
		if err != nil {
			s.logger.Info(ctx, "GetSession", "err", err)
			if errors.Is(err, ErrNoRPCAddressesFound) {
				return session, err
			}
		}
	}

	cfg := gocql.NewCluster(sessionHosts...)
	if config.singleHost {
		cfg.HostFilter = gocql.WhiteListHostFilter(sessionHosts...)
		cfg.DisableInitialHostLookup = true
	}

	if err := s.extendClusterConfigWithAuthentication(clusterID, ni, cfg); err != nil {
		return session, err
	}
	if err := s.extendClusterConfigWithTLS(ctx, clusterID, ni, cfg); err != nil {
		return session, err
	}

	return gocqlx.WrapSession(cfg.CreateSession())
}
```

pkg/util/query/query.go
```diff
@@ -78,3 +81,28 @@ func DescribeSchemaWithInternals(session gocqlx.Session) (DescribedSchema, error
...
+// has already applied all raft commands present in the raft log.
+func RaftReadBarrier(session gocqlx.Session) error {
+	// As this functionality is not available from Scylla API,
+	// SM uses the workaround described in https://github.com/scylladb/scylladb/issues/19213.
+	// Read barrier is performed before any schema change, so it's enough to try to apply
+	// some dummy schema changing CQL statement.
+	// In order to avoid misleading errors, it's good to use the IF EXISTS clause.
+	// Still, it might happen that SM does not have the permissions to perform dummy schema change.
+	// As described in the issue, this type of error still results in performing read barrier,
+	// so SM should treat it as if everything went fine.
```
karol-kokoszka, line 96: Please extend the comment with information that Scylla is expected to extend the API in one of the upcoming releases, so that this workaround in SM is not forgotten and can be removed at some point in time.
`TODO: use API call instead of DROP TABLE call` is one of the options.

pkg/service/backup/service.go
```diff
@@ -817,23 +818,27 @@ func (s *Service) Backup(ctx context.Context, clusterID, taskID, runID uuid.UUID
...
+					return nil
+				}
+				return errors.Wrapf(err, "create single host (%s) session to", descSchemaHost)
 			}
-			defer clusterSession.Close()
+			defer session.Close()
 
-			if err := w.AwaitSchemaAgreement(ctx, clusterSession); err != nil {
-				return errors.Wrap(err, "await schema agreement")
+			if err := query.RaftReadBarrier(session); err != nil {
```
karol-kokoszka, line 832: Isn't is just for Scylla's that support RAFT ?
Please leave the old approach for older versions of Scylla.

karol-kokoszka (changes requested):
The RAFT read barrier should be used only on Scylla server that uses RAFT.

pkg/service/cluster/service.go
Michal-Leszczynski: That's a good idea, but I'm not convinced to the proposed format.
It adds a lot of similar types and requires a lot of (3!) if statements in the method body, which makes it less comprehensible. I would be more in favor of something like:
```go
type SessionConfigOption func(ctx context.Context, clusterID uuid.UUID, client *scyllaclient.Client, cfg *gocql.ClusterConfig) error

func SingleHostSessionOpt(host string) SessionConfigOption {
	return func(ctx context.Context, clusterID uuid.UUID, client *scyllaclient.Client, cfg *gocql.ClusterConfig) error {
		ni, err := client.NodeInfo(ctx, host)
		if err != nil {
			return errors.Wrapf(err, "fetch node (%s) info", host)
		}
		cqlAddr := ni.CQLAddr(host)
		cfg.Hosts = []string{cqlAddr}
		cfg.HostFilter = gocql.WhiteListHostFilter(cqlAddr)
		cfg.DisableInitialHostLookup = true
		return nil
	}
}
```

And:
```go
// GetSession returns CQL session to provided cluster.
func (s *Service) GetSession(ctx context.Context, clusterID uuid.UUID, opts ...SessionConfigOption) (session gocqlx.Session, err error) {
	s.logger.Info(ctx, "Get session", "cluster_id", clusterID)

	client, err := s.CreateClientNoCache(ctx, clusterID)
	if err != nil {
		return session, errors.Wrap(err, "get client")
	}
	defer logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")

	cfg := gocql.NewCluster()
	for _, opt := range opts {
		if err := opt(ctx, clusterID, client, cfg); err != nil {
			return session, err
		}
	}
	// Fill hosts if they weren't specified by the options
	if len(cfg.Hosts) == 0 {
		sessionHosts, err := GetRPCAddresses(ctx, client, client.Config().Hosts)
		if err != nil {
			s.logger.Info(ctx, "GetSession", "err", err)
			if errors.Is(err, ErrNoRPCAddressesFound) {
				return session, err
			}
		}
		cfg.Hosts = sessionHosts
	}

	ni, err := client.AnyNodeInfo(ctx)
	if err != nil {
		return session, errors.Wrap(err, "fetch node info")
	}
	if err := s.extendClusterConfigWithAuthentication(clusterID, ni, cfg); err != nil {
		return session, err
	}
	if err := s.extendClusterConfigWithTLS(ctx, clusterID, ni, cfg); err != nil {
		return session, err
	}

	return gocqlx.WrapSession(cfg.CreateSession())
}
```

The annoying part here is how to avoid querying node info multiple times (or creating uncached client many times), but implementing #3892 should simplify that - with it we could even rewrite filling the TLS/Auth cluster config options with the same approach.

@karol-kokoszka What do you think about this idea (both in the future context of #3892, and the context of this PR)?

karol-kokoszka: This comment is not a blocker for the PR.
So, if you say that #3892 will simplify the logic, I agree.

Please check the one related to RAFT being supported by scylla server.

pkg/service/backup/service.go
Michal-Leszczynski: That's true!
This does not break anything for non raft schema clusters in principle, because:
- the raft read barrier is correct CQL stmt in older scylla versions
- the previously used await schema procedure is bugged
- this schema wasn't safe to use with restore anyway

But I can change it to use the driver await schema method (not bugged) and still await the schema agreement just to continue the old behavior.

karol-kokoszka (approved):

PR #3898, opened 2024-06-24
Fix test tablet setup

Michal-Leszczynski:
This PR does the following:
- fixes tablets test env setup
- automates the process of generating gh actions testing workflows
- automates the process of generating gh testing badges
- improves stability of repair tests which were flaky due to token ranges batching


Michal-Leszczynski:

Makefile
```diff
@@ -173,7 +171,7 @@ start-dev-env: .testing-up deploy-agent build-cli
 
 .PHONY: .testing-up
 .testing-up:
-	@IPV6=$(IPV6) SCYLLA_VERSION=$(SCYLLA_VERSION) RAFT_SCHEMA=$(RAFT_SCHEMA) RAFT_TOPOLOGY=$(RAFT_TOPOLOGY) TABLETS=$(TABLETS) make -C testing build down up
```
karol-kokoszka, line 176: I understand RAFT is always enabled with 6.0, but what about older versions ? 
We are still supporting 
```
Scylla Manager 3.3 supports the following Scylla Enterprise releases:

2024.1
2023.1
And the following Open Source release (limited to 5 nodes see [license 1](https://www.scylladb.com/scylla-manager-software-license-agreement/)):

6.0
5.4
```
https://forum.scylladb.com/t/release-scylla-manager-3-3-0/2231

What about our CI and integration tests for older scylla versions ?

karol-kokoszka (approved):

Makefile
Michal-Leszczynski: I deleted `RAFT_TOPOLOGY` because it was experimental in 5.4, and it is always enabled in 6.0 (and we never tested against it with older Scylla versions).
Also, it was added mostly because it was required for enabling tablets.

Usually SM does not need to be tested against different Scylla configs (except for different versions and IP types), but `RAFT_SCHEMA` and `TABLETS` were important exceptions that impacted SM procedure.

Michal-Leszczynski: @karol-kokoszka can I merge this PR?

karol-kokoszka: Yes, go on. Thanks.

Issue #3901, opened 2024-06-25
Use the replace_node_first_boot option rather than replace_address


bhalevy:
Since https://github.com/scylladb/scylladb/commit/1577aa8098cac7bc2ef0b7f9523044383e14896e (5.2), the `replace_address` and `replace_address_first_boot` options are deprecated and their usage is discouraged, and warned about (when used).
Instead, the user should use the `replace_node_first_boot` which states the replaced node's host_id rather than its ip address.

karol-kokoszka:
Scylla Manager is not using `replace_address` config option at all, but thanks for the heads up !
I removed it from the scylla client we use.

PR #3905, opened 2024-06-25
swagger(scylla): remove endpoint getting replace_address config value

karol-kokoszka:
Fixes https://github.com/scylladb/scylla-manager/issues/3901

---

Please make sure that:


mykaul:
Why did we ever have this endpoint in Manager?

karol-kokoszka:
> Why did we ever have this endpoint in Manager?

This is not the endpoint in the Manager. This is the endpoint defined for the Scylla API client, which is generated based on Swagger and is used by Scylla Manager solely to access the Scylla API.

Issue #3878, opened 2024-06-10
docs: Issue on page Install ScyllaDB Manager


pogzie:
I would like to report an issue on page https://manager.docs.scylladb.com/branch-3.2/install-scylla-manager

### Problem

System Requirements https://manager.docs.scylladb.com/stable/install-scylla-manager.html#system-requirements section has broken links for "Scylla Monitoring Stack" and "Monitoring requirements"

###  Suggest a fix

Run a dead link checker for the whole site or manually fix to https://monitoring.docs.scylladb.com/stable/ and https://monitoring.docs.scylladb.com/stable/install/monitoring-stack.html#minimal-production-system-recommendations respectively.

Issue #3872, opened 2024-05-31
Github Actions CI - add section on README.md


karol-kokoszka:
We would like to move the integration tests badges down in the README.md file under its own section.
The name may be "Scylla integration status" and badges may be divided into explicit sections (scylla version, etc...).

What is the must, is the information of what test is skipped and why.
E.g. "system_auth_v2" data restore is not supported with Scylla 6.0.

Michal-Leszczynski:
Also, we should describe it in our docs!

Issue #3916, opened 2024-07-03
docs: Issue on page Specification


bhalevy:
I would like to report an issue on page https://manager.docs.scylladb.com/branch-3.3/backup/specification

### Problem

Nodes under e.g. `meta/` or `sst/` are identified by a UUID.
I presume it's the node host_id (well, it should be :))
But the manifest file section mentions only that we keep the node's `ip` address, but not its `host_id`.
Is that the case or is the document not up-to-date?

Nodes may change their ip address, but their host_id are stable as long as the node is alive and wasn't removed or replaced.

Also, sstable generation were changes to a uuid as well, and they are now globally unique.
It would be good to update the documentation respectively.

###  Suggest a fix

Michal-Leszczynski:
@karol-kokoszka What do you think about adding the host ID to the manifest? 
In general, I think it wouldn't hurt to add the whole:
```go
// ManifestInfo represents manifest on remote location.
type ManifestInfo struct {
	Location    Location
	DC          string
	ClusterID   uuid.UUID
	NodeID      string
	TaskID      uuid.UUID
	SnapshotTag string
	Temporary   bool // <-- without this field
}
```
to the currently backed up `ManifestContentWithIndex` (so that we back up the whole `ManifestInfoWithContent`).

karol-kokoszka:
> Nodes under e.g. meta/ or sst/ are identified by a UUID.
I presume it's the node host_id (well, it should be :))
But the manifest file section mentions only that we keep the node's ip address, but not its host_id.
Is that the case or is the document not up-to-date?

> Nodes may change their ip address, but their host_id are stable as long as the node is alive and wasn't removed or replaced.

I don't think it's a significant issue that we don't store the host_id in the manifest file, as the path to the given manifest includes/defines the host_id. The UUID that @bhalevy mentioned is indeed the host_id.

However, it wouldn't be a big deal to add it. If we do add it, then all manifests that don't currently include the host_id will be deserialized with an empty string.

karol-kokoszka:
> @karol-kokoszka What do you think about adding the host ID to the manifest?

Sure, why not.

karol-kokoszka:
We plan to extend the backup manifest to support 1-1 restore.
Let's include host_id into the manifest file.

Linking it to the main issue https://github.com/scylladb/scylla-manager/issues/4104

Michal-Leszczynski:
@VAveryanov8 can we close this issue in #4177?

[...]

Issue #3879, opened 2024-06-10
docs: Issue on page Examples


pogzie:
I would like to report an issue on page https://manager.docs.scylladb.com/branch-3.2/repair/examples

### Problem

Example on https://manager.docs.scylladb.com/stable/repair/examples.html#schedule-a-weekly-repair using the Docker example doesnt work. Also, if basing from the Docker example, it might be best to standardize the use of the cluster name. 

```
scylla-manager@354a2ccb890d:/$ sctool tasks -c test
+------------------------+--------------+--------+----------+---------+-------+------------------------+------------+--------+------------------------+
| Task                   | Schedule     | Window | Timezone | Success | Error | Last Success           | Last Error | Status | Next                   |
+------------------------+--------------+--------+----------+---------+-------+------------------------+------------+--------+------------------------+
| healthcheck/alternator | @every 15s   |        | UTC      | 32      | 0     | 10 Jun 24 04:35:09 UTC |            | DONE   | 10 Jun 24 04:35:24 UTC |
| healthcheck/cql        | @every 15s   |        | UTC      | 32      | 0     | 10 Jun 24 04:35:09 UTC |            | DONE   | 10 Jun 24 04:35:24 UTC |
| healthcheck/rest       | @every 1m0s  |        | UTC      | 8       | 0     | 10 Jun 24 04:35:09 UTC |            | DONE   | 10 Jun 24 04:36:09 UTC |
| repair/all-weekly      | 0 23 * * SAT |        | UTC      | 0       | 0     |                        |            | NEW    | 15 Jun 24 23:00:00 UTC |
+------------------------+--------------+--------+----------+---------+-------+------------------------+------------+--------+------------------------+
scylla-manager@354a2ccb890d:/$ sctool repair -c test -s now+4h -i 7d
Flag --start-date has been deprecated, use cron instead
Flag --interval has been deprecated, use cron instead
Error: create repair target: create repair plan: no replicas to repair
Trace ID: LlgQjyfjTnW4hiTQuluHOg (grep in scylla-manager logs)
```

###  Suggest a fix

If we're pushing the reader to not use deprecated flags, this should also be fixed by using `--cron`. Either way, using cron and following the example it errors out. Further investigation required. 

```
scylla-manager@354a2ccb890d:/$ sctool repair -c test --cron @daily
Error: create repair target: create repair plan: no replicas to repair
Trace ID: vUKn8bgGTn621SWtw-CVGQ (grep in scylla-manager logs)
```

[...]

PR #3917, opened 2024-07-03
Ml/update docs

Michal-Leszczynski:
This PR addresses several doc related issues.

Ref https://github.com/scylladb/scylla-manager/issues/3879
Ref https://github.com/scylladb/scylla-manager/issues/3916
Fixes https://github.com/scylladb/scylla-manager/issues/3878
Fixes https://github.com/scylladb/scylla-manager/issues/3872

### DIFF
--- a/.github/cfg/main.go
+++ b/.github/cfg/main.go
@@ -20,7 +20,7 @@ func (cfg integrationTestCfg) name() string {
 	parts := []string{
 		"integration",
 		"tests",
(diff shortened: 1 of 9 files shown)


Michal-Leszczynski:
Note that this PR will conflict with automatic badge generation introduced in #3898.

karol-kokoszka (approved):

PR #3918, opened 2024-07-04
Don't backup views

### LINKED ISSUE #3771: Don't backup views
Backing up views doesn't make any sense as the advised way to restore them is by recreating them on restored table. This would require to first validate that the base table is even being backed up. Checking which tables are views and which are not can currently be done only via cql session. It might be available via scylla client if https://github.com/scylladb/scylladb/issues/17958 is implemented.

cc: @karol-kokoszka @tzach

Michal-Leszczynski:
Fixes #3771

### DIFF
--- a/pkg/service/backup/service.go
+++ b/pkg/service/backup/service.go
@@ -192,6 +192,12 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
 		return t, errors.Wrap(err, "invalid location")
 	}
 
+	// Get live nodes
+	t.liveNodes, err = s.getLiveNodes(ctx, client, t.DC)
+	if err != nil {
+		return t, err
+	}
+
 	targetDCs := strset.New(t.DC...)
 
 	// Filter keyspaces
@@ -199,7 +205,7 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
 	if err != nil {
 		return t, err
 	}
-	rings := make(map[string]scyllaclient.Ring)
+
 	keyspaces, err := client.Keyspaces(ctx)
 	if err != nil {
 		return t, errors.Wrapf(err, "read keyspaces")
@@ -221,6 +227,7 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
 	}
 
 	ringDescriber := scyllaclient.NewRingDescriber(ctx, client)
+	liveNodes := strset.New(t.liveNodes.Hosts()...)
 	for _, keyspace := range keyspaces {
 		tables, err := client.Tables(ctx, keyspace)
 		if err != nil {
@@ -229,6 +236,9 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
 
 		var filteredTables []string
 		for _, tab := range tables {
+			if !f.Check(keyspace, tab) {
+				continue
+			}
 			// Get the ring description and skip local data
 			ring, err := ringDescriber.DescribeRing(ctx, keyspace, tab)
 			if err != nil {
@@ -243,10 +253,12 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
 				if !targetDCs.HasAny(ring.Datacenters()...) {
 					continue
 				}
+				for _, rt := range ring.ReplicaTokens {
+					if !liveNodes.HasAny(rt.ReplicaSet...) {
+						return t, errors.Errorf("the whole replica set %v of %s.%s is down", rt.ReplicaSet, keyspace, tab)
+					}
+				}
 			}
-
-			// Collect ring information
-			rings[keyspace+"."+tab] = ring
 
 			// Do not filter system_schema
 			if keyspace == systemSchema {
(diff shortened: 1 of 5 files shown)


Michal-Leszczynski:
@annastuchlik could you review the last commit related to docs (a6eeaf590a088ad871379098bb393696dc5a95f3)?

pkg/service/backup/service.go
```diff
@@ -1212,3 +1217,70 @@ func (s *Service) DeleteSnapshot(ctx context.Context, clusterID uuid.UUID, locat
...
+		views:      views,
+		liveNodes:  liveDCNodes,
+		dcNodesCnt: dcNodesCnt,
+		rd:         rd,
+	}
+}
+
+// Returns true if table should be backed up.
+// Return error if table should, but can't be backed up.
+func (tf tableFilter) check(ctx context.Context, ks, tab string) (bool, error) {
```
karol-kokoszka, line 1244: It's not intuitive to use this function in the code.
I don't know what it checks, must read the function body to understand and must read in what context it has been used.

The struct is called `tableFilter`, it could be set of Filter abstractions.
```
type filter interface {
   Filter(ctx context.Context, ks, table string) bool
}
```
}

Then you can have set of filters that implements this interface:
- keyspace
- view
- dataIntegrity (ring stuff) etc (although dataIntegrity is not a filter, it's a validity check)

The struct that is responsible for definitng the target (so the Target struct actually) could be created not just via:
```
	t := Target{}
```
but via sth like `BuildTarget(ctx context.Context, clusterID uuid.UUID, properties json.RawMessage) Target` or `NewTarget` in either backup/model.go ot backup/target.go. Worth to move GetTarget from service. It's not a simple getter, it has plenty of logic inside.

This `BuildTarget` (or whatever) can just be split per i.e.:
- build retention policy
- get the cluster data topology (don't know if it's accurate name, but just get all keyspaces and tables available on the cluster)
- filter according to the properties (which keyspaces, tables, no views)
- validate data integrity (check rings) against live nodes

WDYT ?

karol-kokoszka (commented):
I encourage you to do bit bigger refactor of GetTarget, please check the comment and let me know WDYT or if you have some other idea that can make this part more intuitive.

pkg/service/backup/service.go
Michal-Leszczynski: Sure, I will try to make a general refactor of `GetTarget` tomorrow!

karol-kokoszka:
@Michal-Leszczynski it's a draft PR, I understand it's still in progress ?

Michal-Leszczynski:
@karol-kokoszka now it's ready for review!
I added the refactor requested here https://github.com/scylladb/scylla-manager/pull/3918#discussion_r1670905959.

pkg/service/backup/service_backup_integration_test.go
```diff
@@ -2496,3 +2496,86 @@ func TestBackupAlternatorIntegration(t *testing.T) {
...
+				}
+			}
+		}
+	}
+
+	Print("And: validate base table in backup files")
+	if foundTable != len(ManagedClusterHosts()) {
+		t.Fatal("Expected all hosts to back up base table")
+	}
+}
```
karol-kokoszka, line 2581: Take a look at this very nice explanation of what given-when-then is -> https://martinfowler.com/bliki/GivenWhenThen.html

Your prints doesn't follow this pattern.
Please add print with Given -> current state
When -> What action you did
Then -> What is the expected state

pkg/service/backup/service.go
```diff
@@ -117,215 +118,92 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
 	s.logger.Info(ctx, "Generating backup target", "cluster_id", clusterID)
```
karol-kokoszka, line 119: This comment refers to the function name.
`GetTarget` is not sufficient here. First of all, it indicates that the method is kind of simple getter (what is not true).
Besides, even if it would be a simple getter, idiomatic go suggests to name getter this way https://go.dev/doc/effective_go#Getters (Target() instead of GetTarget()) .

Please change the name of the method (on the interface level) to sth like `ConvertRunnerPropertiesToTarget()`

karol-kokoszka (approved):
Very nice ! 
Thank you for the effort.
:+1:  but please consider two additional comments I left.

pkg/service/backup/service.go
Michal-Leszczynski:
```go
// BackupService service interface for the REST API handlers.
type BackupService interface {
	GetTarget(ctx context.Context, clusterID uuid.UUID, properties json.RawMessage) (backup.Target, error)
```
I would say that this method is called `GetTarget` because it is the base for the handler of:
```go
m.Get("/{task_type}/target", h.getTarget)
```
and it seems like this is the convention used in all services interfaces:
```go
// ClusterService service interface for the REST API handlers.
type ClusterService interface {
	ListClusters(ctx context.Context, f *cluster.Filter) ([]*cluster.Cluster, error)
	GetCluster(ctx context.Context, idOrName string) (*cluster.Cluster, error)
	PutCluster(ctx context.Context, c *cluster.Cluster) error
	DeleteCluster(ctx context.Context, id uuid.UUID) error
```
@karol-kokoszka do you want me to rename all of them?

pkg/service/backup/service_backup_integration_test.go
Michal-Leszczynski: done

pkg/service/backup/service.go
karol-kokoszka: Ohhh, so the service interface defines HTTP API.
Please leave the GetTarget, but build another method `ConvertRunnerPropertiesToTarget` that returns the target.
Let GetTarget call this new method.

karol-kokoszka (approved):

Michal-Leszczynski:
@annastuchlik could you take a look at a8743d582b780d5c3a98d1ce4aed4230e3702452?

PR #3926, opened 2024-07-10
Don't wrap errors as plain strings

### LINKED ISSUE #3925: Context cancel returned from location check does not result in paused task status
Discovered in this [SCT run](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/ilya/job/manager-3.3-2024.2/job/ubuntu22-sanity-ipv6-test/2/) - it tries to suspend a running backup task, but the task ends with status ERROR instead of status STOPPED.

The root cause is that errors returned from location check are not embedded, but plain pasted as strings:
```go
func (s *Service) checkHostLocation(ctx context.Context, client *scyllaclient.Client, h string, l Location) error {
	err := client.RcloneCheckPermissions(ctx, h, l.RemotePath(""))
	if err != nil {
		s.logger.Info(ctx, "Location check FAILED", "host", h, "location", l, "error", err)

		if scyllaclient.StatusCodeOf(err) > 0 {
			tip := fmt.Sprintf("make sure the location is correct and credentials are set, to debug SSH to %s and run \"scylla-manager-agent check-location -L %s --debug\"", h, l)
			err = errors.Errorf("%s: %s - %s", h, err, tip) // <- here
		} else {
			err = errors.Errorf("%s: %s", h, err) // <- here
		}
```

This results in assigning status ERROR in scheduler error checking (which is also not ideal because of https://github.com/scylladb/scylla-manager/issues/3884) :
```go
func statusFromError(err error) Status {
	switch {
	case err == nil:
		return StatusDone
	case errors.Is(err, context.Canceled): // <- returns false
		return StatusStopped
	case errors.Is(err, context.DeadlineExceeded):
		return StatusWaiting
	default:
		return StatusError
	}
}
```

cc: @karol-kokoszka @rayakurl @ilya-rarov

Michal-Leszczynski:
Using `errors.Errorf("%s", err)` results in making it impossible to later check nested error cause with `errors.Is` function.
On the other hand, it works fine with `fmt.Errorf("%w", err)`.

Fixes #3925

### DIFF
--- a/pkg/scyllaclient/status.go
+++ b/pkg/scyllaclient/status.go
@@ -81,7 +81,7 @@ func makeAgentError(resp *http.Response) error {
 		payload: new(agentModels.ErrorResponse),
 	}
 	if err := json.Unmarshal(b, ae.payload); err != nil {
-		return errors.Errorf("agent [HTTP %d] cannot read response: %s", resp.StatusCode, err)
+		return fmt.Errorf("agent [HTTP %d] cannot read response: %w", resp.StatusCode, err)
 	}
 
 	return ae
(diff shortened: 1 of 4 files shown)


Michal-Leszczynski:

karol-kokoszka (approved):
nice catch !

PR #3931, opened 2024-07-16
Speed up repair test

### LINKED ISSUE #3929: Repair integration tests can take even up to 69 minutes (!)
After some recent merge, it appears that repair integration tests can take even up to 69 minutes.
See https://github.com/scylladb/scylla-manager/actions/runs/9935863712/job/27442900589

It happened on 6.0.1

integration-tests-6.0.1-IPV4-tablets / Test repair (push) Successful in 69m
[Details](https://github.com/scylladb/scylla-manager/actions/runs/9935863712/job/27442900589)

integration-tests-6.0.1-IPV6-tablets / Test repair (push) Successful in 57m
[Details](https://github.com/scylladb/scylla-manager/actions/runs/9935863707/job/27442998790)

Other env looks better - more realistic, although performance seems to be degragated still 
integration-tests-2024.1.5-IPV4 / Test repair (push) Successful in 17m
[Details](https://github.com/scylladb/scylla-manager/actions/runs/9935863698/job/27442895401)

integration-tests-2024.1.5-IPV4-tablets / Test repair (push) Successful in 17m
[Details](https://github.com/scylladb/scylla-manager/actions/runs/9935863713/job/27442900919)

integration-tests-2024.1.5-IPV6-tablets / Test repair (push) Successful in 16m
[Details](https://github.com/scylladb/scylla-manager/actions/runs/9935863697/job/27442901746)

integration-tests-6.0.1-IPV4 / Test repair (push) Successful in 11m
[Details](https://github.com/scylladb/scylla-manager/actions/runs/9935863704/job/27442905724)

It must be checked and understood.

Michal-Leszczynski:
Helps with #3929, but it is not a fix for that.


pkg/service/repair/service_repair_integration_test.go
```diff
@@ -902,9 +902,9 @@ func TestServiceRepairResumeAllRangesIntegration(t *testing.T) {
 	)
 
 	// Create keyspaces. Low RF increases repair parallelism.
-	tryCreateTabletKeyspace(t, clusterSession, ks1, 2, 1, 4096)
-	tryCreateTabletKeyspace(t, clusterSession, ks2, 1, 1, 4096)
-	tryCreateTabletKeyspace(t, clusterSession, ks3, 1, 1, 4096)
+	tryCreateTabletKeyspace(t, clusterSession, ks1, 2, 1, 256)
+	tryCreateTabletKeyspace(t, clusterSession, ks2, 1, 1, 256)
+	tryCreateTabletKeyspace(t, clusterSession, ks3, 1, 1, 256)
```
karol-kokoszka, line 907: That's the fix for the repair speed ?

Michal-Leszczynski: Yes - this results in 2 things:
- this test tests retries - they don't use batching and our docker env has max_intensity=1, so every range is repaired in a separate repair job - 4096 ranges was way too big of a number
- as we repair less ranges, we do less inserts to SM DB - which makes it more responsive

karol-kokoszka: @Michal-Leszczynski but isn't it that this value defines number of tablets ? And the number of tablets is higher than number of token ranges available on a single node as the token range contains some amount of tablets ?

If so, then 4096 makes sense.
I'm just starting to thing that we should revisit the way how the state is kept and how SM updates it.
Maybe it should be sharded by extending the key ? Like include some ranges into the key ?

Michal-Leszczynski: Tablet to token range mapping is somewhat a fake concept - it was done so that SM can continue on repairing tablet tables like it does with vnode tables, but I don't think that it makes a lot of sense outside of this use case.

Perhaps the low memory of SM DB is the problem:
```yaml
  scylla-manager-db:
    command: --smp 1 --memory 500M 
```
As the regular cluster nodes receive 4 times the memory: ``SCYLLA_ARGS := --smp 2 --memory 2G``.

I agree that the way we store repair state/progress in SM memory rather bad.
The way in which we update state and calculate progress display is also not the best.

Perhaps it would be good to ask someone, what type of progress is useful to aggregate, like is per host progress important if we repair table by table anyway?

Michal-Leszczynski: But the small change introduced in this PR reduces integration tests flakiness, so maybe it's worth to merge it first and deal with the restructurization of repair progress/state in a separate PR?

karol-kokoszka: OK, we know that the way how the repair progress is tracked by SM is bit of problematic. You can change the PR to ready for a review.

karol-kokoszka (approved):
:+1:

PR #3934, opened 2024-07-18
Label tasks and clusters

### LINKED ISSUE #3219: Give possibility for labeling the clusters added to SM

Michal-Leszczynski:
This PR adds a possibility of labeling cluster/tasks with the `--label` flag.
I decided to implement the [approach with a single flag](https://github.com/scylladb/scylla-manager/issues/3219#issuecomment-2231385713), which has the special syntax for defining labels to be removed (similar to the [kubctl label](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)).

Fixes  #3219

### DIFF
--- a/docs/source/sctool/partials/sctool_suspend.yaml
+++ b/docs/source/sctool/partials/sctool_suspend.yaml
@@ -47,9 +47,18 @@ options:
 
         The task run date is aligned with '--start date' value.
         For example, if you select '--interval 7d' task would run weekly at the '--start-date' time.
+    - name: label
+      usage: |
+        A comma-separated list of label modifications. Labels are represented as a key-value store.
+        Character '=' has a special meaning and cannot be a part of label's key nor value.
+        A single modification takes form of:
+        * '<key>=<value>' - sets the label <key> to <value>
+        * '<key>-'        - removes the label
+
+        For example, '--label k1=v1,k2-' will set the label 'k1' to 'v1' and will also remove label 'k2'.
     - name: name
       usage: |
-        Task name that can be used insead of ID.
+        Task name that can be used instead of ID.
     - name: num-retries
       shorthand: r
       default_value: "3"
--- a/pkg/command/cluster/clusteradd/cmd.go
+++ b/pkg/command/cluster/clusteradd/cmd.go
@@ -25,6 +25,7 @@ type command struct {
 
 	id                     string
 	name                   string
+	label                  flag.Label
 	host                   string
 	port                   int64
 	authToken              string
@@ -57,6 +58,7 @@ func (cmd *command) init() {
 	w := cmd.Flags()
 	w.StringVarP(&cmd.id, "id", "i", "", "")
 	w.StringVarP(&cmd.name, "name", "n", "", "")
+	w.Var(&cmd.label, "label", "")
 	w.StringVar(&cmd.host, "host", "", "")
 	w.Int64Var(&cmd.port, "port", 10001, "")
 	w.StringVar(&cmd.authToken, "auth-token", "", "")
@@ -85,6 +87,7 @@ func (cmd *command) run() error {
 	c := &managerclient.Cluster{
 		ID:                     cmd.id,
 		Name:                   cmd.name,
+		Labels:                 cmd.label.NewLabels(),
 		Host:                   cmd.host,
 		AuthToken:              cmd.authToken,
 		Username:               cmd.username,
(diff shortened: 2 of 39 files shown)


Michal-Leszczynski:

@annastuchlik could you take a look just at the last commit a693d7d4b9f43cf993c339a4731987739b05f7ed which consists of the docs related changes?

pkg/command/flag/type.go
```diff
@@ -16,6 +18,97 @@ import (
...
+		return nil
+	}
+	stringReader := strings.NewReader(s)
+	csvReader := csv.NewReader(stringReader)
+	spec, err := csvReader.Read()
+	if err != nil {
+		return fmt.Errorf("parse csv label list: %w", err)
+	}
+	// Taken from https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/label/label.go#L411
+	add := map[string]string{}
```
karol-kokoszka, line 56: Why not to initialize l.add here and work on that map instead of creating new var and just assign it to l.add at the end
```
l.add = make(map[string]string)
```

?

pkg/command/cluster/clusteradd/cmd.go
```diff
@@ -85,6 +87,7 @@ func (cmd *command) run() error {
 	c := &managerclient.Cluster{
 		ID:                     cmd.id,
 		Name:                   cmd.name,
+		Labels:                 cmd.label.ApplyDiff(nil),
```
karol-kokoszka, line 90: Why not to create method ToMap() on label directly ?
I think it's more intuitive than applying a diff of nil to convert Label to map.

pkg/command/cluster/clusteradd/res.yaml
```diff
@@ -28,6 +28,15 @@ name: |
...
   This `alias` name can be used with all commands that accept --cluster parameter.
 
+label: |
+  A comma-separated list of label modifications. Labels are represented as a key-value store.
+  Character '=' has a special meaning and cannot be a part of label's key nor value.
+  A single modification takes form of:
+  * '<key>=<value>' - sets the label <key> to <value>
+  * '<key>-'        - removes the label
+  
+  For example, ``--label k1=v1,k2-`` will set the label 'k1' to 'v1' and will also remove label 'k2'.
```
karol-kokoszka, line 38: not true

pkg/command/cluster/clusterupdate/cmd_integration_api_test.go
```diff
@@ -60,6 +61,17 @@ func TestSctoolClusterUpdateIntegrationAPITest(t *testing.T) {
...
+		{
+			name: "update cluster, add labels",
+			args: []string{"cluster", "update", "--label", "k1=v1,k2=v2"},
+			expectedCluster: &models.Cluster{
+				Labels: map[string]string{
+					"k1": "v1",
+					"k2": "v2",
+				},
+				ForceTLSDisabled: true,
+			},
```
karol-kokoszka, line 73: Please change the initial state of the cluster lines 82-88 and add some label here, and change the args the way that one of the labels is removed.

karol-kokoszka (approved):
Nice ! Few small comments.

Michal-Leszczynski:
@karol-kokoszka This PR is ready for re-review!

docs/source/sctool/partials/sctool_suspend.yaml
```diff
@@ -47,9 +47,18 @@ options:
...
         For example, if you select '--interval 7d' task would run weekly at the '--start-date' time.
+    - name: label
+      usage: |
+        A comma-separated list of label modifications. Labels are represented as a key-value store.
+        Character '=' has a special meaning and cannot be a part of label's key nor value.
+        A single modification takes form of:
+        * '<key>=<value>' - sets the label <key> to <value>
+        * '<key>-'        - removes the label
+
+        For example, '--label k1=v1,k2-' will set the label 'k1' to 'v1' and will also remove label 'k2'.
```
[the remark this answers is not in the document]
karol-kokoszka: I'm fine with doing it in separate PR, but are you sure it require changing command struct ? Comment refers to docs only.

pkg/command/flag/type.go
karol-kokoszka: OK, let's leave it as is.

karol-kokoszka:
@Michal-Leszczynski Thanks! Good to go. It's approved already.

Michal-Leszczynski:
@annastuchlik could you take a look at dc9ce852f68b8b7c2fd47b2512b2a1d4d34e3d1e?

PR #3938, opened 2024-07-23
Remove ansible/restore

mykaul:
Since it's no longer supported, remove it.

---

Please make sure that:


karol-kokoszka (approved):
:+1:

PR #3941, opened 2024-07-24
Delete files in batches

### LINKED ISSUE #3928: Delete files in batches
As suggested in this [comment](https://github.com/scylladb/scylla-manager/pull/3885#pullrequestreview-2162581748), it would be useful to extend rclone so that it's possible to delete files in batches.
This could massively improve backup purge stage and the new [deduplication stage](https://github.com/scylladb/scylla-manager/issues/3827).

### LINKED ISSUE #3945: Incremental backup (no changes in database between backups) failed on backup.deduplicate step
**Cluster configuration:**
- 3-nodes singleDC cluster;
- Dataset: 10 keyspaces with 100 empty tables per each keyspace + keyspace with one table (4000000 entries)

**Steps:**
1. Run the full backup:
`sctool backup -c 45bd62ca-f6ae-4d78-a753-70cd989165b9 --location s3:manager-backup-tests-us-east-1`
2. Wait for the process finishes successfully;
3. Don't do any changes to database;
4. Run the backup one more time:
`sctool backup -c 45bd62ca-f6ae-4d78-a753-70cd989165b9 --location s3:manager-backup-tests-us-east-1`
5. Monitor backup task status

**Actual result:**
When the backup comes to deduplication step, the process fails with the ERROR:
```
Jul 29 11:00:38 manager-regression-mikita-monitor-node-a33578c5-1 scylla-manager[11138]: {"L":"ERROR","T":"2024-07-29T11:00:38.713Z","N":"backup.deduplicate","M":"Failed to deduplicate host","host":"10.12.2.171","keyspace":"system_schema","table":"triggers","error":"deduplication based on .crc32 content: cannot get content of local CRC32 data:system_schema/triggers-4df70b666b05325195a132b54005fd48/snapshots/sm_20240729110033UTC/me-3gi8_0pfy_5gbwg26n6x6awp2eps-big-Digest.crc32: agent [HTTP 404] object not found","_trace_id":"3Mpl_MK7QtyBHAjdx3U9RA","errorStack":"github.com/scylladb/scylla-manager/v3/pkg/service/backup.(*worker).deduplicateHost.func2\n\tgithub.com/scylladb/scylla-manager/v3/pkg/service/backup/worker_deduplicate.go:105\ngithub.com/scylladb/scylla-manager/v3/pkg/util/parallel.Run.func1\n\tgithub.com/scylladb/scylla-manager/v3/pkg/util/parallel/parallel.go:72\nruntime.goexit\n\truntime/asm_amd64.s:1695\n","S":"github.com/scylladb/go-log.Logger.log\n\tgithub.com/scylladb/go-log@v0.0.7/logger.go:101\ngithub.com/scylladb/go-log.Logger.Error\n\tgithub.com/scylladb/go-log@v0.0.7/logger.go:84\ngithub.com/scylladb/scylla-manager/v3/pkg/service/backup.(*worker).deduplicateHost.func3\n\tgithub.com/scylladb/scylla-manager/v3/pkg/service/backup/worker_deduplicate.go:114\ngithub.com/scylladb/scylla-manager/v3/pkg/util/parallel.Run.func1\n\tgithub.com/scylladb/scylla-manager/v3/pkg/util/parallel/parallel.go:79"}
```
Backup task info:
```
ubuntu@manager-regression-mikita-monitor-node-a33578c5-1:~$ sctool info backup/de67765c-ce6c-4aed-9c4d-edee0a5b21a5 -c 45bd62ca-f6ae-4d78-a753-70cd989165b9
Name:	backup/de67765c-ce6c-4aed-9c4d-edee0a5b21a5
Cron:   {"spec":"","start_date":"0001-01-01T00:00:00Z"} (next activation date: 29 Jul 24 11:11:06 UTC)
Tz:	Etc/UTC
Retry:	3 (initial backoff 10m)
Properties:
- location: 's3:manager-backup-tests-us-east-1'

╭──────────────────────────────────────┬────────────────────────┬───
───────┬────────╮
│ ID                                   │ Start time             │ Duration │ Status │
├──────────────────────────────────────┼────────────────────────┼──────────┼────────┤
│ c6d289c9-4d99-11ef-a1a4-02560c89a57d │ 29 Jul 24 11:00:33 UTC │ 33s      │ ERROR  │
╰──────────────────────────────────────┴────────────────────────┴──────────┴────────╯
```

**Expected result:**
The backup finishes successfully.

**Environment:**
- scylla-manager-server version: 3.3.0~0.20240727.8f6f1c0b8-SNAPSHOT;

**Logs:**
scylla-manager logs - [scylla-manager.log](https://github.com/user-attachments/files/16412226/scylla-manager.log)

Michal-Leszczynski:
This PR adds a way to delete files in batches via RcloneDeletePaths/RcloneDeletePathsInBatches client method and uses it during the backup purge and deduplicate stage.

Fixes #3945
Fixes #3928


pkg/service/backup/purger.go
```diff
@@ -379,44 +379,33 @@ func (p purger) deleteFiles(ctx context.Context, location Location, files fileSe
...
 	// maxParallelDirs is aligned with MaxIdleConnsPerHost in scyllaclient.DefaultTransport.
 	const maxParallelDirs = 100
 
 	f := func(i int) error {
-		var (
-			dir  = dirs[i]
-			rerr error
-		)
+		dir := dirs[i]
+		const maxBatchSize = 1000
```
karol-kokoszka, line 388: How did you choose this value of 1000 ?

karol-kokoszka (commented):
Nice. So this PR just adds support for deleting files in batches + updates purger to use this approach instead of removing them one by one.

I wonder what RClone does under the hood, how the batch is handled. Maybe it's just split per files listed in filter and deleted one by one still.
It's one API call instead of dozens, so the improvement is good.

@Michal-Leszczynski please validate the branch with SCT and make sure that some of the test is actually deleting old snapshot.
If we don't test the deleting scenario, then we need to include @mikliapko to add this test case.

Michal-Leszczynski:
> I wonder what RClone does under the hood and how it handles batches. Maybe it just splits the files listed in the filter and deletes them one by one. However, it's still one API call instead of dozens, so the improvement is significant.

Rclone indeed deletes files one by one, but it does so in parallel.
The parallelization is controlled with the same config option as the upload/download (`Transfers` config field). 
Since deleting files shouldn't use the buffer memory pool, maybe we could set this config option to a higher value?
This would be done only for the context of given delete batch of files call, so it shouldn't impact other things.

pkg/service/backup/purger.go
Michal-Leszczynski: I wanted to limit batch size, so that we don't send request with millions of file name, as it could create problems with transport and timeouts. 1000 seems like it's small enough not to negatively impact the transport, and deleting 1000 files shouldn't take too long as well. What do you think about this number? 

On the other hand, I could make this API call asynchronous and use long polling for checking it's progress.
This would resolve the problem with timeout, but it would still be better not to delete all the files in a single batch anyway.

karol-kokoszka:
> Since deleting files shouldn't use the buffer memory pool, maybe we could set this config option to a higher value?
This would be done only for the context of given delete batch of files call, so it shouldn't impact other things.

I guess it could be set to higher value, but as RClone is doing deletes one by one (in parallel), not via batch delete request directly on cloud provider, the request can be throttled by S3 (or google storage) if there are too many.
The bit of the challenge (or maybe not ?) would to find the sweet spot for number of parallel deletes https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html

pkg/scyllaclient/client_rclone_agent_integration_test.go
```diff
@@ -44,6 +46,134 @@ func TestRcloneS3ListDirAgentIntegration(t *testing.T) {
...
+			after:     []string{"b", "d"},
+			batchSize: 2,
+		},
+		{
+			name:      "delete thousands of files",
+			before:    append([]string{"a"}, lotsOfFiles...),
+			delete:    append([]string{}, lotsOfFiles...),
+			after:     []string{"a"},
+			batchSize: 1000,
+		},
```
karol-kokoszka, line 102: Thanks ! Looks good,
Please add 'delete nothing' edge case.
```
		{
			name:      "delete nothing",
			before:    []string{"a", "b", "c"},
			delete:    []string{},
			after:     []string{"a", "b", "c"},
			batchSize: 1,
		},
```
It PASSES according to my check done locally.

karol-kokoszka (approved):
:+1: just one nit

pkg/scyllaclient/client_rclone_agent_integration_test.go
Michal-Leszczynski: It's actually partially by mistake.
`RcloneDeletePaths` returns error on empty delete (because rclone implementation that I wrote returns an error for safety), but `RcloneDeletePathsInBatches` does not (it doesn't send any request to rclone server).
So this is somewhat of a gray area, but I can add this case to document this behavior.

PR #3943, opened 2024-07-25
cluster: update to include all hosts into connection validity check

### LINKED ISSUE #3874: Initial host of the cluster may change
Scylla Manager keep 'host' as a property of cluster in the database.
This `host` is coordinator node that is considered to be used as a first node when SM creates the scylla client to interact with Scylla API.
If the `host` value is not correct, then the only possibility of updating this value is by calling `sctool cluster update --host <new_host>`.

It can be improved, by letting SM to override the `host` value with one of the existing nodes when the SM realizes that the `host` doesn't exist anymore.

karol-kokoszka:
Fixes #3874 

---

Please make sure that:

### DIFF
--- a/pkg/service/cluster/service.go
+++ b/pkg/service/cluster/service.go
@@ -33,6 +33,9 @@ type ProviderFunc func(ctx context.Context, id uuid.UUID) (*Cluster, error)
 // ChangeType specifies type on Change.
 type ChangeType int8
 
+// ErrNoValidKnownHost is thrown when it was not possible to connect to any of the currently known hosts of the cluster.
+var ErrNoValidKnownHost = errors.New("unable to connect to any of cluster's known hosts")
+
 // ChangeType enumeration.
 const (
 	Create ChangeType = iota
@@ -119,17 +122,6 @@ func (s *Service) Client(ctx context.Context, clusterID uuid.UUID) (*scyllaclien
 	return s.clientCache.Client(ctx, clusterID)
 }
 
-func (s *Service) createClientNoValidation(c *Cluster) (*scyllaclient.Client, error) {
-	config := scyllaclient.DefaultConfigWithTimeout(s.timeoutConfig)
-	config.Hosts = c.KnownHosts
-	if c.Port != 0 {
-		config.Port = fmt.Sprint(c.Port)
-	}
-	config.AuthToken = c.AuthToken
-
-	return scyllaclient.NewClient(config, s.logger.Named("client"))
-}
-
 // CreateClientNoCache creates Scylla API that load balances calls to every node from given cluster.
 // There may be a situation that cluster keeps outdated information about list of available hosts.
 // To work it around:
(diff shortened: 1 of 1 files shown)


Michal-Leszczynski (approved):

Michal-Leszczynski (changes requested):

pkg/service/cluster/service.go
```diff
@@ -462,42 +464,38 @@ func (s *Service) PutCluster(ctx context.Context, c *Cluster) (err error) {
...
+		return err
 	}
+	defer logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")
 
-	// For every live host check there are no errors
 	var errs error
-	for i, err := range client.CheckHostsConnectivity(ctx, live) {
-		errs = multierr.Append(errs, errors.Wrap(err, live[i]))
+	for i, err := range client.CheckHostsConnectivity(ctx, knwonHosts) {
+		errs = multierr.Append(errs, errors.Wrap(err, knwonHosts[i]))
```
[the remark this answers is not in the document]
karol-kokoszka: Yes, it makes sense to bring it back actually.
Done.

pkg/service/cluster/service.go
```diff
@@ -462,42 +464,44 @@ func (s *Service) PutCluster(ctx context.Context, c *Cluster) (err error) {
...
-	} else if err := s.loadKnownHosts(c); err != nil {
+	if err := s.loadKnownHosts(c); err != nil && !errors.Is(err, gocql.ErrNotFound) {
 		return errors.Wrap(err, "load known hosts")
 	}
+	if c.Host != "" {
+		c.KnownHosts = append([]string{c.Host}, c.KnownHosts...)
+	}
 
-	client, err := s.createClientNoValidation(c)
+	knwonHosts, err := s.filterKnownHosts(ctx, c)
```
Michal-Leszczynski, line 474: knwonHosts -> knownHosts

pkg/service/cluster/service.go
```diff
@@ -462,42 +464,44 @@ func (s *Service) PutCluster(ctx context.Context, c *Cluster) (err error) {
...
+	config := s.clientConfig(c)
+	client, err := scyllaclient.NewClient(config, s.logger.Named("client"))
 	if err != nil {
-		return errors.Wrap(err, "status")
+		return err
 	}
+	defer logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")
 
-	// Get live hosts
+	status, err := client.Status(ctx)
```
Michal-Leszczynski, line 492: err is ignored

pkg/service/cluster/service.go
```diff
@@ -462,42 +464,44 @@ func (s *Service) PutCluster(ctx context.Context, c *Cluster) (err error) {
...
 	var errs error
 	for i, err := range client.CheckHostsConnectivity(ctx, live) {
 		errs = multierr.Append(errs, errors.Wrap(err, live[i]))
 	}
 	if errs != nil {
 		return service.ErrValidate(errors.Wrap(errs, "connectivity check"))
 	}
-
-	// Update known hosts.
-	c.KnownHosts = live
```
Michal-Leszczynski, line 499: So now we set known hosts to all known hosts, not only to the live ones.
This could result in some additional error logs, but client should be able to handle down nodes in some way.

karol-kokoszka:
I didn't push my recent changes. Sorry.

pkg/service/cluster/service.go
karol-kokoszka: known hosts are the ones discovered with the API call, not the ones that are live only

pkg/service/cluster/service.go
karol-kokoszka: fixed

Michal-Leszczynski:
@karol-kokoszka restore test flakiness is fixed in commit 79068b16e18f4bd1f50c437a7b6aa81877904080 in PR #3941.

Michal-Leszczynski:
@karol-kokoszka can we merge this PR? Those changes influence https://github.com/scylladb/scylla-manager/issues/3935 which I am working on.

karol-kokoszka:
Done

PR #3956, opened 2024-08-07
Disable compaction during restore

### LINKED ISSUE #3953: Disable compaction during restore
As observed in benchmarks, restore (mainly l&s) slows down noticeably when the compaction is running.
The compactions should be disabled during restore and re-enabled afterwards (similar to `tombstone_gc`).
image
image

Michal-Leszczynski:
Fixes #3953

### DIFF
--- a/pkg/managerclient/restore.go
+++ b/pkg/managerclient/restore.go
@@ -6,10 +6,12 @@ package managerclient
 const (
 	RestoreStageInit          = "INIT"
 	RestoreStageDropViews     = "DROP_VIEWS"
+	StageDisableCompaction    = "DISABLE_COMPACTION"
 	RestoreStageDisableTGC    = "DISABLE_TGC"
 	RestoreStageData          = "DATA"
 	RestoreStageRepair        = "REPAIR"
 	RestoreStageEnableTG      = "ENABLE_TGC"
+	StageEnableCompaction     = "ENABLE_COMPACTION"
 	RestoreStageRecreateViews = "RECREATE_VIEWS"
 	RestoreStageDone          = "DONE"
 )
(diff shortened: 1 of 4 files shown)


Michal-Leszczynski:

pkg/managerclient/restore.go
```diff
@@ -6,21 +6,25 @@ package managerclient
 const (
 	RestoreStageInit          = "INIT"
 	RestoreStageDropViews     = "DROP_VIEWS"
+	StageDisableCompaction    = "DISABLE_COMPACTION"
```
karol-kokoszka, line 9: Please follow the same pattern by either removing `Restore` prefix from all the stages, or adding prefix `Restore` to the new stages.

pkg/service/restore/restore_integration_test.go
```diff
@@ -334,3 +338,103 @@ func TestRestoreTablesVnodeToTabletsIntegration(t *testing.T) {
...
+				close(reachedDataStageChan)
+			}
+			<-waitCopyPaths
+		}
+		return nil, nil
+	}))
+
+	Print("Run restore")
+	wg := sync.WaitGroup{}
+	wg.Add(1)
```
karol-kokoszka, line 393: Please use one approach for syncing the code.
Either sync.WaitGroup{} or synchronization by closing the channel.

For single goroutine, channel is ok.

karol-kokoszka (commented):
Two nits.

Michal-Leszczynski:
@karol-kokoszka I'm wondering if we should before merging this PR.
I don't know if we have the evidence that this is the right thing to do for any type of restore.
Perhaps it could be controlled by a new flag (e.g. `--disable-compaction`)?

Michal-Leszczynski:
Closed in favor of #4060.

PR #3986, opened 2024-08-20
Restore improvements

Michal-Leszczynski:
Based on #3956.

This PR introduces 3 big changes:
1. Indexing
Restore task no longer iterates over manifests, and restores all the files from given manifest before proceeding to the next one. 
The problem with this approach was that it limited the pool of sstables used for creating batches and it resulted in nodes spending more time in the idle state in between restored manifests.
Now, SM indexes all of the files from all manifests at the beginning, and then creates batches from all of them, without the need to finish one manifest before another.

2. Batching.
Restore task no longer batches files based on `--batch-size` flag.
The problem with this approach was that it didn't take shard_cnt or sstable size into consideration. It was also difficult to guess the right value for this flag without knowing backup statistics.
Now, SM creates batches containing multiple of node shard_cnt sstables. It keeps on adding the biggest (not yet restored) sstables to the batch until it reaches size of 5% of the expected node workload (total_workload/total_shard_cnt_in_the_cluster*node_shard_cnt).

3. Additional, testing flags.
Those flags are there just for the testing purposes. Some of them will be removed, some might stay, depending on how useful we find them:
- `--table-parallel` - how many tables should be restored from a single node at a single time (might be useful for many smaller tables)
- `--stream-to-all-replicas` - runs l&s without the primary_replica_only option and skips the post restore repair (might be useful in a cluster with small RF and big amount of replica sets)
- `--unpin-agent-cpu` - unpins agent from cpus for the time of the restore (it looks like it increases download speed to some extent)


karol-kokoszka:
@Michal-Leszczynski this is stalled PR that was used when experimenting with restore speed.
Do you still need it for some purpose or it can be closed ?

Michal-Leszczynski:
It can be closed - a subset of useful fixes included in this PR have already been merged to master.

PR #3997, opened 2024-08-26
docs: fix a typo

tchaikov:
replace "sheduled" with "scheduled"

---

Please make sure that:


Michal-Leszczynski:
Thanks for reporting!

PR #4008, opened 2024-08-29
Improve schema backup

### LINKED ISSUE #3995: Dont allow to backup without schema
For Scylla versions >= 6.0/2024.2 SM backups and restores schema from the output of DESC SCHEMA WITH INTERNALS cql query. This means that schema can be backed up as CQL file only when SM has CQL credentials to the cluster. This results in backups of Scylla 6.0/2024.2 not containing the schema required for restore purposes - SM allows for such backups to proceed because:
- in previous Scylla versions the CQL credentials weren't needed for the backup to finish
- there is a workaround to restore schema manually (SM still backs up schema sstables and they can be uploaded to a single node cluster in order to run DESC SCHEMA WITH INTERNALS on it)

This is messy, so SM shouldn't allow for a backup not containing proper schema. It should require CQL creds for Scylla versions >= 6.0/2024.2 for running the backup (in case user would really like to run a backup without CQL creds, we could add a flag like `--dont-backup-schema` that allows to omit schema backup when deliberately set by the user). 

https://github.com/scylladb/scylla-manager/issues/3831 could make things easier as it would allow for SM to obtain the output of DESC SCHEMA WITH INTERNALS without the need for CQL creds (when the maintenance socket is open).

cc: @karol-kokoszka @pdbossman

### LINKED ISSUE #4007: Don't assume that SM can connect to all hosts via CQL
Right now, when SM has CQL credentials to managed cluster, it assumes that it can connect to any of theses nodes.
There are specific cluster configurations (connected different tls certs for different dcs), where that's not the case.

For example, when SM tries to make a read barrier before dumping schema for the backup, it just takes a random node and tries to connect to it via CQL. It does not retry if it does not have permissions to connect to this node:
https://github.com/scylladb/scylla-manager/blob/153218165b75c8f8e11b0a64313b1655a383ef5b/pkg/service/backup/service.go#L703-L724

cc: @mykaul @pdbossman @nopzdk

Michal-Leszczynski:
This PR makes a several changes regarding schema backup.
They aim to increase schema backup safety by making it impossible to skip schema backup by simply not providing CQL creds (which was possible for Scylla version >= 6.0). Skipping schema backup is still possible (some users might not want to provide the CQL creds), but requires user to deliberately add the new `--skip-schema` flag. 
They also make the schema backup more robust for multi Scylla version cluster, or a cluster where SM does not have CQL access to all of the nodes.  

Fixes #4007
Fixes #3995

### DIFF
--- a/pkg/command/backup/cmd.go
+++ b/pkg/command/backup/cmd.go
@@ -37,6 +37,7 @@ type command struct {
 	dryRun           bool
 	showTables       bool
 	purgeOnly        bool
+	skipSchema       bool
 }
 
 func NewCommand(client *managerclient.Client) *cobra.Command {
@@ -88,6 +89,7 @@ func (cmd *command) init() {
 	w.Unwrap().BoolVar(&cmd.dryRun, "dry-run", false, "")
 	w.Unwrap().BoolVar(&cmd.showTables, "show-tables", false, "")
 	w.Unwrap().BoolVar(&cmd.purgeOnly, "purge-only", false, "")
+	w.Unwrap().BoolVar(&cmd.skipSchema, "skip-schema", false, "")
 }
 
 func (cmd *command) run(args []string) error {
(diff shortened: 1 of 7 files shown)


Michal-Leszczynski:

Michal-Leszczynski:
@karol-kokoszka I believe that this should also be included in 3.3.2 for safety reasons.

karol-kokoszka (approved):

PR #4010, opened 2024-08-30
Managerclient as submodule

### LINKED ISSUE #4012: Extract managerclient to separate submodule
By providing a way of simply importing managerclient as dependency we remove the need of importing the full scylla-manager repo with its dependencies to whatever project that wants to consume manager API.

Scylla Cloud is going to benefit out of it, as currently we cannot simply update the dependency to the latest manager due to broken dependency tree.

karol-kokoszka:
Fixes #4012 

The goal is to extract managerclient as a separate submodule, so that it can be imported to other projects simply as 
```
go get github.com/scylladb/scylla-manager/v3/pkg/managerclient
```
instead of importing full scylla-manager as dependency.
Doing the above is difficult if the repository uses google cloud SKD as a dependency, as our inhouse rclone (dependency to sm) uses replace directive on that lib to our inhouse google SDK....

Managerclient is intended to be used by Scylla Cloud.

**Note:**
Modules are named with v3/ prefix, to keep the backward compatibility with the imports.
v3/ prefix is added to the new path, as it must match the module name.

---

Please make sure that:


Michal-Leszczynski:
@karol-kokoszka I understand that we want to wait for https://github.com/scylladb/siren/pull/11958 in order to validate correctness of this PR?

karol-kokoszka:
> @karol-kokoszka I understand that we want to wait for https://github.com/scylladb/siren/pull/11958 in order to validate correctness of this PR?

Yes, I want to import dependency to siren first and validate it.
I see that some enums are missing like types of tasks etc. But I would prefer to address it in other PR by extending the swagger definition.

karol-kokoszka:
Siren builds properly when the managerclient submodule is used as dependency only.
I started stable e2e suite to verify it bit deeper.

Anyway, wants to merge it after the 3.3.2 patch fix.
The current PR is not gonna be included into this quick 3.3.2 fixing backups for 1 node clusters.

karol-kokoszka:
Scylla Cloud e2e passed https://jenkins.scylladb.com/view/siren/job/siren-jobs/job/siren-backend-e2e-manual/5696/

Michal-Leszczynski (approved):
LGTM

PR #4021, opened 2024-09-05
Don't blacklist restored tables

### LINKED ISSUE #3998: Improve SM backup/restore table filtering
Blacklisting has its flaws (e.g. https://github.com/scylladb/scylla-enterprise/issues/4551).
SM should get information about:
- system/user keyspaces
- local/non-local keyspace

Directly from Scylla (and not try to assume it basing on some heuristics), and use it for more robust table filtering.

Michal-Leszczynski:
Blacklisting approach was error-prone and difficult to test (e.g. Scylla cluster might have some leftover cdc tables from previous releases that are not present in the fresh clusters).
It also allows for a more unified mechanism of excluding raft managed tables (replicated locally) from being restored.

Rebased on #4034.

Fixes #3998


Michal-Leszczynski:

pkg/scyllaclient/client_scylla.go
```diff
@@ -314,6 +314,22 @@ func (c *Client) KeyspacesByType(ctx context.Context) (map[KeyspaceType][]string
...
 
+// AllTables returns all tables grouped by keyspace.
+func (c *Client) AllTables(ctx context.Context) (map[string][]string, error) {
+	resp, err := c.scyllaOps.ColumnFamilyNameGet(&operations.ColumnFamilyNameGetParams{Context: ctx})
+	if err != nil {
+		return nil, err
+	}
+	out := make(map[string][]string)
+	for _, kst := range resp.Payload {
+		parts := strings.Split(kst, ":")
```
karol-kokoszka, line 325: Could you please add simple validation checking if the len(parts) > 1 ?
It's to avoid the situation that scylla manager panics when (somehow) the response is not ...:...

pkg/service/restore/worker.go
```diff
@@ -183,6 +153,74 @@ func (w *worker) initTarget(ctx context.Context, properties json.RawMessage) err
...
+
+	// Skip views
+	views, err := query.GetAllViews(session)
+	if err != nil {
+		return nil, errors.Wrap(err, "get cluster views")
+	}
+	skip = append(skip, views.List()...)
+
+	// Exclude collected patterns
+	var out []string
```
karol-kokoszka, line 217: You can preallocate the size of out as it's equal to size of skip.
```
        out := make([]string, len(skipPatterns))
	for i, pattern := range skipPatterns {
		out[i] = "!" + pattern
	}
```

karol-kokoszka (approved):

PR #4024, opened 2024-09-05
Don't rely on ring replication strategy in repair

### LINKED ISSUE #4022: Don't rely on ring replication strategy in repair
Ref https://github.com/scylladb/scylla-manager/issues/3989#issuecomment-2309704442

Michal-Leszczynski:
Rebased on #4034.

Fixes #4022

### DIFF
--- a/pkg/service/repair/repair_test.go
+++ b/pkg/service/repair/repair_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/scylladb/scylla-manager/v3/pkg/dht"
 	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
 	"github.com/scylladb/scylla-manager/v3/pkg/service/repair"
 )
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:

pkg/service/repair/repair_test.go
```diff
@@ -132,148 +147,142 @@ func TestShouldRepairRing(t *testing.T) {
...
 		},
 		{
 			Ring: scyllaclient.Ring{
-				HostDC:      hostDC,
-				Replication: scyllaclient.SimpleStrategy,
-				RF:          2,
+				ReplicaTokens: []scyllaclient.ReplicaTokenRanges{
+					rs("h1", "h2"), rs("h2", "h3"), rs("h3", "h4"),
+					rs("h4", "h5"), rs("h5", "h6"), rs("h6", "h7"),
+					rs("h7", "h8"), rs("h8", "h9"), rs("h9", "h1"),
```
karol-kokoszka, line 173: Is this even possible scenario ? 
Replica sets here are of the same table.

rs("h2", "h3") means that data is replicated in dc2 (that may have RF=1) and in dc3 on one node (what means that RF for this DC is 1 as well), the next rs is ("h3", "h4"), what means that data is on two nodes from dc3 (RF = 2).

karol-kokoszka (commented):
Can you add some comments to the test cases explaining the scenario and reasoning the output ?

pkg/service/repair/repair_test.go
Michal-Leszczynski: I basically just converted previous tests based on explicit replication strategy to token rings.
This test here tests for SimpleReplicationStrategy, so that's why replica count per DC is not preserved.

Michal-Leszczynski:
@karol-kokoszka I added requested explanations.

pkg/service/repair/repair_test.go
```diff
@@ -124,164 +125,185 @@ func TestShouldRepairRing(t *testing.T) {
...
+				},
+				HostDC: hostDC,
 			},
-			DCs:      []string{"dc1", "dc2", "dc3", "dc4"},
-			Expected: false,
+			DCs:      allDCs,
+			Expected: false, // can't repair rf=1
 		},
 		{
+			Name: "SimpleStrategy{rf=2}",
```
karol-kokoszka, line 172: But this is not true. The RF per DC is not 2 here.
Besides rf=2 is a very much shortcut here, as the replication factor is set per datacenter.
Please either remove this test-case or provide realistic replica sets for this test.

pkg/service/repair/repair_test.go
```diff
@@ -124,164 +125,185 @@ func TestShouldRepairRing(t *testing.T) {
...
+				ReplicaTokens: []scyllaclient.ReplicaTokenRanges{rs("h1")},
+				HostDC:        hostDC,
 			},
-			DCs:      []string{"dc1", "dc2", "dc3", "dc4"},
-			Expected: false,
+			DCs:      allDCs,
+			Expected: false, // can't repair local keyspace
 		},
 		{
+			Name: "SimpleStrategy{rf=1}",
```
karol-kokoszka, line 159: Please explain RF per every single DC

karol-kokoszka (approved):
:+1:

PR #4029, opened 2024-09-10
Don't add cluster to SM DB before validation finishes

Michal-Leszczynski:
Commit https://github.com/scylladb/scylla-manager/commit/5bf6b35699c0645ba8598a53516e03ba7e01383f introduced a bug that when adding cluster to SM finished with error,
cluster was still saved in SM DB, but only with the ID and known hosts fields.
That's because method validateHostsConnectivity used discoverAndSetClusterHosts
instead of discoverClusterHosts, which happened before inserting cluster to SM DB.

In case client was mistakenly initialized with "" host, getting host from the pool would panic.
This should be replaced with a simple error. This PR adds additional validation to client config,
which prohibits specifying empty host there. It also makes host pool return an error instead of panic
in case of invalid host pool response.

This PR also includes testes for mentioned issues (validated that those tests weren't passing without the fixes).

Fixes #4028

PR #4033, opened 2024-09-11
Patch fix 3.3.3

karol-kokoszka:
Fixes regression https://github.com/scylladb/scylla-enterprise/issues/4692
https://github.com/scylladb/scylla-manager/pull/4029

---

Please make sure that:

### DIFF
--- a/pkg/scyllaclient/config.go
+++ b/pkg/scyllaclient/config.go
@@ -4,6 +4,7 @@ package scyllaclient
 
 import (
 	"net/http"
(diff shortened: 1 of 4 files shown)


karol-kokoszka:
@Michal-Leszczynski can you check operator using this branch as well ?

Michal-Leszczynski:
Operator tests started here: https://github.com/scylladb/scylla-operator/pull/2105#issuecomment-2343777226.

PR #4040, opened 2024-09-23
Restore: rely on run.units instead of target.keyspace

### LINKED ISSUE #4037: Interrupting restore targeting view tables is broken
This was the case in https://github.com/scylladb/scylla-enterprise/issues/4707.

View restoration happens by:
- saving view description in SM DB
- dropping the view from the cluster
- restoring view base table
- re-creating view from the saved description

SM re-calculates which tables should be ignored during restore at the beginning of each the task execution.
The problem is when SM restore task is interrupted after it already dropped the view tables.
SM has no way to difference view table from a regular one except for querying `system_schema.views` table, which won't contain information about already dropped views. Even though SM saved the views description in its DB, it does not use it (and it should) for checking if backed up table is a view that is restored in a different way.

Michal-Leszczynski:
Target.keyspace is calculated at the beginning of each restore task run.
It is vulnerable for changes occurring during restore process, e.g.:
- restore targeting some views is started
- target.keyspace excludes restored views (checked via CQL query)
- restore is paused
- restore is resumed
- recalculated target.keyspace no longer excludes restored views

Units are calculated only once at the beginning of the restore,
so they are not vulnerable to mentioned changes and should
be used to determine which tables should be restored.

Fixes #4037

### DIFF
--- a/pkg/service/restore/model.go
+++ b/pkg/service/restore/model.go
@@ -4,6 +4,7 @@ package restore
 
 import (
 	"reflect"
+	"slices"
 	"sort"
 	"time"
 
@@ -109,6 +110,14 @@ func (u Unit) MarshalUDT(name string, info gocql.TypeInfo) ([]byte, error) {
 func (u *Unit) UnmarshalUDT(name string, info gocql.TypeInfo, data []byte) error {
 	f := gocqlx.DefaultMapper.FieldByName(reflect.ValueOf(u), name)
 	return gocql.Unmarshal(info, data, f.Addr().Interface())
+}
+
+func unitsContainTable(units []Unit, ks, tab string) bool {
+	idx := slices.IndexFunc(units, func(u Unit) bool { return u.Keyspace == ks })
+	if idx < 0 {
+		return false
+	}
+	return slices.ContainsFunc(units[idx].Tables, func(t Table) bool { return t.Table == tab })
 }
 
 // Table represents restored table, its size and original tombstone_gc mode.
--- a/pkg/service/restore/restore_integration_test.go
+++ b/pkg/service/restore/restore_integration_test.go
@@ -7,17 +7,21 @@ package restore_test
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/pkg/errors"
+	"github.com/scylladb/scylla-manager/v3/pkg/service/backup"
 	. "github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
 	. "github.com/scylladb/scylla-manager/v3/pkg/testutils"
 	. "github.com/scylladb/scylla-manager/v3/pkg/testutils/db"
 	. "github.com/scylladb/scylla-manager/v3/pkg/testutils/testconfig"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/maputil"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/query"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
 )
 
 func TestRestoreTablesUserIntegration(t *testing.T) {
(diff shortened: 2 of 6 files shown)


Michal-Leszczynski:

Michal-Leszczynski:
@karol-kokoszka Could you take a look?

pkg/service/restore/restore_integration_test.go
```diff
@@ -334,3 +338,119 @@ func TestRestoreTablesVnodeToTabletsIntegration(t *testing.T) {
...
+		"restore_tables": true,
+	}
+	err = runPausedRestore(t, func(ctx context.Context) error {
+		h.dstCluster.RunID = uuid.NewTime()
+		rawProps, err := json.Marshal(props)
+		if err != nil {
+			return err
+		}
+		return h.dstRestoreSvc.Restore(ctx, h.dstCluster.ClusterID, h.dstCluster.TaskID, h.dstCluster.RunID, rawProps)
+	}, 45*time.Second, 2)
```
karol-kokoszka, line 446: How do you assure that the restore is paused after the view is removed ?
Maybe I'm wrong, but it looks bit random to me.

pkg/service/restore/model.go
```diff
@@ -111,6 +112,14 @@ func (u *Unit) UnmarshalUDT(name string, info gocql.TypeInfo, data []byte) error
 	return gocql.Unmarshal(info, data, f.Addr().Interface())
 }
 
+func unitsContainTable(units []Unit, ks, tab string) bool {
+	idx := slices.IndexFunc(units, func(u Unit) bool { return u.Keyspace == ks })
+	if idx < 0 {
+		return false
+	}
+	return slices.ContainsFunc(units[idx].Tables, func(t Table) bool { return t.Table == tab })
+}
```
karol-kokoszka, line 121: This is the part assuring that view SSTables are skip ?
It checks if the keyspace.table is in units passed to restore.
View is not included into the units, right ?

karol-kokoszka (commented):
Just a few questions.

pkg/service/restore/restore_integration_test.go
Michal-Leszczynski: I don't - the general idea is that this test should pause restore in random places, so not only when we set up some hook via client interceptor. This might give flaky false positives, but it increases the general coverage.

But since the restore is paused after 45 seconds and views are dropped at the beginning of the restore, it's really unlikely that the views won't be dropped before the first pause.

pkg/service/restore/model.go
Michal-Leszczynski: Units are stored in SM DB and are calculated only during the first run execution.
They don't contain the views, which are also stored in SM DB.

Michal-Leszczynski:
@karol-kokoszka The first round of comments has been addressed!

pkg/service/restore/restore_integration_test.go
karol-kokoszka: > I don't - the general idea is that this test should pause restore in random places, so not only when we set up some hook via client interceptor. 

But the origin for this change comes from the fact that the bug is found which root cause is:
- view is dropped 
- restore is paused
- restore is resumed
- Scylla Manager takes the SSTables of the view and restores them
Let the test to reproduce the issue explicitly.

> This might give flaky false positives, but it increases the general coverage.

We don't want flaky false positives.

We not to use interceptor on some calls or just use a test hook indicating that the view is drop ?

Michal-Leszczynski: I agree that we should explicitly test for pausing after views are dropped, but we somewhat need to introduce a flaky test pausing restore at random moments, as it improves coverage with places that we didn't think could cause problems with pause/resume. It won't cause CI to fail, but it might catch some issue from time to time.

pkg/service/restore/restore_integration_test.go
```diff
@@ -334,3 +338,119 @@ func TestRestoreTablesVnodeToTabletsIntegration(t *testing.T) {
 
 	validateTableContent[int, int](t, h.srcCluster.rootSession, h.dstCluster.rootSession, ks, tab, c1, c2)
 }
+
+func TestRestoreTablesPausedIntegration(t *testing.T) {
```
karol-kokoszka, line 342: nice !
I copied this test together with helpers to master branch, expecting that it is gonna fail with the same error that was met on prod env and it did :)
```
    restore_integration_test.go:448: create dir worker: get table version: not found
```

karol-kokoszka (approved):
:+1: thanks !

PR #4055, opened 2024-10-02
Restore improvement: batch

### LINKED ISSUE #3979: Batch sstables of similar size
Assume the following scenario: node with 8 shards, batch with 8 sstables, 1 is really big, 7 are small.
As only a single shard works on load&stream of given sstable, we would end up in a situation where the 7 shards taking care of small sstables would finish their work quickly and wait for the single shard working on a big sstable.
Creating batches of similarly sized sstables would improve load&stream performance.

### LINKED ISSUE #4059: Batch sstables basing on percentage of the node's data
As described in [this comment](https://github.com/scylladb/scylla-manager/issues/3979#issuecomment-2337585029), SM should make it easier to set optimal --batch-size for restoring as fast as possible.

In order to achieve that, we should add a special `--batch-size=0` value meaning that we want to create batches that result in the fastest restore. This follows the convention used by other flags like `--parallel` or `--rate-limit`.

Setting `--batch-size=0` would result in sending batches of unlimited amount of sstables, but of size up to 5% of expected node workload.

Michal-Leszczynski:
This PR introduces the following changes to batching mechanism:
- batches contain --batch-size*node_shard_count sstables instead of just --batch-size sstables (this was the initial design in[ restore 3.1 design doc](https://docs.google.com/document/d/1ugf6XCs28c4pwKsw9Lbu_EFb6ZSgMpJ_-Usf-Q9XAP4/edit#heading=h.f7jv2tbf2ofd))
- sorts sstables by decreasing size so that batches contain sstables of more similar size (#3979)
- added a special --batch-size=0 value which results in creating batches of not limited amount of sstables, but their total size is up to 5% of expected total node workload (#4059)

Fixes #3979
Fixes #4059

### DIFF
--- a/pkg/service/restore/batch.go
+++ b/pkg/service/restore/batch.go
@@ -28,6 +28,9 @@ func newBatchDispatcher(workload []LocationWorkload, batchSize int, hostShardCnt
 	for _, sh := range hostShardCnt {
 		shards += sh
 	}
+	if shards == 0 {
+		shards = 1
+	}
 	return &batchDispatcher{
 		mu:                    sync.Mutex{},
 		workload:              workload,
@@ -125,8 +128,7 @@ func (b *batchDispatcher) DispatchBatch(host string) (batch, bool) {
 	if dir == nil {
 		return batch{}, false
 	}
-	out := b.createBatch(l, t, dir, host)
-	return out, true
+	return b.createBatch(l, t, dir, host)
 }
 
 // Returns location for which batch should be created.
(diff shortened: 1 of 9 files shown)


Michal-Leszczynski:

pkg/service/restore/batch.go
```diff
@@ -153,15 +168,49 @@ func (b *batchDispatcher) chooseRemoteDir(table *TableWorkload) *RemoteDirWorklo
 }
 
 // Returns batch and updates RemoteDirWorkload and its parents.
-func (b *batchDispatcher) createBatch(l *LocationWorkload, t *TableWorkload, dir *RemoteDirWorkload) batch {
-	i := min(b.batchSize, len(dir.SSTables))
-	sstables := dir.SSTables[:i]
-	dir.SSTables = dir.SSTables[i:]
+func (b *batchDispatcher) createBatch(l *LocationWorkload, t *TableWorkload, dir *RemoteDirWorkload, host string) (batch, bool) {
```
karol-kokoszka, line 171: As per our sync.
Batches are created per manifest / keyspace / table. Please make sure that the last batch isn't smaller than the number of shards in the destination node.

karol-kokoszka (approved):

Michal-Leszczynski:
@karol-kokoszka I addressed the comments, could you take a look again?

karol-kokoszka (approved):

PR #4057, opened 2024-10-02
deps(managerclient): update manager deps to point to the commit from master

karol-kokoszka:
---

Please make sure that:


Michal-Leszczynski:
@karol-kokoszka what is this PR about? Did we have some problems with 3.3.3 release which included the package split? I see that you already merged 385fb1b05b1e0c3398855fa11c514b032e871775 to master, so why is this needed?

karol-kokoszka:
@Michal-Leszczynski managerclient module was poiting to the github.com/scylladb/scylla-manager/v3/pkg/utils dependency with commit taken from the branch different than master.
One of the scylla-cloud dev faced an issue because of that trying to build siren on GH actions env.

Michal-Leszczynski:
So does it mean that we have a regression in SM 3.3.3?

karol-kokoszka:
No, managerclient is independent from 3.3.3. We are not tagging (yet) the releases of managerclient.

Michal-Leszczynski (approved):

PR #4070, opened 2024-10-15
Restore improvement: rate limit

### LINKED ISSUE #3947: Use rate limit in the restore
Right now restore does not touch the rclone rate limit, so it basically uses the rate limit set by the last backup.
Restore should either:
- always set rate limit to 0 (disable rate limit)
- add a flag which allows for configuring it (with 0 as the default value)

cc: @gcarmin @tzach

Michal-Leszczynski:
This PR adds a way to control rate limit in the context of restore via new `--rate-limit` restore flag with default value 0 - meaning no rate limit at all.
It borrows rate limit syntax from backup pkg for consistency - meaning that it's possible to specify rate limit per dc.

Fixes #3947

### DIFF
--- a/pkg/scyllaclient/client_rclone.go
+++ b/pkg/scyllaclient/client_rclone.go
@@ -27,10 +27,23 @@ import (
 func (c *Client) RcloneSetBandwidthLimit(ctx context.Context, host string, limit int) error {
 	p := operations.CoreBwlimitParams{
 		Context:       forceHost(ctx, host),
-		BandwidthRate: &models.Bandwidth{Rate: fmt.Sprintf("%dM", limit)},
+		BandwidthRate: &models.Bandwidth{Rate: marshallRateLimit(limit)},
 	}
 	_, err := c.agentOps.CoreBwlimit(&p) // nolint: errcheck
 	return err
+}
+
+// RcloneGetBandwidthLimit gets bandwidth limit of all the current and future
+// transfers performed under current client session.
+func (c *Client) RcloneGetBandwidthLimit(ctx context.Context, host string) (string, error) {
+	p := operations.CoreBwlimitParams{
+		Context: forceHost(ctx, host),
+	}
+	resp, err := c.agentOps.CoreBwlimit(&p)
+	if err != nil {
+		return "", err
+	}
+	return resp.Payload.Rate, nil
 }
 
 // RcloneSetTransfers sets the default amount of transfers on rclone server.
@@ -737,4 +750,8 @@ func rcloneSplitRemotePath(remotePath string) (fs, path string, err error) {
 		path = dirParts[1]
 	}
 	return
+}
+
+func marshallRateLimit(limit int) string {
+	return fmt.Sprintf("%dM", limit)
 }
--- a/pkg/scyllaclient/client_rclone.go
+++ b/pkg/scyllaclient/client_rclone.go
@@ -240,14 +240,22 @@ func (c *Client) rcloneMoveOrCopyFile(ctx context.Context, host, dstRemotePath, 
 	return err
 }
 
+const (
+	// TransfersFromConfig describes transfers value which results in setting transfers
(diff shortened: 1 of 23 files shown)


pkg/service/backup/backup.go
```diff
@@ -3,59 +3,12 @@
...
-		dc, str := dcAtPos(i)
-		if dc == "" {
-			continue
-		}
-		if !allDCs.Has(dc) {
-			err = multierr.Append(err, errors.Errorf("%q no such datacenter %s", str, dc))
-		}
-	}
-	return
-}
```
karol-kokoszka, line 32: All this logic is moved to dclimit.go, right ?

Michal-Leszczynski: Yes - the only difference is that the error messages were slightly changed due to functions being more general.

pkg/scyllaclient/client_rclone.go
```diff
@@ -27,12 +27,25 @@ import (
...
 		Context:       forceHost(ctx, host),
-		BandwidthRate: &models.Bandwidth{Rate: fmt.Sprintf("%dM", limit)},
+		BandwidthRate: &models.Bandwidth{Rate: marshallRateLimit(limit)},
 	}
 	_, err := c.agentOps.CoreBwlimit(&p) // nolint: errcheck
 	return err
 }
 
+// RcloneGetBandwidthLimit gets bandwidth limit of all the current and future
+// transfers performed under current client session.
```
karol-kokoszka, line 37: Is it true that it's set only in the context of current client session ?
Isn't it changing the value permanently ?

pkg/scyllaclient/client_rclone.go
```diff
@@ -243,11 +264,11 @@ func (c *Client) RcloneMoveDir(ctx context.Context, host string, transfers int,
...
 // Remote path format is "name:bucket/path".
 // If specified, a suffix will be added to otherwise overwritten or deleted files.
-func (c *Client) RcloneCopyDir(ctx context.Context, host string, transfers int, dstRemotePath, srcRemotePath, suffix string) (int64, error) {
-	return c.rcloneMoveOrCopyDir(ctx, host, transfers, dstRemotePath, srcRemotePath, false, suffix)
+func (c *Client) RcloneCopyDir(ctx context.Context, host string, transfers, rateLimit int, dstRemotePath, srcRemotePath, suffix string) (int64, error) {
+	return c.rcloneMoveOrCopyDir(ctx, host, transfers, rateLimit, dstRemotePath, srcRemotePath, false, suffix)
 }
 
-func (c *Client) rcloneMoveOrCopyDir(ctx context.Context, host string, transfers int, dstRemotePath, srcRemotePath string, doMove bool, suffix string) (int64, error) {
+func (c *Client) rcloneMoveOrCopyDir(ctx context.Context, host string, transfers, rateLimit int, dstRemotePath, srcRemotePath string, doMove bool, suffix string) (int64, error) {
```
karol-kokoszka, line 271: Just a suggestion, as the transfers and ratelimit comes in pair.
Maybe it's worth to introduce simple struct name "transferParameters" or something else describing its purpose and keep the transfers and ratelimit inside ?

karol-kokoszka (approved):
:+1:

karol-kokoszka:
@Michal-Leszczynski Is it possible to give another commit name when you choose to "squash and merge" ? I see that there are plenty of "Restore improvement: transfers" etc commits on master.
`feat(restore): <explain>` looks better I guess.

pkg/scyllaclient/client_rclone.go
Michal-Leszczynski: For now they do come in pair, but it probably won't be the case for the new Scylla API which will replace the rclone API, so I would leave it as until we start using new Scylla API and see what is always needed.

karol-kokoszka: Makes sense

PR #4073, opened 2024-10-17
Restore improvement: unpin agent CPU

### LINKED ISSUE #3951: Allow for more pinned CPUs in the context of full speed restore
On startup agent tries to pin to CPUs. The code is presented below, but the logic looks like this:
- if desired (single) CPU is specified in `scylla-manager-agent-yaml`, pin to this CPU
- find CPUs not taken by Scylla (if any) (taken from `/etc/scylla.d/cpuset.conf`), pin to those CPUs
- otherwise, don't pin to any CPU, but set the max amount of utilized CPUs at any time to 1

```go
	// Try to get CPUs to pin to
	var cpus []int
	if s.config.CPU != agent.NoCPU {
		cpus = []int{s.config.CPU}
	} else if free, err := findFreeCPUs(); err != nil {
		if os.IsNotExist(errors.Cause(err)) || errors.Is(err, cpuset.ErrNoCPUSetConfig) {
			// Ignore if there is no cpuset file
			s.logger.Debug(ctx, "Failed to find CPUs to pin to", "error", err)
		} else {
			s.logger.Error(ctx, "Failed to find CPUs to pin to", "error", err)
		}
	} else {
		cpus = free
	}
	// Pin to CPUs if possible
	if len(cpus) == 0 {
		s.logger.Info(ctx, "Running on all CPUs")
		runtime.GOMAXPROCS(1)
	} else {
		if err := pinToCPUs(cpus); err != nil {
			s.logger.Error(ctx, "Failed to pin to CPUs", "cpus", cpus, "error", err)
		} else {
			s.logger.Info(ctx, "Running on CPUs", "cpus", cpus)
		}
		runtime.GOMAXPROCS(len(cpus))
	}
```
We should allow for controlling the CPU pinning in the context of faster restore speed.
The current implementation of restore does not allow for files download (agent) and load&stream (Scylla) at the same time, so even when they share the CPUs, they "wouldn't fight for them that much" (note that Scylla is still receiving data send to it by other load&streaming nodes in the Cluster while the download is happening).
Anyway, pinning CPUs to agent in the context of full speed restore should be bench-marked.

cc: @mykaul @tzach

Michal-Leszczynski:
This PR adds new flag `--unpin-agent-cpu` to restore task (default `false`).
It allows to unpinn agent from CPUs during restore (and re-pin it afterwards).

Moreover, for improved granularity, this PR also extends the `cpu` field of 
`scylla-manager-agent.yaml` config, so that user can specify a list of cpus
instead of just a single one. This allows for greater, but manual, control
over agent cpu pinning.

Fixes #3951


Michal-Leszczynski:
The tests already passed (except for the IPV6 ones, but that's because I forgot to modify their test config - now they should pass as well).

pkg/cmd/agent/agent.go
```diff
@@ -3,20 +3,49 @@
...
+	"github.com/scylladb/go-log"
 	"github.com/scylladb/scylla-manager/v3/pkg/config/agent"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/cpuset"
+	"github.com/scylladb/scylla-manager/v3/swagger/gen/agent/models"
 )
 
-func newAgentHandler(c agent.Config, rclone http.Handler) *chi.Mux {
+func newAgentHandler(c agent.Config, rclone http.Handler, logger log.Logger) *chi.Mux {
 	m := chi.NewMux()
+	ctx := context.Background()
```
karol-kokoszka, line 22: There is no need for this context.

pkg/cmd/agent/agent.go
```diff
@@ -3,20 +3,49 @@
...
+			render.Respond(writer, request, err)
+		}
+		casted := make([]int64, len(cpus))
+		for i := range cpus {
+			casted[i] = int64(cpus[i])
+		}
+		render.Respond(writer, request, &models.Cpus{CPU: casted})
+	})
+	m.Post("/pin_cpu", func(writer http.ResponseWriter, request *http.Request) {
+		if err := findAndPinCPUs(ctx, c, logger); err != nil {
```
karol-kokoszka, line 38: By doing it, you pass the context created during handler creation.
It makes no sense, as the context here is the request.
There is a context available in `request *http.Reqest`, please use this one.
`request.Context()`

pkg/config/agent/agent.go
```diff
@@ -121,3 +121,36 @@ func Obfuscate(c Config) Config {
...
+		return nil
+	}
+	var arr []int
+	errArr := unmarshal(&arr)
+	if errArr == nil {
+		*c = arr
+		return nil
+	}
+	return errors.Wrap(multierr.Append(errV, errArr), "parse CPU as an int or int array")
+}
```
karol-kokoszka, line 147: What if there is no entry in config yaml ? 
It will produce a nil slice.

And that means agent will be not be pinned to any CPU, but only limited to one process.
It's expected, OK.

karol-kokoszka (approved):
:+1: Thanks.
Please just use request.Context() in API call.

PR #4077, opened 2024-10-22
Parallelized cluster hosts discovering + async config cache update

### LINKED ISSUE #4074: Cluster update API call may timeout when one of the known hosts is down
The current logic for updating the cluster in Scylla Manager includes validating host connectivity. This process involves the Manager querying one of the known hosts (stored in the ManagerDB) to discover other cluster hosts via the GET /storage_service/host_id API and mapping them to the correct data center using GET /snitch/datacenter.

After discovering live hosts, the Scylla Manager Agent API on each host is pinged to validate whether the Scylla Manager Agent is installed and responsive.

This is necessary because the Scylla Manager server requires the Scylla Manager Agent to be present on every node in order to execute all supported tasks.

Unfortunately, the logic for discovering all nodes in the cluster involves iterating over all the nodes saved in the database. The Manager keeps trying until it receives a response from the Scylla Manager Agent on one of the nodes. The default timeout for this operation is 30 seconds, with a few retries. The Manager moves on to the next node in the queue if the current node fails to respond.

If the first node in the list is down or unavailable, it blocks the cluster update, forcing it to wait until the server proceeds to a valid node.

The default manager client has the following timeout configuration: [Scylla Manager Client Timeout](https://github.com/scylladb/scylla-manager/blob/d044b1738488013960dc90289a947440e3e922cf/v3/pkg/managerclient/client.go#L38-L51).

This means that even if the cluster update eventually succeeds, the API call to the Manager server might timeout. As a result, the Scylla Manager CLI, which is a wrapper over the Manager server API, will also report an error when updating the cluster.

The Manager server should call all the currently known hosts in parallel and proceed with the first one that responds, canceling calls to the other nodes once a valid response is received.

karol-kokoszka:
Fixes #4074

This PR makes the cluster hosts discovery more robust, as when the cluster.host is DOWN,
it probes all other hosts in parallel and returns response from the fastest one.
Additionally, this PR makes the call to cluster config cache async on updating cluster.

---

Please make sure that:

### DIFF
--- a/pkg/service/cluster/service.go
+++ b/pkg/service/cluster/service.go
@@ -230,11 +230,12 @@ func (s *Service) discoverClusterHosts(ctx context.Context, c *Cluster) (knownHo
 	}()
 
 	// Read results until the channel is closed
-	for hosts := range result {
+	hosts, ok := <-result
+	if ok {
 		return hosts.known, hosts.live, nil
 	}
 
-	// If no valid results, return error<
+	// If no valid results, return error
 	return nil, nil, ErrNoValidKnownHost
 }
 
--- a/pkg/service/cluster/service.go
+++ b/pkg/service/cluster/service.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/gocql/gocql"
@@ -34,7 +35,10 @@ type ProviderFunc func(ctx context.Context, id uuid.UUID) (*Cluster, error)
 type ChangeType int8
 
 // ErrNoValidKnownHost is thrown when it was not possible to connect to any of the currently known hosts of the cluster.
-var ErrNoValidKnownHost = errors.New("unable to connect to any of cluster's known hosts")
+var (
+	ErrNoValidKnownHost    = errors.New("unable to connect to any of cluster's known hosts")
+	ErrNoLiveHostAvailable = errors.New("no single live host available")
+)
 
 // ChangeType enumeration.
 const (
@@ -156,7 +160,7 @@ func (s *Service) clientConfig(c *Cluster) scyllaclient.Config {
 }
 
 func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) error {
-	knownHosts, err := s.discoverClusterHosts(ctx, c)
+	knownHosts, _, err := s.discoverClusterHosts(ctx, c)
 	if err != nil {
 		if errors.Is(err, ErrNoValidKnownHost) {
 			s.logger.Error(ctx, "There is no single valid known host for the cluster. "+
@@ -171,55 +175,124 @@ func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) er
 	return errors.Wrap(s.setKnownHosts(c, knownHosts), "update known_hosts in SM DB")
 }
 
-func (s *Service) discoverClusterHosts(ctx context.Context, c *Cluster) ([]string, error) {
-	var contactPoints []string
+const (
+	discoverClusterHostsTimeout = 5 * time.Second
+)
+
+func (s *Service) discoverClusterHosts(ctx context.Context, c *Cluster) (knownHosts, liveHosts []string, err error) {
 	if c.Host != "" {
-		contactPoints = append(contactPoints, c.Host) // Go with the designated contact point first
+		knownHosts, liveHosts, err := s.discoverClusterHostUsingCoordinator(ctx, c, discoverClusterHostsTimeout, c.Host)
+		if err != nil {
+			s.logger.Error(ctx, "Couldn't discover hosts using stored coordinator host, proceeding with other known ones",
+				"coordinator-host", c.Host, "error", err)
+		} else {
+			return knownHosts, liveHosts, nil
+		}
 	} else {
 		s.logger.Error(ctx, "Missing --host flag. Using only previously discovered hosts instead", "cluster ID", c.ID)
 	}
-	contactPoints = append(contactPoints, c.KnownHosts...) // In case it failed, try to contact previously discovered hosts
-
-	for _, cp := range contactPoints {
-		if cp == "" {
-			s.logger.Error(ctx, "Empty contact point", "cluster ID", c.ID, "contact points", contactPoints)
-			continue
-		}
-
-		config := scyllaclient.DefaultConfigWithTimeout(s.timeoutConfig)
-		if c.Port != 0 {
-			config.Port = strconv.Itoa(c.Port)
-		}
-		config.AuthToken = c.AuthToken
-		config.Hosts = []string{cp}
-
-		client, err := scyllaclient.NewClient(config, s.logger.Named("client"))
-		if err != nil {
-			s.logger.Error(ctx, "Couldn't connect to contact point", "contact point", cp, "error", err)
-			continue
-		}
-
-		knownHosts, err := s.discoverHosts(ctx, client)
-		logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")
-		if err != nil {
-			s.logger.Error(ctx, "Couldn't discover hosts", "host", cp, "error", err)
-			continue
-		}
-		return knownHosts, nil
+	if len(c.KnownHosts) < 1 {
+		return nil, nil, ErrNoValidKnownHost
 	}
 
-	return nil, ErrNoValidKnownHost
+	wg := sync.WaitGroup{}
+	type hostsTuple struct {
+		live, known []string
+	}
+	result := make(chan hostsTuple, len(c.KnownHosts))
+	discoverContext, discoverCancel := context.WithCancel(ctx)
+	defer discoverCancel()
+
+	for _, cp := range c.KnownHosts {
+		wg.Add(1)
+
+		go func(host string) {
+			defer wg.Done()
+
+			knownHosts, liveHosts, err := s.discoverClusterHostUsingCoordinator(discoverContext, c, discoverClusterHostsTimeout, host)
+			if err != nil {
+				// Only log if the context hasn't been canceled
+				if !errors.Is(discoverContext.Err(), context.Canceled) {
+					s.logger.Error(ctx, "Couldn't discover hosts", "host", host, "error", err)
+				}
+				return
+			}
+			result <- hostsTuple{
+				live:  liveHosts,
+				known: knownHosts,
+			}
+		}(cp)
+	}
+
+	go func() {
+		wg.Wait()
+		close(result)
+	}()
+
+	// Read results until the channel is closed
+	for hosts := range result {
+		return hosts.known, hosts.live, nil
+	}
+
+	// If no valid results, return error<
+	return nil, nil, ErrNoValidKnownHost
+}
+
+func (s *Service) discoverClusterHostUsingCoordinator(ctx context.Context, c *Cluster, apiCallTimeout time.Duration,
+	host string,
+) (knownHosts, liveHosts []string, err error) {
+	config := scyllaclient.DefaultConfigWithTimeout(s.timeoutConfig)
+	if c.Port != 0 {
+		config.Port = strconv.Itoa(c.Port)
+	}
+	config.Timeout = apiCallTimeout
+	config.AuthToken = c.AuthToken
+	config.Hosts = []string{host}
+
+	client, err := scyllaclient.NewClient(config, s.logger.Named("client"))
+	if err != nil {
+		return nil, nil, err
+	}
+	defer logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")
+
+	liveHosts, err = client.GossiperEndpointLiveGet(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	knownHosts, err = s.discoverHosts(ctx, client, liveHosts)
+	if err != nil {
+		return nil, nil, err
+	}
+	return knownHosts, liveHosts, nil
 }
 
 // discoverHosts returns a list of all hosts sorted by DC speed. This is
 // an optimisation for Epsilon-Greedy host pool used internally by
 // scyllaclient.Client that makes it use supposedly faster hosts first.
-func (s *Service) discoverHosts(ctx context.Context, client *scyllaclient.Client) (hosts []string, err error) {
+func (s *Service) discoverHosts(ctx context.Context, client *scyllaclient.Client, liveHosts []string) (hosts []string, err error) {
+	if len(liveHosts) == 0 {
+		return nil, ErrNoLiveHostAvailable
+	}
+
 	dcs, err := client.Datacenters(ctx)
 	if err != nil {
 		return nil, err
 	}
-	closest, err := client.ClosestDC(ctx, dcs)
+	// remove dead nodes from the map
+	liveSet := make(map[string]struct{})
+	for _, host := range liveHosts {
+		liveSet[host] = struct{}{}
+	}
+	filteredDCs := make(map[string][]string)
+	for dc, hosts := range dcs {
+		for _, host := range hosts {
+			if _, isLive := liveSet[host]; isLive {
+				filteredDCs[dc] = append(filteredDCs[dc], host)
+			}
+		}
+	}
+
+	closest, err := client.ClosestDC(ctx, filteredDCs)
 	if err != nil {
 		return nil, err
 	}
@@ -397,7 +470,7 @@ func (s *Service) PutCluster(ctx context.Context, c *Cluster) (err error) {
 	}
 
 	// Check hosts connectivity.
-	if err := s.validateHostsConnectivity(ctx, c); err != nil {
+	if err := s.ValidateHostsConnectivity(ctx, c); err != nil {
 		var tip string
 		switch scyllaclient.StatusCodeOf(err) {
 		case 0:
@@ -487,36 +560,34 @@ func (s *Service) PutCluster(ctx context.Context, c *Cluster) (err error) {
 	return s.notifyChangeListener(ctx, changeEvent)
 }
 
-func (s *Service) validateHostsConnectivity(ctx context.Context, c *Cluster) error {
+// ValidateHostsConnectivity validates that scylla manager agent API is available and responding on all live hosts.
+// Hosts are discovered using cluster.host + cluster.knownHosts saved to the manager's database.
+func (s *Service) ValidateHostsConnectivity(ctx context.Context, c *Cluster) error {
 	if err := s.loadKnownHosts(c); err != nil && !errors.Is(err, gocql.ErrNotFound) {
 		return errors.Wrap(err, "load known hosts")
 	}
 
-	knownHosts, err := s.discoverClusterHosts(ctx, c)
+	knownHosts, liveHosts, err := s.discoverClusterHosts(ctx, c)
 	if err != nil {
 		return errors.Wrap(err, "discover cluster hosts")
 	}
 	c.KnownHosts = knownHosts
 
+	if len(liveHosts) == 0 {
+		return util.ErrValidate(errors.New("no live nodes"))
+	}
+
 	config := s.clientConfig(c)
+	config.Hosts = liveHosts
 	client, err := scyllaclient.NewClient(config, s.logger.Named("client"))
 	if err != nil {
 		return err
 	}
 	defer logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")
 
-	status, err := client.Status(ctx)
-	if err != nil {
-		return errors.Wrap(err, "cluster status")
-	}
-	live := status.Live().Hosts()
-	if len(live) == 0 {
-		return util.ErrValidate(errors.New("no live nodes"))
-	}
-
 	var errs error
-	for i, err := range client.CheckHostsConnectivity(ctx, live) {
-		errs = multierr.Append(errs, errors.Wrap(err, live[i]))
+	for i, err := range client.CheckHostsConnectivity(ctx, liveHosts) {
+		errs = multierr.Append(errs, errors.Wrap(err, liveHosts[i]))
 	}
 	if errs != nil {
 		return util.ErrValidate(errors.Wrap(errs, "connectivity check"))
--- a/pkg/service/cluster/service_integration_test.go
+++ b/pkg/service/cluster/service_integration_test.go
@@ -33,6 +33,121 @@ import (
 	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
 )
 
+func TestValidateHostConnectivityIntegration(t *testing.T) {
+	Print("given: the fresh cluster")
+	var (
+		ctx          = context.Background()
+		session      = CreateScyllaManagerDBSession(t)
+		secretsStore = store.NewTableStore(session, table.Secrets)
+		c            = &cluster.Cluster{
+			AuthToken: "token",
+			Host:      ManagedClusterHost(),
+		}
+	)
+	s, err := cluster.NewService(session, metrics.NewClusterMetrics(), secretsStore, scyllaclient.DefaultTimeoutConfig(),
+		server.DefaultConfig().ClientCacheTimeout, log.NewDevelopment())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = s.PutCluster(context.Background(), c)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	allHosts := ManagedClusterHosts()
+	for _, tc := range []struct {
+		name      string
+		hostsDown []string
+		result    error
+		timeout   time.Duration
+	}{
+		{
+			name:      "coordinator host is DOWN",
+			hostsDown: []string{ManagedClusterHost()},
+			result:    nil,
+			timeout:   6 * time.Second,
+		},
+		{
+			name:      "only one is UP",
+			hostsDown: allHosts[:len(allHosts)-1],
+			result:    nil,
+			timeout:   6 * time.Second,
+		},
+		{
+			name:      "all hosts are DOWN",
+			hostsDown: allHosts,
+			result:    cluster.ErrNoValidKnownHost,
+			timeout:   11 * time.Second, // the 5 seconds calls will timeout twice
+		},
+		{
+			name:      "all hosts are UP",
+			hostsDown: nil,
+			result:    nil,
+			timeout:   6 * time.Second,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			defer func() {
+				for _, host := range tc.hostsDown {
+					if err := StartService(host, "scylla"); err != nil {
+						t.Logf("error on starting stopped scylla service on host={%s}, err={%s}", host, err)
+					}
+					if err := RunIptablesCommand(host, CmdUnblockScyllaREST); err != nil {
+						t.Logf("error trying to unblock REST API on host = {%s}, err={%s}", host, err)
+					}
+				}
+			}()
+
+			Printf("then: validate that call to validate host connectivity takes less than %v seconds", tc.timeout.Seconds())
+			testCluster, err := s.GetClusterByID(context.Background(), c.ID)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if err := callValidateHostConnectivityWithTimeout(ctx, s, tc.timeout, testCluster); err != nil {
+				t.Fatal(err)
+			}
+			Printf("when: the scylla service is stopped and the scylla API is timing out on some hosts")
+			// It's needed to block Scylla REST API, so that the clients are just hanging when they call the API.
+			// Scylla service must be stopped to make the node to report DOWN status. Blocking REST API is not
+			// enough.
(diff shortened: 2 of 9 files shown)


karol-kokoszka:
@Michal-Leszczynski Please take a look on the PR.

Michal-Leszczynski:
@karol-kokoszka It looks like tests are failing.

karol-kokoszka:
Yes, some of them are failing and I have no clue yet why.

karol-kokoszka:
And I'm not able to reproduce it locally, using the same setup as was used on failing test.
It basically PASSes locally.
I'm starting to think that it's something related to the env on gh actions.

karol-kokoszka:
Looks that ensuring that all nodes are UP on the test that was failing previously helped.
We are starting/stopping services using `supervisorctl` as we refer to the docker environment. Service start doesn't mean that the node is UP.
I added simple utility function that ensures that nodes are UP.

@Michal-Leszczynski please review.

pkg/service/healthcheck/service_integration_test.go
```diff
@@ -547,6 +553,36 @@ func tryStartAgent(t *testing.T, hosts []string) {
 	}
 }
 
+func ensureNodesAreUP(t *testing.T, hosts []string, timeout time.Duration) error {
```
Michal-Leszczynski, line 556: Can't we just create a client with all nodes and get their status from there?
If it's important to get how each nodes self reports its status, instead of the statuses perceived by a single node, we could either create per node client or `forceHost` in status requests.
It would allow us to skip manual output parsing and make things easier in general.

pkg/service/healthcheck/service_integration_test.go
```diff
@@ -547,6 +553,36 @@ func tryStartAgent(t *testing.T, hosts []string) {
...
+	}
+
+	go func() {
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		return nil
```
Michal-Leszczynski, line 580: So the error (that could come from the fact that a node is down) is ignored (except for being logged), so the whole function can return nil even when nodes are down?

pkg/testutils/exec.go
```diff
@@ -101,3 +103,35 @@ func StartService(h, service string) error {
...
+func WaitForNodeUP(h string, timeout time.Duration) error {
+	nodeIsReady := make(chan struct{})
+	done := make(chan struct{})
+	go func() {
+		defer close(nodeIsReady)
+		for {
+			select {
+			case <-done:
+				return
+			case <-time.After(time.Second):
```
Michal-Leszczynski, line 118: So we wait 1 second even before the first check? This could be problematic with small timeouts like 1 second.

pkg/service/healthcheck/service_integration_test.go
```diff
@@ -100,6 +105,7 @@ func TestStatus_Ping_Independent_From_REST_Integration(t *testing.T) {
 	}
 
 	// When #1 -> default scenario where everything works fine
+	// Retry 5 times, as the service could be stopped by other
```
Michal-Leszczynski, line 108: Where does this 5 come from?

pkg/service/healthcheck/service_integration_test.go
```diff
@@ -52,6 +54,9 @@ func TestStatus_Ping_Independent_From_REST_Integration(t *testing.T) {
 	tryUnblockREST(t, ManagedClusterHosts())
 	tryUnblockAlternator(t, ManagedClusterHosts())
 	tryStartAgent(t, ManagedClusterHosts())
+	if err := ensureNodesAreUP(t, ManagedClusterHosts(), time.Minute); err != nil {
+		t.Fatalf("not all nodes are UP, err = {%v}", err)
+	}
```
Michal-Leszczynski, line 59: Why was this added to this test?

pkg/service/cluster/service_integration_test.go
```diff
@@ -33,6 +33,114 @@ import (
...
+	} {
+		t.Run(tc.name, func(innerT *testing.T) {
+			defer func() {
+				for _, host := range tc.hostsDown {
+					if err := StartService(host, "scylla"); err != nil {
+						innerT.Logf("error on starting stopped scylla service on host={%s}, err={%s}", host, err)
+					}
+					if err := RunIptablesCommand(host, CmdUnblockScyllaREST); err != nil {
+						innerT.Logf("error trying to unblock REST API on host = {%s}, err={%s}", host, err)
+					}
```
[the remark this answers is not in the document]
karol-kokoszka: Being honest I don't know.
I want to reach the state where the API calls are hanging, so that the call may timeout. This part is done by calling the CmdBlockScyllaREST.
Besides that, I want the node to report DOWN status.

Let me do the check with just blocking the API.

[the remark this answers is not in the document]
karol-kokoszka: OK, so making the API unresponsive doesn't make the node to be DOWN.

When the scylla service is just stopped, then the VM (or container) refuses connection and there is no hanging on calling the API.

What makes sense, as Scylla is using gossip protocol to discover other nodes. Blocking gossip protocol would do the job I guess. But I don't see it as sommething necessary, as the service can be just stopped.

pkg/service/cluster/service_integration_test.go
```diff
@@ -33,6 +33,114 @@ import (
...
+				}
+			}()
+
+			// then: validate that call takes less than 5 seconds
```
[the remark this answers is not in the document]
karol-kokoszka: sure, makes sense

pkg/service/cluster/service_integration_test.go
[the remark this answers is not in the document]
Michal-Leszczynski: Okay, so I get why blocking REST API is not enough, but wouldn't just stopping the node work?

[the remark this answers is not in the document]
karol-kokoszka: > When the scylla service is just stopped, then the VM (or container) refuses connection and there is no hanging on calling the API.

> What makes sense, as Scylla is using gossip protocol to discover other nodes. Blocking gossip protocol would do the job I guess. But I don't see it as sommething necessary, as the service can be just stopped.

The service must be stopped to block gossiper and cause the node to be reported as DOWN.

pkg/service/cluster/service_integration_test.go
```diff
@@ -33,6 +33,114 @@ import (
...
+				if err := StopService(host, "scylla"); err != nil {
+					innerT.Fatal(err)
+				}
+				if err := RunIptablesCommand(host, CmdBlockScyllaREST); err != nil {
+					innerT.Error(err)
+				}
+			}
+
+			// then: validate that call still takes less than 5 seconds
+			if err := callValidateHostConnectivityWithTimeout(ctx, s, 30*time.Second, testCluster); !errors.Is(err, tc.result) {
```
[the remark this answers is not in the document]
karol-kokoszka: Yeah, that's a good question.
I was bit too much optimistic with 5 secs for the cluster with unresponsive coordinator-host.

First call to coordniator already exhausts the 5 sec budged.
My tests shown that 11 seconds is a minimum to pass.

Let me change the expectation for 15 seconds. It's still OK.

pkg/service/cluster/service_integration_test.go
```diff
@@ -33,6 +33,114 @@ import (
...
+			hostsDown: allHosts,
+			result:    cluster.ErrNoValidKnownHost,
+		},
+		{
+			name:      "all hosts are UP",
+			hostsDown: nil,
+			result:    nil,
+		},
+	} {
+		t.Run(tc.name, func(innerT *testing.T) {
```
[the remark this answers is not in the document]
karol-kokoszka: The only reasoning for that is to make it clear that test context here is not the global one.
It doesn't prohibit calling t instead of innerT when t is shadowed.
It's just cleaner IMHO, but I have no strong arguments to convince you that it's better than shadowing.

pkg/service/cluster/service.go
```diff
@@ -171,44 +174,96 @@ func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) er
...
-		knownHosts, err := s.discoverHosts(ctx, client)
-		logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")
-		if err != nil {
-			s.logger.Error(ctx, "Couldn't discover hosts", "host", cp, "error", err)
-			continue
+			knownHosts, liveHosts, err := s.discoverClusterHostUsingCoordinator(discoverContext, c, 5*time.Second, host)
+			if err != nil && !errors.Is(err, context.Canceled) {
+				s.logger.Error(ctx, "Couldn't discover hosts", "host", host, "error", err)
+				return
+			}
```
Michal-Leszczynski, line 212: What if the error is `context.Canceled`? Shouldn't we return as well?

pkg/service/cluster/service.go
```diff
@@ -171,44 +174,96 @@ func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) er
...
+	case <-done: // all goroutines are done
+		select {
+		case hosts := <-result:
+			return hosts.known, hosts.live, nil
+		default:
+			return nil, nil, ErrNoValidKnownHost
 		}
-		return knownHosts, nil
 	}
+}
```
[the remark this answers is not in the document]
Michal-Leszczynski: Okay, I see that this was made in such way in order to handle no valid host case.
But wouldn't simply closing the result chan be cleaner?

pkg/service/cluster/service_integration_test.go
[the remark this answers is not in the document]
Michal-Leszczynski: I missed that part. Could you put it into comment, so that someone won't try to remove one of those parts in the future?

pkg/service/cluster/service_integration_test.go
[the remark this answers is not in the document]
karol-kokoszka: ok, let me change it then

pkg/service/cluster/service_integration_test.go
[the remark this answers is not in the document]
karol-kokoszka: Sure, let me put that into the comment

pkg/service/cluster/service.go
```diff
@@ -171,44 +174,96 @@ func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) er
...
+	client, err := scyllaclient.NewClient(config, s.logger.Named("client"))
+	if err != nil {
+		return nil, nil, err
+	}
+	knownHosts, err = s.discoverHosts(ctx, client)
+	logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")
+	if err != nil {
+		return nil, nil, err
+	}
+	liveHosts, err = client.GossiperEndpointLiveGet(ctx)
```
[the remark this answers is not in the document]
karol-kokoszka: seems it is...
it should be deferred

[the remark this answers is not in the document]
karol-kokoszka: I wondered why it didn't fail, but it looks that Close() here is just closing idle connections
```
// Close closes all the idle connections.
func (c *Client) Close() error {
	if t, ok := c.config.Transport.(*http.Transport); ok {
		t.CloseIdleConnections()
	}
	if c.hostPool != nil {
		c.hostPool.Close()
	}
	return nil
}
```

pkg/service/cluster/service.go
```diff
@@ -171,44 +174,96 @@ func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) er
 	return errors.Wrap(s.setKnownHosts(c, knownHosts), "update known_hosts in SM DB")
 }
 
-func (s *Service) discoverClusterHosts(ctx context.Context, c *Cluster) ([]string, error) {
-	var contactPoints []string
+func (s *Service) discoverClusterHosts(ctx context.Context, c *Cluster) (knownHosts, liveHosts []string, err error) {
 	if c.Host != "" {
-		contactPoints = append(contactPoints, c.Host) // Go with the designated contact point first
+		knownHosts, liveHosts, err := s.discoverClusterHostUsingCoordinator(ctx, c, 5*time.Second, c.Host)
```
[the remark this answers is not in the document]
karol-kokoszka: The call here has different purpose that the calls that are relying on this timeout value from config.
The timeout is used on scylla and agent clients.

Here, we expect to have quick answer.

I exported this value to const.

pkg/service/cluster/service.go
```diff
@@ -171,44 +174,96 @@ func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) er
...
-		}
+		go func(host string) {
+			defer wg.Done()
 
-		knownHosts, err := s.discoverHosts(ctx, client)
-		logutil.LogOnError(ctx, s.logger, client.Close, "Couldn't close scylla client")
-		if err != nil {
-			s.logger.Error(ctx, "Couldn't discover hosts", "host", cp, "error", err)
-			continue
+			knownHosts, liveHosts, err := s.discoverClusterHostUsingCoordinator(discoverContext, c, 5*time.Second, host)
```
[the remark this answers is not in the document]
karol-kokoszka: Moved to const.

pkg/service/cluster/service.go
```diff
@@ -171,44 +174,96 @@ func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) er
...
+				return
+			}
+			select {
+			case result <- hostsTuple{
+				live:  liveHosts,
+				known: knownHosts,
+			}:
+			case <-discoverContext.Done():
+				return
+			}
```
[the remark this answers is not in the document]
karol-kokoszka: When the discoverContext is canceled, then the knownHosts and liveHosts will be empty, but you are right. It's not a problem to send it to result channel...
Thanks ! I'll remove this case from select.

pkg/service/cluster/service.go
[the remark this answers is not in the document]
karol-kokoszka: I was thinking about very corner case, where all goroutines are done at almost the same moment, one of them wrote to the channel, but the idle goroutine that was waiting on wg.Wait is woken up and closed the done channel.
Then it is random of which case is going to be selected.
If it's <-done, then I still want to validate if there is something to read from result.

Does it make sense ?

I don't know how much realistic scenario is it.

pkg/service/cluster/service.go
```diff
@@ -487,36 +542,34 @@ func (s *Service) PutCluster(ctx context.Context, c *Cluster) (err error) {
...
-	live := status.Live().Hosts()
-	if len(live) == 0 {
-		return util.ErrValidate(errors.New("no live nodes"))
-	}
-
 	var errs error
-	for i, err := range client.CheckHostsConnectivity(ctx, live) {
-		errs = multierr.Append(errs, errors.Wrap(err, live[i]))
+	for i, err := range client.CheckHostsConnectivity(ctx, liveHosts) {
+		errs = multierr.Append(errs, errors.Wrap(err, liveHosts[i]))
```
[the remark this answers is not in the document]
karol-kokoszka: It's a copied code, not changed in this PR.
Let's leave it.

pkg/service/healthcheck/service_integration_test.go
karol-kokoszka: We could use the client.
But if someone called to blockScyllaAPI before, then it will never return the status found with the gossiper as the call will be blocked.
Let's leave the "nodetool status" to do this check.

pkg/service/healthcheck/service_integration_test.go
karol-kokoszka: Yes, error is just logged. It comes from the fact that WaitForNodeUP may only return error when it times out.
But the logic is blurry, I agree. Even though that the function will not return that node is UP when it's DOWN, but it will return timeout error.

Anyway, let me try to make the logic bit more easy to read and understand.

pkg/testutils/exec.go
karol-kokoszka: You are right. Fixed

pkg/service/healthcheck/service_integration_test.go
karol-kokoszka: I don't know. Possibly some copy&paste leftover. Thanks.

pkg/service/cluster/service.go
karol-kokoszka: not necessary, but yes let's return

pkg/service/cluster/service_integration_test.go
```diff
@@ -33,6 +33,114 @@ import (
...
+
+	done := make(chan error)
+	go func() {
+		done <- s.ValidateHostsConnectivity(callCtx, c)
+	}()
+
+	select {
+	case <-time.After(timeout):
+		cancel()
+		return fmt.Errorf("expected s.ValidateHostsConnectivity to complete in less than %v seconds, time exceeded", timeout.Seconds())
```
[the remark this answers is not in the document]
Michal-Leszczynski: I agree, but why would you use `context.WithTimeout` instead of `context.WithCancel`?
The second option makes it more clear that the timeout mechanism is controlled by the `select` clause,
and that the cancel is done just for safety.

pkg/service/cluster/service.go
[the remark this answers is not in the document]
Michal-Leszczynski: But wouldn't using `hosts, ok := <-result` instead of `hosts := result` do the trick?
It allows to distinguish between the success and `ErrNoValidKnownHost` without nested select statements.

pkg/service/healthcheck/service_integration_test.go
Michal-Leszczynski: > But if someone called to blockScyllaAPI before, then it will never return the status found with the gossiper as the call will be blocked.

It should return after timeout, right? If we want for it to be quick, we can set the timeout to a low value.

pkg/service/healthcheck/service_integration_test.go
Michal-Leszczynski: > Even though that the function will not return that node is UP when it's DOWN, but it will return timeout error.

But the timeout error in this case is synonymous with node not being in UN state, right?
And then it's ignored - which means that this function has a race between the timeout set within its body and the timeout set in the per-node check. So it could return no error when all nodes are down and timeout on UN check.

pkg/service/cluster/service.go
Michal-Leszczynski: Why is it not necessary? If we don't return, we still put some undefined values into the `result` chan, and the main function body reading from this chan will believe that everything is ok, when it's not.

karol-kokoszka: It's exactly "undefined" values, but just nil.
The main function body is expected to read to correct result already, so it would be put to the channel but never consumed.

karol-kokoszka: But if the parent context is cancelled, then it triggers child ones to be cancelled as well.
It means that the function would return "nil, nil, nil", what is wrong.

But now, after fixing this what you suggested it will work fine and the "ErrNoValidKnownHost" is going to be returned. Nice !

pkg/service/cluster/service_integration_test.go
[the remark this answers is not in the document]
karol-kokoszka: Actually, replacing context.WithTimeout to context.WithCancel makes sense. Let me change it.

pkg/service/cluster/service.go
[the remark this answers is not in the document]
karol-kokoszka: But it's random which case statement is going to chosen if condition of both is true.
```
	select {
	case hosts, ok := <-result: // valid condition
	case <-done: // valid condition
	}
```

https://go.dev/ref/spec#Select_statements

> If one or more of the communications can proceed, a single one that can proceed is chosen via a uniform pseudo-random selection. Otherwise, if there is a default case, that case is chosen. If there is no default case, the "select" statement blocks until at least one of the communications can proceed.

Or you suggest something else ?

pkg/service/healthcheck/service_integration_test.go
karol-kokoszka: It's not about the timeout only, but about actual response.
If the API is blocked I won't know which nodes are UP/DN, but I do know it when I use nodetool status.

pkg/service/cluster/service.go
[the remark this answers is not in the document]
Michal-Leszczynski: I'm suggesting to get rid of `done` and just wait on the result chan (and perhaps on some `time.After` if needed).

pkg/service/cluster/service.go
karol-kokoszka: That's a valid statement.
Replaced with `discoverContext.Err() != context.Canceled`
I want to avoid the situation where one of the nodes reported the live and known hosts. 
DiscoverContext is cancelled and the goroutinet logs this information.These logs are unnecessary. But would be good to know if the error reason is different.

pkg/service/cluster/service_integration_test.go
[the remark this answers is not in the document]
Michal-Leszczynski: Did you investigate why it takes 11 seconds and not 5 seconds + epsilon?

pkg/service/healthcheck/service_integration_test.go
Michal-Leszczynski: Okay, but getting back to the beginning - we have a test that stops a node and blocks its API. At the end of this test we want to start the node and unblock its API. We also want to wait for those changes to really set in, so that they don't affect other tests by accident.

1. Why can't we simply create a client to all of the nodes and send some per host requests (e.g. `client.Snapshots`) to all of them sequentially? Node needs to be both up and responsive on API for it to succeed. We can do it sequentially, because it's not like querying the endpoint does any work - it just check responsiveness, so making it in parallel does not provide visible time improvement, but it adds some complexity to the code. If you prefer to do it in parallel, we should use already implemented and tested functions to do so (e.g. `parallel.Run`)
2. Shouldn't we add this waiting step to the test that stops nodes? I would say that it's easier to to remember to run this function at the end of the test that modifies node state, instead of running it at the beginning of every test that requires nodes to be in the correct state (so basically all of the tests). In this PR you encountered this problem only for a single test, but that's because of given sequential order of test execution that we shouldn't rely on.

Michal-Leszczynski:
@karol-kokoszka I resolved some already fixed comments.
Comments regarding checking node state in tests ([com1](https://github.com/scylladb/scylla-manager/pull/4077#discussion_r1814492480), [com2](https://github.com/scylladb/scylla-manager/pull/4077#discussion_r1814500665), [com3](https://github.com/scylladb/scylla-manager/pull/4077#discussion_r1814511841)) are tied to this [conversation](https://github.com/scylladb/scylla-manager/pull/4077#discussion_r1829166168).
When all comments are resolved, I will re-review this PR and we will hopefully be able to merge it!

pkg/service/cluster/service.go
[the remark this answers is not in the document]
karol-kokoszka: Actually you are right.
My concerns are not valid when the result chan is closed.
And it's more idiomatic to close channel than building such an extensive select statement.
Fixed

pkg/service/healthcheck/service_integration_test.go
karol-kokoszka: > Why can't we simply create a client to all of the nodes and send some per host requests (e.g. client.Snapshots) to all of them sequentially? Node needs to be both up and responsive on API for it to succeed. We can do it sequentially, because it's not like querying the endpoint does any work - it just check responsiveness, so making it in parallel does not provide visible time improvement, but it adds some complexity to the code. If you prefer to do it in parallel, we should use already implemented and tested functions to do so (e.g. parallel.Run)

That's an interesting point, but I think there's a distinction here. A node doesn't necessarily need to be fully "UP" to respond to some API calls—it could still be in a "Joining" state, for example.
Using nodetool status gives us a more reliable check to ensure the node is fully operational (UP) as part of the cluster. For this test, I also want to ensure the agent service is running and that ports are unblocked, which nodetool helps verify comprehensively.

> Shouldn't we add this waiting step to the test that stops nodes? I would say that it's easier to to remember to run this function at the end of the test that modifies node state, instead of running it at the beginning of every test that requires nodes to be in the correct state (so basically all of the tests). In this PR you encountered this problem only for a single test, but that's because of given sequential order of test execution that we shouldn't rely on.

That's a valid point, and I agree it could simplify some workflows if we add the waiting step to the test that stops nodes. However, I still prefer to harden and explicitly check the assumptions for this specific test to ensure it behaves predictably.

pkg/service/cluster/service.go
```diff
@@ -171,55 +175,124 @@ func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) er
...
+		wg.Wait()
+		close(result)
+	}()
+
+	// Read results until the channel is closed
+	for hosts := range result {
+		return hosts.known, hosts.live, nil
+	}
+
+	// If no valid results, return error<
```
Michal-Leszczynski, line 237:
```suggestion
	// If no valid results, return error
```

pkg/service/cluster/service.go
```diff
@@ -171,55 +175,124 @@ func (s *Service) discoverAndSetClusterHosts(ctx context.Context, c *Cluster) er
...
+	return nil, nil, ErrNoValidKnownHost
```
Michal-Leszczynski, line 238: Why do you want to iterate over results, if we just wait for a single one?
Wouldn't it be more idiomatic to just:

```go
hosts, ok := <- results
```

pkg/service/healthcheck/service_integration_test.go
Michal-Leszczynski: > That's an interesting point, but I think there's a distinction here. A node doesn't necessarily need to be fully "UP" to respond to some API calls—it could still be in a "Joining" state, for example.
Using nodetool status gives us a more reliable check to ensure the node is fully operational (UP) as part of the cluster. For this test, I also want to ensure the agent service is running and that ports are unblocked, which nodetool helps verify comprehensively.

Nodetool is just a simple client that uses Scylla API - very similar to what agent does. Agent can also issue status API call - the only problem here is that our implementation does not allow to specify host (SM kind of assumes that it doesn't matter which host will be chosen). But in order to deal with that, we can simply apply `ClientContextWithSelectedHost` (exported `forceHost`).

I just don't like using shell commands and parsing raw text when we already have a proper way of getting this information in our repo.

Also, the comment about using `parallel.Run` still stands.

> That's a valid point, and I agree it could simplify some workflows if we add the waiting step to the test that stops nodes. However, I still prefer to harden and explicitly check the assumptions for this specific test to ensure it behaves predictably.

I'm not against validating it here, but we could probably save some time in the future if we also validate it after `TestValidateHostConnectivityIntegration`. Otherwise, some newly written test might start failing on CI, when they pass locally, and it could be annoying to debug.

karol-kokoszka: @Michal-Leszczynski I'm not gonna change it. Let's leave it parallel and include `nodetool status`.

Michal-Leszczynski: What about the last comment regarding `TestValidateHostConnectivityIntegration`?

karol-kokoszka: I want this part with ensureNodesAreUP to help identify the root cause of the situation where a node is down, but it was expected to be up. Without it, the test error may contain misleading information that is just a result of the test assuming an incorrect cluster state.

So, it's just one step. It still doesn't point to the exact root cause.
The root cause lies in a previous test that was incorrectly implemented and didn’t clean up properly after execution. This means the test that failed was actually correct.

In this PR, this is exactly the case. TestStatus_Ping_Independent_From_REST_Integration was failing, but the reason for the failure was the TestValidateHostConnectivityIntegration test introduced in this PR. The former stopped failing after applying a fix to TestValidateHostConnectivityIntegration.

IMHO, it's better to be explicit in TestStatus_Ping_Independent_From_REST_Integration about the reason for the failure. In this case, the issue was that TestValidateHostConnectivityIntegration did not restart the service after execution.

By adding ensureNodesAreUP, it cuts the test execution and explicitly indicates that the Scylla service is not running.

What already existed in TestStatus_Ping_Independent_From_REST_Integration, such as:

```
tryUnblockCQL(t, ManagedClusterHosts())
tryUnblockREST(t, ManagedClusterHosts())
tryUnblockAlternator(t, ManagedClusterHosts())
tryStartAgent(t, ManagedClusterHosts())
```
... was not enough to terminate the execution early, which allowed the test to continue.

By creating a client and calling some random endpoint on it, I could end up in a situation where previous tests cleaned up after themselves, but since supervisorctl start is asynchronous, it was not fast enough to complete before the current test started.

karol-kokoszka: Code is updated to make the block/unblock/try methods public and to use them in TestValidateHost... tests as well.

pkg/service/cluster/service.go
karol-kokoszka: updated

pkg/service/cluster/service.go
karol-kokoszka: actually you're right, fixed

Michal-Leszczynski (approved):
LGTM!

PR #4080, opened 2024-10-25
Bump go to 1.23

Michal-Leszczynski:
This PR bumps go version to 1.23. Among other things, it has support for workspace vendoring.
It also removes the 'replace' directive to SM submodules (left there by mistake), and bumps
golangci-lint to 1.61.0, which has support for go 1.23.


Michal-Leszczynski:

karol-kokoszka (approved):
Are all of these changes related to bumping go and golint only ? 
I'm giving +1, please squash and merge, so that it will be easier to revert in case of any problems.

Michal-Leszczynski:
> Are all of these changes related to bumping go and golint only ?

Yes - bumping versions resulted in many linter errors - I divided linter related changes into commits, so that it's easier to go through them during review.

PR #4095, opened 2024-10-31
go.mod: remove replace directive

Michal-Leszczynski:
It is a left-over that I forgot to change after
developing changes across SM submodules.
This commit also bumps SM submodules deps.


karol-kokoszka:
Is it a duplicate of https://github.com/scylladb/scylla-manager/pull/4096 ?

Michal-Leszczynski:
> Is it a duplicate of https://github.com/scylladb/scylla-manager/pull/4096 ?

Actually yes, I completely forgot that such PR was already there.

PR #4088, opened 2024-10-29
Restore: add and fill host info in restore progress

Michal-Leszczynski:
This PR creates and fills new swagger host restore progress definitions from  #4082 in restore service.
It also updates managerclient to correctly display bandwidth and shard information in `sctool progress`.

Fixes #4042

PR #4096, opened 2024-11-04
go.mod: remove replace directive

Michal-Leszczynski:
It was a left-over from development.


karol-kokoszka (approved):
I guess you need to merge the bw display in progress first.

Michal-Leszczynski:
> I guess you need to merge the bw display in progress first.

The bw PR already contains commits which removes the replace directive, so it's better to merge this PR first and remove those commits from bw PR before merging.

karol-kokoszka:
Ok, seems that these changes were inlcuded into https://github.com/scylladb/scylla-manager/pull/4088
This PR is not valid anymore, closing.

PR #4099, opened 2024-11-05
go.mod: bump SM submodules deps before release

Michal-Leszczynski:


Michal-Leszczynski:
@karol-kokoszka This PR is ready for review! We should include it into the 3.4 release.

karol-kokoszka (approved):

PR #4107, opened 2024-11-12
feat(swagger): scylla, add read barier endpoint

VAveryanov8:
This adds scylla read barier api [endpoint](https://github.com/scylladb/scylladb/blob/6efde20939d072406b6e80a7ece1026134b0268a/api/api-doc/raft.json#L66-L97) which will be used in a backup process.

---

Please make sure that:


Michal-Leszczynski (approved):
LGTM!

Issue #3932, opened 2024-07-17
Use read barrier API


Michal-Leszczynski:
As described in https://github.com/scylladb/scylladb/issues/19213, we can now move from the "dummy schema change" workaround (also described in the mentioned issue) for Scylla versions starting at 6.1.
As this change has not been backported to Scylla 6.0, we still need to use the workaround there.

Michal-Leszczynski:
Context:
When making a backup, it's important that backed up schema is not older than backed up sstables (otherwise sstables cannot be correctly interpreted). That's why we first snapshot all interesting tables, and only after that we query schema via CQL (`DESCRIBE SCHEMA WITH INTERNALS`).
The problem is that this query is routed to some single node in the cluster and it is possible that the schema from this node is older than already snapshot-ed sstables from some other node. In order to fix this, we perform raft read barrier in between snapshot-ing sstables and querying schema (it makes sure that all nodes fetch the new schema). It was previously done by relying on some mechanism performed under the hood when Scylla was evaluating dummy schema changing query, but now we should move to the proper approach.

VAveryanov8:
> use this method instead of query.ReadBarrier when Scylla version is at least [6.1 or 2024.2](https://github.com/scylladb/scylla-manager/blob/bd6d1317d8f4770f9805152ed0713829cac0d737/pkg/scyllaclient/client_agent.go#L215)

hey @Michal-Leszczynski,  should we keep `query.ReadBarrier` method for Scylla version >= 6.0 then?

Michal-Leszczynski:
> hey @Michal-Leszczynski, should we keep query.ReadBarrier method for Scylla version >= 6.0 then?

We want to use new Scylla API for Scylla versions supporting it (>= 6.1 / >= 2024.2).
We want to use `query.ReadBarrier` when we back-up schema via cql and the Scylla API is not supported (>= 6.0 and < 6.1)

Michal-Leszczynski:
Fixed in #4110.

PR #4110, opened 2024-11-12
feat(backup): use raft read barier api for scylla >= 6.1, >= 2024.2

VAveryanov8:
This updates backup process to use raft read barier api when scylla
version is >= 6.1 (oss), >= 2024.2 (enterprise). Resolves issue https://github.com/scylladb/scylla-manager/issues/3932.

---

Please make sure that:


pkg/scyllaclient/client_scylla.go
```diff
@@ -1125,6 +1125,28 @@ func (c *Client) ControlTabletLoadBalancing(ctx context.Context, enabled bool) e
 	return err
 }
 
+// RaftReadBarier triggers read barrier for the given Raft group to wait for previously committed commands in this group to be applied locally.
+// For example, can be used on group 0 (default, when group is not provided) to wait for the node to obtain latest schema changes.
+func (c *Client) RaftReadBarier(ctx context.Context, groupID string, timeout int64) error {
+	params := operations.RaftReadBarrierPostParams{
+		Context: ctx,
+	}
```
Michal-Leszczynski, line 1133: `RaftReadBarier` -> `RaftReadBarrier` (small typo)

I think that I might have confused you earlier.
`RaftReadBarrier` has a scope of a single node - it ensures the node that gets this API call applies previously committed raft commands - but it does not ensure it for other nodes.
It means that the node that gets `RaftReadBarrier` API call and the node that we query for `DESC SCHEMA WITH INTERNALS` needs to be the same node (we use a single host gocql session in order to ensure that both `DESC SCHEMA` and cql raft read barrier were routed to the same node).

You can use `forceHost` in order to ensure that.

pkg/service/backup/worker_schema.go
```diff
@@ -174,22 +181,40 @@ func backupAndRestoreFromDescSchemaHosts(ctx context.Context, client *scyllaclie
...
-				mu.Unlock()
+			if method == "" {
+				return nil
+			}
+			mu.Lock()
+			defer mu.Unlock()
+			if _, ok := hostsByMethod[method]; !ok {
+				hostsByMethod[method] = []string{h}
+			} else {
+				hostsByMethod[method] = append(hostsByMethod[method], h)
```
Michal-Leszczynski, line 196: Can't we use just the `else` clause?

pkg/service/backup/worker_schema.go
```diff
@@ -72,6 +72,13 @@ func (w *worker) safeBackupAndRestoreSchemaDump(ctx context.Context, descSchemaH
 	return nil
 }
 
+func (w *worker) raftReadBarier(ctx context.Context, session gocqlx.Session, useReadBarierAPI bool) error {
```
Michal-Leszczynski, line 75: Why do you prefer to have `useReadBarierAPI` as a `bool` here?
It's nice that you created a dedicated type for raft read barrier methods, so why don't we use it here?

pkg/service/backup/worker_schema.go
```diff
@@ -72,6 +72,13 @@ func (w *worker) safeBackupAndRestoreSchemaDump(ctx context.Context, descSchemaH
...
+	if useReadBarierAPI {
+		return w.Client.RaftReadBarier(ctx, "", 0)
```
Michal-Leszczynski, line 77: Setting timeout to 0 results in setting it to 60s (taken from swagger docs).
On the other hand, scyllaclient has it's own timeout configuration.
I would say that it's better to align those timeouts (retries can probably stay the same).

You can use `customTimeout` for this.

Michal-Leszczynski (changes requested):
Nice work - especially custom types and tests! 
I left a few comments - the one regarding sending raft read barrier to specific host is important, others are more of a nits.

pkg/scyllaclient/client_scylla.go
VAveryanov8: Yes, make sense. For some reason I thought that client is already initialized to use particular host - I will fix that :)

pkg/service/backup/worker_schema.go
VAveryanov8: Yes, indeed! good catch!

pkg/service/backup/worker_schema.go
VAveryanov8: Thought it might be overkill here... but yeah, maybe it's better to use it here as well for the sake of consistency

pkg/service/backup/worker_schema.go
VAveryanov8: yeah, ok, I'll align them

BTW: where I can find config files with scyllaclient timeout defined?

VAveryanov8:
@Michal-Leszczynski I've updated the PR, please have a look

Michal-Leszczynski (approved):
LGTM!

PR #4111, opened 2024-11-13
TMP: test scylla backup api

Michal-Leszczynski:


karol-kokoszka:
@Michal-Leszczynski what is the status of this PR ?
Were you able to execute integration tests of Scylla Manager utilizing the new Scylla API for backup ?

Michal-Leszczynski:
> @Michal-Leszczynski what is the status of this PR ?
Were you able to execute integration tests of Scylla Manager utilizing the new Scylla API for backup ?

From the looks of it, all backup tests not related to pause/resume passed. On the other hand, restore tests (which also use backup)  failed. The implementation is still WIP, but it shouldn't take much time to finish it. I plan to do it after taking care of the remaining 3.4.1 items.

PR #4114, opened 2024-11-15
fix: handle ssl only scylla cluster setup

### LINKED ISSUE #4079: Scylla Manager, under certain condition, is unable to use only SSL port (9142) to restore data
scylla-manager and scylla-manager-agent : 3.2.8
Scylla version: 2024.1.9
### ISSUE:

While we were testing data restore through SSL (port 9142) we were unable to proceed due to below errors.

The whole cluster nodes are configured to use only SSL(9142 and 9042 is commented in scylla.yaml file in cluster's node).

scylla.yaml config in each scylla node
```
#native_transport_port: 9042 
native_transport_port_ssl: 9142

client_encryption_options:
    enabled: true
    certificate: /etc/ssl/db.crt
    keyfile: /etc/ssl/db.key
    truststore: /etc/ssl/cadb.pem

```
scylla node does not listen on 9042 port
```
[root@node3 ~]# netstat -ant | grep 9042
[root@node3 ~]# netstat -ant | grep 9142
tcp        0      0 192.168.254.85:19142    0.0.0.0:*               LISTEN
tcp        0      0 192.168.254.85:9142     0.0.0.0:*               LISTEN
```
sctool cluster add : works fine
sctool backup   : works fine
**sctool restore ... --restore-schema/--restore-tables : it fails with the same error**

```
# sctool restore -c test-cluster-ssl --snapshot-tag sm_20241024064123UTC --restore-schema  --location s3:scylla-backup
Error: create restore target, units and views: create worker: get CQL cluster session: gocql: unable to create session: unable to discover protocol version: dial tcp :0->192.168.254.48:9042: connect: connection refused
Trace ID: AkZV8BhoTPG1zy-wHT5qIg (grep in scylla-manager logs)
```

Even if we provide --ssl-user-cert-file and --ssl-user-key-file, it doesn't seem to be working.
Once we re-enable non-ssl port on scylla node (scylla.yaml), we're hit the following error.

```
root@scyllamanager:~# sctool restore -c test-cluster-ssl --snapshot-tag sm_20241024064123UTC --restore-schema  --location s3:scylla-backup
Error: create restore target, units and views: create worker: get CQL cluster session: gocql: unable to create session: unable to discover protocol version: tls: first record does not look like a TLS handshake
Trace ID: H9d3CeRGQb-gWXY_LHx8dA (grep in scylla-manager logs)
```
Next step, set **--force-tls-disabled=true**. Now, it does not complain about TLS anymore.
```
root@scyllamanager:~# sctool cluster update -c test-cluster-ssl --force-tls-disabled=true
# sctool restore -c test-cluster-ssl --snapshot-tag sm_20241024064123UTC  --restore-schema --location s3:scylla-backup
restore/4175e435-c52b-4aa1-a95e-927e4218c854
```

VAveryanov8:
This fixes how SM decides which port to use when connecting to Scylla
nodes.
Also adds SSL_ENABLED flag to Makefile, so that when you run
SSL_ENABLED=true make start-dev-env the scylla cluster will be created
with ssl_only config.

Fixes #4079 

---
Please make sure that:


pkg/service/cluster/service.go
```diff
@@ -617,9 +617,12 @@ func SingleHostSessionConfigOption(host string) SessionConfigOption {
 			return errors.Wrapf(err, "fetch node (%s) info", host)
 		}
 		cqlAddr := ni.CQLAddr(host)
+		if ni.ClientEncryptionEnabled {
```
VAveryanov8, line 620: I think here we should also consider cluster level ssl options like `ForceTLSDisabled` and `ForceNonSSLSessionPort`.

Michal-Leszczynski: Agree - I believe that it would be nicer to have just a single method that we use when we want to obtain valid CQL addr. It would need to accept mentioned SM flags as well. The current methods (`ni.CQLAddr`, `ni.CQLSSLAddr`) should be removed or un-exported so that we don't make this mistake in the future.
It would also result in de-duplicating some code.

Michal-Leszczynski: As you noticed, this fragment of code does not work correctly, as in some Scylla configurations it's possible to use SSL on a regula CQL port.

pkg/service/cluster/service.go
```diff
@@ -779,7 +773,11 @@ func GetRPCAddresses(ctx context.Context, client *scyllaclient.Client, hosts []s
 			combinedError = multierr.Append(combinedError, err)
 			continue
 		}
-		sessionHosts = append(sessionHosts, ni.CQLAddr(h))
+		addr := ni.CQLAddr(h)
+		if ni.ClientEncryptionEnabled && !clusterSSLDisabled {
+			addr = ni.CQLSSLAddr(h)
+		}
```
Michal-Leszczynski, line 779: We could use this new method choosing correct CQL address (either regular or SSL) here.

Michal-Leszczynski (changes requested):
Nice work!

pkg/service/cluster/service.go
VAveryanov8: Ok, thanks for your feedback,   I've replaced  CQLAddr and CQLSSLAddr with one function which takes into account cluster ssl setting

testing/scylla/config/scylla-ssl.yaml
```diff
@@ -1,4 +1,5 @@
 # Scylla SSL enabled configuration.
+# This file merged with testing/config/scylla.yaml during start-dev-env process when SSL_ENABLED=true.
```
karol-kokoszka, line 2:
```suggestion
# This file is merged with testing/config/scylla.yaml during start-dev-env process when SSL_ENABLED=true.
```

karol-kokoszka: And overall that's great that you introduce possibility to enable SSL in local env + test it !

testing/scylla/config/scylla-ssl.yaml
```diff
@@ -1,9 +1,652 @@
...
+# one logical cluster from joining another.
+# It is recommended to change the default value when creating a new cluster.
+# You can NOT modify this value for an existing cluster
+cluster_name: 'Managed Cluster'
+
+# This defines the number of tokens randomly assigned to this node on the ring
+# The more tokens, relative to other nodes, the larger the proportion of data
+# that this node will store. You probably want all nodes to have the same number
+# of tokens assuming they have equal hardware capability.
+num_tokens: 256
```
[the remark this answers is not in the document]
Michal-Leszczynski: Nice! I like the current solution.

pkg/service/cluster/service.go
Michal-Leszczynski: Thanks!

pkg/service/cluster/service.go
```diff
@@ -610,16 +610,16 @@ func (s *Service) ListNodes(ctx context.Context, clusterID uuid.UUID) ([]Node, e
...
 		ni, err := client.NodeInfo(ctx, host)
 		if err != nil {
 			return errors.Wrapf(err, "fetch node (%s) info", host)
 		}
-		cqlAddr := ni.CQLAddr(host)
+		cqlAddr := ni.CQLAddr(host, clusterTLSAddrDisabled)
 		cfg.Hosts = []string{cqlAddr}
-		cfg.HostFilter = gocql.WhiteListHostFilter(cqlAddr)
 		cfg.DisableInitialHostLookup = true
+		cfg.HostFilter = gocql.WhiteListHostFilter(cqlAddr)
```
Michal-Leszczynski, line 622: I don't like the fact that `GetSession` chooses correct CQL addr on its own, but when we want to create it with `SingleHostSessionConfigOption` we have to specify `clusterTLSAddrDisabled` separately. I think that it should be taken care of by the cluster service - especially that it has enough info to do it.

The signature of the `SessionConfigOption` is the problem here, but we can change it to:
```go
type SessionConfigOption func(ctx context.Context, cluster *Cluster, client *scyllaclient.Client, cfg *gocql.ClusterConfig) error
```

This has the following benefits:
- we don't need to pass cluster config to other services (e.g. backup)
- we limit the chances of creating incorrect single host session in the future

Please let me know your opinion on that!

Makefile
```diff
@@ -20,6 +20,8 @@ SCYLLA_VERSION?=scylla:6.0.1
 IP_FAMILY?=IPV4
 RAFT_SCHEMA?=none
 TABLETS?=enabled
+# if true starts the scylla cluster with ssl only config
+SSL_ENABLED?=false
```
Michal-Leszczynski, line 24: I would say that we should have it enabled by default, as this is the more common user setup.
This brings me to our CI - the current checks passed only for disabled SSL.
It would be good if they run on both enabled and disabled SSL.
Please take a look at our gh actions config:
- https://github.com/scylladb/scylla-manager/blob/master/.github/actions/test-setup/action.yml sets up testing env
- https://github.com/scylladb/scylla-manager/tree/master/.github/cfg generates workflows based on the config file

Workflows should support `SSL_ENABLED` input. For most workflows it should be set to true (as in default), but we can set it to false in some of them. Note that we can't test the whole cofig matrix, as it would require too many workflows, but it's enough to have some cases with `SSL_ENABLED` set to false. For example:

```yam
...
- scylla-version: scylla-enterprise-nightly:latest-enterprise
  ip-family: IPV4
  raft-schema: none
  tablets: disabled
  ssl-enabled: false

- scylla-version: scylla-enterprise:2024.1.12
  ip-family: IPV6
  raft-schema: none
  tablets: none
  ssl-enabled: false
...
```

pkg/service/cluster/service.go
Michal-Leszczynski: We currently have only a single option for `SessionConfigOption`, so let's hope the scope won't be too big!

VAveryanov8: yes, it was easier then I thought :)

Makefile
VAveryanov8: makes sense! I've updated CI and made several fixes to our integration tests, because some of them assumed that the cluster could only be configured without SSL.

Michal-Leszczynski (approved):
The SSL setup (both locally and on gh actions) will be really useful, thanks!

Issue #4109, opened 2024-11-12
Proposal/Discussion: How to manage v3/swagger (and maybe other v3/* packages)


VAveryanov8:
As far as I understand `v3/*` contains go modules (and packages) that can be used not only by Scylla Manager, but by some other projects (siren?). 

Current integration with Scylla Manager is done as with any regular external dependency - entry in go.mod that points to repository url. 

**Advantages:**
1. SM can use different version of `v3/*` packages (Do we need it?) 

**Disadvantages:**
- If you need to change something in `v3/*` package and use your changes right away in SM, then it's required to complete two steps in general. Here is an **example:** imaging you need to use new api from scylla db, then
  1. Copy swagger definitions from Scylla to Scylla Manager repo, generate client and propagate changes to master branch
  2. In order to use latest changes it's required to update go.mod with latest version of `v3/swagger`from master branch
- Updating `v3/*` dependency may force automatic update of other transitive dependencies (minimal version selection) 

**What we can do differently:**

use `replace` directive in go.mod pointing to a local version of `v3/*` packages, e.g.
```
...
replace (
    github.com/scylladb/scylla-manager/v3/swagger => ./v3/swagger
)
...
```

Then SM is always using local(latest) version of `v3/swagger`, but other projects are relying on versioning of SM. 

What do you think? Maybe I'm mussing some context here 

P.S.
Similar method is used in kubernetes project: 
https://github.com/kubernetes/kubernetes/blob/6b031e50b2abfe15434beef6c7e01056091a26ea/go.mod#L225 
https://github.com/kubernetes/kubernetes/tree/master/staging#external-repository-staging-area

Michal-Leszczynski:
> As far as I understand v3/* contains go modules (and packages) that can be used not only by Scylla Manager, but by some other projects (siren?).

Correct - although I believe that we only care for the `managerclient` module to be exposed.
Note that `managerclient` is importing the `swagger` module as well.
@karol-kokoszka could you describe why we have separate `swagger` and `utils` modules?

> SM can use different version of v3/* packages (Do we need it?)

We don't.

>  Similar method is used in kubernetes project:

I see that they are also using the go workspaces alongside manual replace directives - I don't really know why - I will try to learn a little bit more about that.

But in general I agree that working on cross module changes is really annoying - right now I'm using local go workspaces, but it would be nicer to have a solution that's integrated with the repository.

VAveryanov8:
superseded by #4121

PR #4115, opened 2024-11-15
Proposal: add replace directive to go.mod to use local version of v3/* modules

VAveryanov8:
This simplifies workflow when it's needed to make some changes in v3/* modules.

Relates to #4109 

---

Please make sure that:


VAveryanov8:
Ok, here is how it works: 
1. From the SM perspective, when scylla-manager repository is pulled, then `scylla-manager/go.mod` file is the main module and all go tooling will consider `replace` directives of `scylla-manger/go.mod`. 
2. When some external module (e.g. siren) is using `v3/pkg/managerclient`, then main module is `siren/go.mod` and `replace` directives of `scylla-manager/go.mod` and `v3/pkg/managerclient/go.mod` are ignored, but instead go tooling uses versions from their `require` sections. 

Here is short example to show this behavior. Imaging we have some `managerclient_test` module, looking like this: 
**main.go:**
```go
package main

import (
	"github.com/scylladb/scylla-manager/v3/pkg/managerclient"
)

func main() {
	_, _ = managerclient.NewClient("example.com")
}

```
We pulled managerclient package from this branch (by commit hash): 
```bash 
go get github.com/scylladb/scylla-manager/v3/pkg/managerclient@0cfb5ab6da3bdc631afc138cbf4c6ba2d7cba24a 
``` 

Then go.mod will look like this (i've stripped all the deps except scylla-manager/v3*): 
```go
module managerclienttest

go 1.23.2

require github.com/scylladb/scylla-manager/v3/pkg/managerclient v0.0.0-20241118090522-0cfb5ab6da3b

require (
	....
	github.com/scylladb/scylla-manager/v3/pkg/util v0.0.0-20241030073626-e409ae491c83 // indirect
	github.com/scylladb/scylla-manager/v3/swagger v0.0.0-20241030073626-e409ae491c83 // indirect
	...
)

```

We can see that managerclient version is from this branch, but `v3/pkg/util` and `v3/swagger` have version that are listed in `require` section of `managerclient/go.mod` from this branch. 

This means, that if we want to update managerclient dependencies for the external usage, then we need to explicitly update `managerclient/go.mod`, but when working on `scylla-manger` only local versions of `v3/*` modules will be used and there is no need to do several pull requests when we need to update `v3/swagger` for example.  

P.S.
This behavior is described in [go.mod reference](https://go.dev/ref/mod#go-mod-file-replace)

> replace directives only apply in the main module’s go.mod file and are ignored in other modules. See [Minimal version selection](https://go.dev/ref/mod#minimal-version-selection) for details.

VAveryanov8:
> So if I understand correctly, we should use replace directives only in the scylla manager module go.mod file, but not in the managerclient module go.mod file?

`replace` directive only applies to main module, so if (in theory) we want to use `managerclient` as a main module, e.g adding `main` package  in `managerclient`, then there is a place for `replace` directive, otherwise indeed it's not needed.

> Also, this could still create problems when someone would want to import scylla manager module.  

You mean problems like build version of scylla-manager is different from was what imported as module, because of the different dependencies version in go.mod?  Yes, additional effort to update go.mod file is still  be required in any case I think :( 

So in a nutshell by adding `replace` directive we can simplify the way we work with scylla manager, when making changes in both `v3/*` modules and main scylla-manager module. But still the additional effort to keep go.mod file updated is required.

BTW: do we keep managerclient/go.mod up to date with latest versions of `v3/swagger` and `v3/pkg/util`?

karol-kokoszka:
Maybe go.work is not a bad option at all (even if it's commit to the repo).

Issue #3742, opened 2024-03-03
docs: Issue in page Restore


pdbossman:
I would like to report an issue in page https://manager.docs.scylladb.com/branch-3.2/restore/index

### Problem
This comment is on the overall restore procedure, and changes may be on this page or others. 

When I am doing a restore, I don't want backup or repair to interfere with it.  But there are no instructions about whether Scylla Manager will handle that for me or not.

Can I suspend scylla manager and still run a restore?  I doubt it, and it's unclear if this is desired/designed for.
In the absense of clear guidance, I've been modifying scheduled repair/backup tasks with --enabled=false

###  Suggest a fix

Please update the procedure with clear guidance on what to do about other scylla manager tasks during a restore.
Indicate whether anything or nothing needs to be done, and if nothing needs to be done regarding backup/repairs, explain how Scylla Manager will behave to set the person executing the procedures mind at ease.

PR #4124, opened 2024-11-25
docs: adds info about restore and backup/repair tasks interference

VAveryanov8:
This extends restore procedure docs with how to avoid conflicts between restore and backup/repair tasks in a cluster.

Fixes: #3742

---

Please make sure that:

### DIFF
--- a/docs/source/restore/index.rst
+++ b/docs/source/restore/index.rst
@@ -89,14 +89,15 @@ and :ref:`resume <task-start>` it.
 Interference with backup and repair tasks
 =========================================
 
-It is advisable to manually make sure that the restore task does not interfere with backup or repair tasks.
+The restore task will not start if backup or repair tasks are running in the cluster at that time, and vice versa, if a restore task is running, backup or repair tasks will not start. 
+You will see an error in the output of the ``sctool tasks`` command if there is a conflict between tasks. 
+Here is an example output of ``sctool progress -c ${cluster_name} {task_id}`` when there is an interference between tasks::
 
-#. Do not start backup or repair tasks when restore task is running.
-#. If scheduled runs of backup or repair tasks can interfier with restore task, then disable them until restore is complete::
-
-    # check Next column to determine when next run of a task is scheduled 
-    $ sctool tasks -a
-
-    # disable repair/backup tasks if needed
-    $ sctool repair update -c CLUSTER-ID SCHEDULED-REPAIR-TASK-ID --enabled=false
-    $ sctool backup update -c CLUSTER-ID SCHEDULED-BACKUP-TASK-ID --enabled=false
+    $  ./sctool.dev progress -c my-cluster backup/06cd5d90-6daa-4215-acd3-d19f6782b5b6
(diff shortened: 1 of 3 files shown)


VAveryanov8:
I'm not sure which SM version will contain changes made in #4122, so I've put 3.5 version in the docs temporarily.

docs/source/restore/index.rst
```diff
@@ -85,3 +85,20 @@ Restore speed is controlled by many parameters (see :ref:`sctool restore <sctool
 The ``--unpin-agent-cpu`` is disabled by default, but in case you observe small download
 bandwidth, you could try to :ref:`pause <task-stop>` restore task, :ref:`update <restore-update>` it with ``--unpin-agent-cpu``,
 and :ref:`resume <task-start>` it.
+
+Interference with backup and repair tasks
+=========================================
+
+Starting from Scylla Manager v3.5.0, the restore task will not start if backup or repair tasks are running in the cluster at that time, and vice versa, if backup or repair tasks are running, the restore task will not start.
+
+Prior to 3.5.0, it is advisable to manually make sure that the restore task does not interfere with backup or repair tasks.
```
Michal-Leszczynski, line 94: SM docs are versioned per minor version, so there is no need to write things like "starting from version".
On the other hand, I see that you wanted to document previous behavior as well.
So the proper way of handling this would be to create 2 separate PRs:
- one adding docs about current state (merged to master and branch-3.4)
- one updating docs with new functionality description (rebased on the first one, merged to master)

Michal-Leszczynski:
> I'm not sure which SM version will contain changes made in https://github.com/scylladb/scylla-manager/pull/4122, so I've put 3.5 version in the docs temporarily.

3.5 is a good candidate - we shouldn't put it into 3.4.1 as this is a feature.

Michal-Leszczynski:
@VAveryanov8 is this PR ready for re-review?
In general, if you wait for my re-review, it will be faster if you request the re-review in the reviewers section on github or by pinging me directly.

docs/source/restore/index.rst
VAveryanov8: yes, that make sense to me! I've split this pr into two part. This one is related to documenting new behavior and #4147 about current(old) behavior

VAveryanov8:
@Michal-Leszczynski  btw: do we use labels to set the target version for PRs?

docs/source/restore/index.rst
Michal-Leszczynski: Thanks for the additional effort!

Michal-Leszczynski:
> @Michal-Leszczynski btw: do we use labels to set the target version for PRs?

In theory we should. I was talking about it during our meeting, but then I figured that if we stick to SEMVER, then patch releases would be really rare for SM, so hand picking commits for them does not seem that bad.

Michal-Leszczynski (approved):
Thanks!

Michal-Leszczynski:
@VAveryanov8 the other PR documenting previous behavior has been merged.
Please rebase this one so it can be merged as well.

PR #4145, opened 2024-12-02
PoC for 1-1 restore (VNODES)

karol-kokoszka:
This PR is just to explore the 1-1 restore.

`.1-1-restore-poc/256000_rows_backuptest_data_big_table.tar` is the archived snapshot.
To execute the PoC that restores simple backuptest_data.big_table table:

0. Clean the scylla data folders on containers.
```
make clean_restore
```
2. Extract information about the token ring from the backup manifest
```
make extract_tokens_and_schema
```
3. Set initial_token to the `scylla.yaml` per every node.
```
make set_initial_tokens
```
4. Start the cluster
```
make start-dev-env
```
5. Recreate schema
```
make recreate_schema
```
6. Extract SSTables from archive
```
make extract_sstables
```
7. Copy nodes SSTables to the /var/lib/scylla/data 
```
make copy_sstables
```
8. Refresh the restored table.
```
make refresh_nodes
```
9. Query the amount of rows
```
make query_data
```

256000 rows are restored.
Calling `make query_data` before refreshing the table results with `0 rows available`.

---

Please make sure that:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -257,3 +257,212 @@ vendor: ## Fix dependencies and make vendored copies
 .PHONY: help
 help:
 	@awk -F ':|##' '/^[^\t].+?:.*?##/ {printf "\033[36m%-25s\033[0m %s\n", $$1, $$NF}' $(MAKEFILE_LIST)
+
+.PHONY: extract_tokens_and_schema
+extract_tokens_and_schema:
+	@echo "Extracting tokens from .1-1-restore-poc/256000_rows_backuptest_data_big_table.tar..."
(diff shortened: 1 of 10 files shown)


karol-kokoszka:
Current scenario covers VNODEs only.
Another scenario with tablets must be created and tested (Is it enough to restore system tables related to tablets ?).

karol-kokoszka:
UPDATE: This solution won't work for tablets and is not expected to work with tablets.
For tablet keyspaces, we will failback to load and stream.

PR #4146, opened 2024-12-02
Rclone: stream files as they are being listed

### LINKED ISSUE #4132: RcloneListDirIter: call callback before listing the whole dir
The idea of the `RcloneListDirIter` implementation on the agent/rclone side is that it should stream dir entries on the fly.
This approach has a few benefits:
- easier timeout handling
- better performance
- less memory pressure on the agent with limited resources (no need to store all dir entries in the memory - as both golang objects and json encodings)

The problem is that currently the "stream dir entries on the fly" part is done with the whole dir granularit. 
Take a look at the [rclone listing code](https://github.com/scylladb/rclone/blob/afe1fd2aa65ea4806541047752bb39a21ba72bcf/fs/walk/walk.go#L395-L408):
```go
entries, err := listDir(ctx, f, includeAll, job.remote) // <- lists the whole dir
var jobs []listJob
if err == nil && job.depth != 0 {
	entries.ForDir(func(dir fs.Directory) {
		// Recurse for the directory
		jobs = append(jobs, listJob{
			remote: dir.Remote(),
			depth:  job.depth - 1,
		})
	})
}
mu.Lock()
err = fn(job.remote, entries, err) // <- callback function responsible for streaming dir entries to SM
mu.Unlock()
```
It is called only after listing all entries in the dir, so when listing just a single large dir (which is the most common way of using `RcloneListDirIter` during backup), it totally nullifies the "stream dir entries on the fly".

On the other hand, most backend's APIs naturally support chunked listing. E.g., by default we list up to 1000 objects with a single API call to s3. Take a look at the [rclone s3 listing code](https://github.com/scylladb/rclone/blob/afe1fd2aa65ea4806541047752bb39a21ba72bcf/fs/walk/walk.go#L395-L408):
```go
for {
	// FIXME need to implement ALL loop
	req := s3.ListObjectsInput{
		Bucket:    &bucket,
		Delimiter: &delimiter,
		Prefix:    &directory,
		MaxKeys:   &f.opt.ListChunk, // <- 1000 by default
		Marker:    marker,
	}
...
	if !aws.BoolValue(resp.IsTruncated) { // <- Keep on making API calls until all objects have been listed
		break
	}
```

So in order to make `RcloneListDirIter` useful, the agent/rclone implementation should call the callback after each chunked call to s3, and not only after listing the whole dir.

Ref: https://github.com/scylladb/scylla-enterprise/issues/4861

Michal-Leszczynski:
This PR changes the rclone version used by SM-agent server.
The goal is improve huge dir listing by streaming the files to the SM as the dir is being listed. Rclone lists files in chunks (1000 is the default value for s3), so we can start streaming those files after each chunk has been listed, not only after the whole dir has been listed.
This allows for more reliable timeout handling on SM side, and also decreases memory pressure on agent.

Since this is timeout/performance change, I didn't prepare a dedicated test for it, but I created a benchmark for that (see last [commit](https://github.com/scylladb/scylla-manager/pull/4146/commits/e3fe0c9464f8c151920761b73eb8b23716f444bb) for details). It measures some interesting averages when listing a flat dir with 5555 files 1000 times. Note that this has been tested on my local docker setup without any rate limiting, but I still think that the results show what needs to be seen.

The results without the fix:
```
Avg total time: 157.916848ms           (average time of listing the whole dir)
Avg first latency: 146.564896ms        (average time after the first item has been listed)
Avg max cross chunk latency: 19.196µs  (average time between listing the last item from given chunk and the first from the next chunk)
Avg max within chunk latency: 112.92µs (average time between listing files from the same chunk)
```
The results after the fix:
```
Avg total time: 143.211603ms
Avg first latency: 26.344503ms
Avg max cross chunk latency: 26.144269ms
Avg max within chunk latency: 320.72µs
```

What is important, is that this fix reduces the time needed for listing the first item from `~146ms` to `~26ms`. Let's have in mind that 5555 files is not a huge amount for backed up sstables, and that the local setup has faster connections than the real one. It also slightly reduces the total time needed for listing, but the values are too similar to tell it for sure.

Fixes  #4132

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -127,6 +127,6 @@ require (
 
 replace (
 	github.com/gocql/gocql => github.com/scylladb/gocql v1.12.0
-	github.com/rclone/rclone => github.com/scylladb/rclone v1.54.1-0.20240312172628-afe1fd2aa65e
+	github.com/rclone/rclone => github.com/scylladb/rclone v1.54.1-0.20241203155938-9f3550e7e668
 	google.golang.org/api v0.114.0 => github.com/scylladb/google-api-go-client v0.34.1-patched
 )
--- a/go.sum
+++ b/go.sum
@@ -1051,8 +1051,8 @@ github.com/scylladb/gocqlx/v2 v2.8.0 h1:f/oIgoEPjKDKd+RIoeHqexsIQVIbalVmT+axwvUq
 github.com/scylladb/gocqlx/v2 v2.8.0/go.mod h1:4/+cga34PVqjhgSoo5Nr2fX1MQIqZB5eCE5DK4xeDig=
 github.com/scylladb/google-api-go-client v0.34.1-patched h1:DW+T0HA+74o6FDr3TFzVwgESabOB1eTwb4woE6oUziY=
 github.com/scylladb/google-api-go-client v0.34.1-patched/go.mod h1:RriRmS2wJXH+2yd9PRTEcR380U9AXmurWwznqVhzsSc=
-github.com/scylladb/rclone v1.54.1-0.20240312172628-afe1fd2aa65e h1:lJRphCtu+nKd+mfo8whOTeFkgjMWvk8iCSlqgibKSa8=
-github.com/scylladb/rclone v1.54.1-0.20240312172628-afe1fd2aa65e/go.mod h1:JGZp4EvCUK+6AM1Fe1dye5xvihTc/Bk0WnHHSCJOePM=
+github.com/scylladb/rclone v1.54.1-0.20241203155938-9f3550e7e668 h1:3beIciiaygPjiUPLN0+wVlnfe4JnB1Wb/uV1MChzRbM=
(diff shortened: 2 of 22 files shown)


VAveryanov8:
Nice benchmark! LGTM! 

As I understand all the heavy lifting was done in scylladb/rclone repository?

VAveryanov8 (approved):

karol-kokoszka (approved):
:+1: 

Please monitor https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/ SCT tests after merging.
It may be even worth to start the tests just on this branch.

Michal-Leszczynski:
> As I understand all the heavy lifting was done in scylladb/rclone repository?

Correct, but we treat it more as a playground, so it would be good to review vendored changes as a part of this PR.

mykaul:
Is this another private patch on top of our forked rclone?

karol-kokoszka:
> It turned out that we are using recursive RcloneListDirIter for that (@karol-kokoszka I think that you were implementing it - any idea why the recursive version was chosen)?

I don't think there is any special cause that lead to choose recursive over the flat approach.

PR #4151, opened 2024-12-10
Unify workflows between master and branch

Michal-Leszczynski:
`master` and `branch-**` have different gh actions specification due to ac49dff8.
We should avoid that because:
- it's problematic to keep track of it
- it causes conflicts on patch releases


karol-kokoszka (approved):

VAveryanov8 (approved):
LGTM!

Issue #4127, opened 2024-11-27
Possibility to query instance metadata for AWS


karol-kokoszka:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#instancedata-inside-access
IMDSv2 must be supported.

PR #4154, opened 2024-12-10
feat: adds cloudmeta package with AWS provider

VAveryanov8:
This adds basic interface and models to work with different cloud metadata providers and also adds implementation for aws metadata provider. 
This package will be used in order to extend backup manifest with additional information. 

This changes has been manually tested with the help of https://github.com/aws/amazon-ec2-metadata-mock container. 

Fixes: #4127 

---

Please make sure that:

### DIFF
--- a/pkg/cloudmeta/aws.go
+++ b/pkg/cloudmeta/aws.go
@@ -5,46 +5,29 @@ package cloudmeta
 import (
 	"context"
 
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/pkg/errors"
 )
 
-// AWSMetadata is a wrapper around ec2 metadata client.
-type AWSMetadata struct {
+// awsMetadata is a wrapper around ec2 metadata client.
+type awsMetadata struct {
 	ec2meta *ec2metadata.EC2Metadata
 }
 
-// NewAWSMetadata is a constructor for  AWSMetadata service.
-func NewAWSMetadata() (*AWSMetadata, error) {
+// newAWSMetadata is a constructor for  AWSMetadata service.
+func newAWSMetadata() (*awsMetadata, error) {
 	session, err := session.NewSession()
 	if err != nil {
 		return nil, errors.Wrap(err, "session.NewSession")
 	}
-	return &AWSMetadata{
+	return &awsMetadata{
 		ec2meta: ec2metadata.New(session),
 	}, nil
 }
 
-// NewAWSMetadataWithEndpoint is a constructor for  AWSMetadata service that allows your to overwrite default metadata endpoint.
-// Might be useful for tests.
-func NewAWSMetadataWithEndpoint(endpoint string) (*AWSMetadata, error) {
-	session, err := session.NewSession()
-	if err != nil {
-		return nil, errors.Wrap(err, "session.NewSession")
-	}
-	cfg := aws.NewConfig()
-	if endpoint != "" {
-		cfg = cfg.WithEndpoint(endpoint)
-	}
-	return &AWSMetadata{
-		ec2meta: ec2metadata.New(session, cfg),
-	}, nil
-}
-
 // Metadata return InstanceMetadata from aws if available.
-func (aws *AWSMetadata) Metadata(ctx context.Context) (InstanceMetadata, error) {
+func (aws *awsMetadata) Metadata(ctx context.Context) (InstanceMetadata, error) {
 	if !aws.ec2meta.AvailableWithContext(ctx) {
 		return InstanceMetadata{}, errors.New("metadata is not available")
 	}
--- a/pkg/cloudmeta/aws.go
+++ b/pkg/cloudmeta/aws.go
@@ -1,0 +1,50 @@
+// Copyright (C) 2024 ScyllaDB
+
+package cloudmeta
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/pkg/errors"
+)
+
+// AWSMetadata is a wrapper around ec2 metadata client.
+type AWSMetadata struct {
+	ec2meta *ec2metadata.EC2Metadata
+}
+
+// NewAWSMetadata is a constructor for  AWSMetadata service.
+// testEndpoint can be provided if you want to overwrite the default metadata endpoint, otherwise leave it empty.
+func NewAWSMetadata(testEndpoint string) (*AWSMetadata, error) {
+	session, err := session.NewSession()
+	if err != nil {
+		return nil, errors.Wrap(err, "session.NewSession")
+	}
+	cfg := aws.NewConfig()
+	if testEndpoint != "" {
+		cfg = cfg.WithEndpoint(testEndpoint)
+	}
+	return &AWSMetadata{
+		ec2meta: ec2metadata.New(session, cfg),
+	}, nil
+}
+
+// Metadata return InstanceMetadata from aws if available.
+func (aws *AWSMetadata) Metadata(ctx context.Context) (InstanceMetadata, error) {
+	if !aws.ec2meta.AvailableWithContext(ctx) {
+		return InstanceMetadata{}, errors.New("metadata is not available")
+	}
+
+	instanceData, err := aws.ec2meta.GetInstanceIdentityDocumentWithContext(ctx)
+	if err != nil {
+		return InstanceMetadata{}, errors.Wrap(err, "aws.metadataClient.GetInstanceIdentityDocument")
+	}
+
+	return InstanceMetadata{
+		CloudProvider: CloudProviderAWS,
+		InstanceType:  instanceData.InstanceType,
+	}, nil
+}
--- a/pkg/cloudmeta/aws.go
+++ b/pkg/cloudmeta/aws.go
@@ -17,15 +17,26 @@ type AWSMetadata struct {
 }
 
 // NewAWSMetadata is a constructor for  AWSMetadata service.
-// testEndpoint can be provided if you want to overwrite the default metadata endpoint, otherwise leave it empty.
-func NewAWSMetadata(testEndpoint string) (*AWSMetadata, error) {
+func NewAWSMetadata() (*AWSMetadata, error) {
+	session, err := session.NewSession()
+	if err != nil {
+		return nil, errors.Wrap(err, "session.NewSession")
+	}
+	return &AWSMetadata{
+		ec2meta: ec2metadata.New(session),
+	}, nil
+}
+
+// NewAWSMetadataWithEndpoint is a constructor for  AWSMetadata service that allows your to overwrite default metadata endpoint.
+// Might be useful for tests.
+func NewAWSMetadataWithEndpoint(endpoint string) (*AWSMetadata, error) {
 	session, err := session.NewSession()
 	if err != nil {
 		return nil, errors.Wrap(err, "session.NewSession")
 	}
 	cfg := aws.NewConfig()
-	if testEndpoint != "" {
-		cfg = cfg.WithEndpoint(testEndpoint)
+	if endpoint != "" {
+		cfg = cfg.WithEndpoint(endpoint)
 	}
 	return &AWSMetadata{
 		ec2meta: ec2metadata.New(session, cfg),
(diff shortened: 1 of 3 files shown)


VAveryanov8:
integration tests are still running, but this is a completely new package,  so it shouldn't affect existing functionality in any case :)

pkg/cloudmeta/metadata.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return &CloudMeta{
+		providers: []CloudMetadataProvider{
+			awsMeta,
+		},
+		providerTimeout: defaultTimeout,
+	}, nil
+}
+
+// GetInstanceMetadata tries to fetch instance metadata from AWS, GCP, Azure providers in order.
```
Michal-Leszczynski, line 52: Is the order important? 
In theory we could query the metadata in parallel - which is not needed for this use case, but why would we specify that?

pkg/cloudmeta/metadata_test.go
```diff
@@ -0,0 +1,140 @@
...
+			t.Fatalf("unexpected err: %v", err)
+		}
+		if meta.InstanceType != "" {
+			t.Fatalf("meta.InstanceType should be empty, got %v", meta.InstanceType)
+		}
+
+		if meta.CloudProvider != "" {
+			t.Fatalf("meta.CloudProvider should be empty, got %v", meta.CloudProvider)
+		}
+	})
```
Michal-Leszczynski, line 26: Cool that you are writing detailed tests!

The lack of empty line between 18-19 and an empty line 22 seems inconsistent?

I would say that it's safer and more intuitive to return an error when no provider is specified and the instance metadata is queried. It most likely indicates an error in code development. We can make it a named error so that it can be handled accordingly if needed.
What is you opinion on that?

pkg/cloudmeta/aws.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// AWSMetadata is a wrapper around ec2 metadata client.
+type AWSMetadata struct {
+	ec2meta *ec2metadata.EC2Metadata
+}
+
+// NewAWSMetadata is a constructor for  AWSMetadata service.
+// testEndpoint can be provided if you want to overwrite the default metadata endpoint, otherwise leave it empty.
+func NewAWSMetadata(testEndpoint string) (*AWSMetadata, error) {
```
Michal-Leszczynski, line 21: What do you think about changing `testEndpoint` to just `endpoint`?
It can still be specified in the comment, that `endpoint` should be used mostly for testing.
It just looks strange to me to see a "test" argument in the regular code base directly.

Or perhaps it would be nicer to have a separate method for `NewAWSMetadataWithEndpoint` which would share most of the body with `NewAwsMetadata`?

No strong opinions on that.

Michal-Leszczynski (commented):
Nice work!
Just left a few NITs.

Michal-Leszczynski:
Also in the context of this [comment](https://github.com/scylladb/scylla-manager/pull/4159#pullrequestreview-2498746302), does AWS SDK handle re-tries by default?

pkg/cloudmeta/metadata.go
VAveryanov8: I think it's not important, but I decided for now to keep them in order of popularity, so we the most popular one is queried first and etc ... so basically I just specified what behavior to expect from this function. 

yes, we may want to change that later, but it depends on how this package will be used. 
There is also a possibility (theoretical) that we may want to enable/disable some providers based on configuration, e.g. query only aws or gcp. 

let me know if doesn't make sense for you

pkg/cloudmeta/aws.go
VAveryanov8: yes, make sense, it may look confusing without having actual test in the commit :) 
I'll update this part.

VAveryanov8:
> Also in the context of this https://github.com/scylladb/scylla-manager/pull/4159#pullrequestreview-2498746302, does AWS SDK handle re-tries by default?

good question! yes, by default ec2metadata client will have 2 retires - https://github.com/scylladb/scylla-manager/blob/master/vendor/github.com/aws/aws-sdk-go/aws/ec2metadata/service.go#L94

pkg/cloudmeta/metadata_test.go
VAveryanov8: yep, agree, it's better to return error!

pkg/cloudmeta/aws.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, errors.Wrap(err, "session.NewSession")
+	}
+	cfg := aws.NewConfig()
+	if endpoint != "" {
+		cfg = cfg.WithEndpoint(endpoint)
+	}
+	return &AWSMetadata{
+		ec2meta: ec2metadata.New(session, cfg),
+	}, nil
+}
```
karol-kokoszka, line 44: I understand you want to introduce some flexibility in terms of choosing different endpoint of AWS service, but why only endpoint ? 
I suggest you to accept `aws.Config` instead of just endpoint.

Besides it's more idiomating to pass whole configuration to constructor in go, instead of small pieces and hardcoding the rest.

pkg/cloudmeta/aws.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Metadata return InstanceMetadata from aws if available.
+func (aws *AWSMetadata) Metadata(ctx context.Context) (InstanceMetadata, error) {
+	if !aws.ec2meta.AvailableWithContext(ctx) {
+		return InstanceMetadata{}, errors.New("metadata is not available")
```
karol-kokoszka, line 49: Can you extract `errors.New("metadata is not available")` to exported var ErrMetadataNotAvailable ?

karol-kokoszka: You need to call errors.WithStackTrace(ErrMetadataNotAvailable) later on, when returning this error.

pkg/cloudmeta/metadata.go
```diff
@@ -0,0 +1,83 @@
+// Copyright (C) 2024 ScyllaDB
+
+package cloudmeta
+
+import (
+	"context"
+	"errors"
```
karol-kokoszka, line 7: Can you use "github.com/pkg/errors" instead ?

pkg/cloudmeta/metadata.go
```diff
@@ -0,0 +1,83 @@
...
+	}, nil
+}
+
+// ErrNoProviders will be returned by CloudMeta service, when it hasn't been initialized with any metadata provider.
+var ErrNoProviders = errors.New("no metadata providers found")
+
+// GetInstanceMetadata tries to fetch instance metadata from AWS, GCP, Azure providers in order.
+func (cloud *CloudMeta) GetInstanceMetadata(ctx context.Context) (InstanceMetadata, error) {
+	if len(cloud.providers) == 0 {
+		return InstanceMetadata{}, ErrNoProviders
```
karol-kokoszka, line 58: Please wrap the error with errors.WithStack

pkg/cloudmeta/metadata.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, provider := range cloud.providers {
+		meta, err := cloud.runWithTimeout(ctx, provider)
+		if err != nil {
+			mErr = errors.Join(mErr, err)
+			continue
+		}
+		return meta, nil
+	}
+
+	return InstanceMetadata{}, mErr
```
karol-kokoszka, line 70: It can be parallelized, first that returns expected response is the actual cloud provider for an instance.

pkg/cloudmeta/metadata.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (cloud *CloudMeta) runWithTimeout(ctx context.Context, provider CloudMetadataProvider) (InstanceMetadata, error) {
+	ctx, cancel := context.WithTimeout(ctx, cloud.providerTimeout)
+	defer cancel()
```
karol-kokoszka, line 75: cancel function is not needed here, you can skip it.
Context will be cancelled by the timeout.

pkg/cloudmeta/metadata_test.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		if meta.InstanceType != "x-test-2" {
+			t.Fatalf("meta.InstanceType should be 'x-test-2', got %v", meta.InstanceType)
+		}
+
+		if meta.CloudProvider != "test_provider_2" {
+			t.Fatalf("meta.CloudProvider should be 'test_provider_2', got %v", meta.CloudProvider)
+		}
+	})
```
karol-kokoszka, line 92: There is quite much of repetitions in this test code.
I suggest to follow Table-Driven Tests, where you just define what's given and what's expected with the struct.
Example:
```
    tests := []struct {
        name           string
        providers      []CloudMetadataProvider
        expectedError  error
        expectedMeta   InstanceMetadata
    }{
        {
            name:           "no active providers",
            providers:      nil,
            expectedError:  ErrNoProviders,
            expectedMeta:   InstanceMetadata{},
        },
            for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
            cloudmeta := &CloudMeta{providers: tt.providers}
            meta, err := cloudmeta.GetInstanceMetadata(context.Background())

            if !errors.Is(err, tt.expectedError) {
                t.Fatalf("expected error %v, got %v", tt.expectedError, err)
            }

            if meta != tt.expectedMeta {
                t.Fatalf("expected metadata %+v, got %+v", tt.expectedMeta, meta)
            }
        })
    }
    ```

karol-kokoszka:
Thanks @VAveryanov8 , just a few comments.

pkg/cloudmeta/aws.go
karol-kokoszka: I would suggest to allow to override the whole aws config instead of just endpoint.

karol-kokoszka:
> > Also in the context of this [#4159 (review)](https://github.com/scylladb/scylla-manager/pull/4159#pullrequestreview-2498746302), does AWS SDK handle re-tries by default?
> 
> good question! yes, by default ec2metadata client will have 2 retires - https://github.com/scylladb/scylla-manager/blob/master/vendor/github.com/aws/aws-sdk-go/aws/ec2metadata/service.go#L94

I think default way of handling retries is fine here.

pkg/cloudmeta/aws.go
Michal-Leszczynski: This is expected that the metadata might not be available, as we don't know which provider is used.
Adding stack to this kind of error will probably just clog the logs. Why do you think that the stack is useful here?

karol-kokoszka: That's valid statement. We are probing all providers. OK, not needed here.

Michal-Leszczynski: Or actually, we know it form the `--location` provider part.
So there is no need to query all of the providers, which means that there is no need for the `CloudMeta` at all?

karol-kokoszka: But theoretically, you can have VMs deployed whenever you want and use the other location than the cloud provider you use for your VMs.
It's more solid to just probe it instead of relying on `--location`.

pkg/cloudmeta/aws.go
VAveryanov8: the only use case I had in mind is to change the endpoint, so AWSMetadata service can be tested against the mock,otherwise I wouldn't expose this method at all. But I see that exposing it makes a lot of confusion. 

Basically until we have CloudMeta service  - all other providers can be private to this package, they are not intended to be used outside of it. Maybe it would better just to make every provider (aws, gcp, azure) as a private service? What do you think?

karol-kokoszka: I agree they can be private.
The only thing that this service is expected to return is the provider and instance-type.

pkg/cloudmeta/metadata_test.go
VAveryanov8: yes, it's just my preference to avoid table driven tests as they usually lack of readability. 
I'll update this to align with the rest of the project :)

pkg/cloudmeta/metadata.go
VAveryanov8: yes, thanks, will update this.

pkg/cloudmeta/metadata.go
VAveryanov8: agree, is indeed not need here, but out linter doesn't like that idea `lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (govet)` :) I also prefer not to ignore cancel functions, because if someone will extend this function, then it easy to miss that cancel is ignored.

VAveryanov8:
@karol-kokoszka @Michal-Leszczynski Thanks for the review! I've updated this pr accordingly, so it's ready to be reviewed again :eyes:

karol-kokoszka (approved):
:+1:

karol-kokoszka:
BTW: please "Squash and merge" with the commit name "feat: adds cloudmeta package with AWS provider"

Michal-Leszczynski (approved):
LGTM!

Issue #4128, opened 2024-11-27
Possibility to query instance metadata on GCP


karol-kokoszka:
https://cloud.google.com/compute/docs/metadata/predefined-metadata-keys#instance-metadata (machine-type)

PR #4155, opened 2024-12-11
feat: adds gcp metadata provider support to cloudmeta package

VAveryanov8:
This adds gcp metadata provider support to cloudmeta package.

This changes has been manually tested on gcp node in lab env.

Fixes: #4128

P.S. retries are handled by gcp sdk, here is the note from documentation of `GetWithContext(...)`method.

```
Without an extra deadline in the context this call can take in the worst case, with internal backoff retries, up to 15 seconds (e.g. when server is responding slowly). Pass context with additional timeouts when needed.
```

---

Please make sure that:

### DIFF
--- a/pkg/cloudmeta/gcp.go
+++ b/pkg/cloudmeta/gcp.go
@@ -58,7 +58,7 @@ func parseMachineTypeResponse(resp string) (string, error) {
 		return "", errUnexpectedFormat
 	}
 
-	if parts[2] != "machineType" {
+	if parts[2] != "machineTypes" {
 		return "", errUnexpectedFormat
 	}
 
--- a/pkg/cloudmeta/gcp.go
+++ b/pkg/cloudmeta/gcp.go
@@ -10,20 +10,20 @@ import (
 	"github.com/pkg/errors"
 )
 
-// GCPMetadata is a wrapper around gcp metadata client.
-type GCPMetadata struct {
+// gcpMetadata is a wrapper around gcp metadata client.
+type gcpMetadata struct {
 	meta *metadata.Client
 }
 
-// NewGCPMetadata returns gcp metadata provider.
-func NewGCPMetadata() *GCPMetadata {
-	return &GCPMetadata{
+// newGCPMetadata returns gcp metadata provider.
+func newGCPMetadata() *gcpMetadata {
+	return &gcpMetadata{
 		meta: metadata.NewClient(nil),
 	}
 }
 
 // Metadata returns InstanceMetadata from gcp if available.
-func (gcp *GCPMetadata) Metadata(ctx context.Context) (InstanceMetadata, error) {
+func (gcp *gcpMetadata) Metadata(ctx context.Context) (InstanceMetadata, error) {
 	machineType, err := gcp.getMachineType(ctx)
 	if err != nil {
 		return InstanceMetadata{}, errors.Wrap(err, "gcp.meta.GetWithContext")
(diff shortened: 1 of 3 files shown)


VAveryanov8:
This pr depends on #4154

pkg/cloudmeta/metadata.go
```diff
@@ -19,8 +19,12 @@ type InstanceMetadata struct {
 // CloudProvider is enum of supported cloud providers.
 type CloudProvider string
 
-// CloudProviderAWS represents aws provider.
-var CloudProviderAWS CloudProvider = "aws"
+var (
```
Michal-Leszczynski, line 22: Let's make it const!

pkg/cloudmeta/gcp.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "gcp.meta.GetWithContext")
+	}
+
+	parts := strings.Split(machineType, "/")
+	if len(parts) < 2 {
+		return "", errors.Errorf("unexpected machine-type format: %s", machineType)
+	}
+
+	return parts[len(parts)-1], nil
```
Michal-Leszczynski, line 49: Could you paste a link to the docs with this format?
Why do you assert only `len(parts) < 2` instead checking for equality. Can the number of slashes be different?

VAveryanov8: yes, you can find `machine-type` format description on this page - https://cloud.google.com/compute/docs/metadata/predefined-metadata-keys . 

I just want to extract the latest part from `/`-separated string, so firstly I make sure this string is indeed separated with `/` and secondly the latest part is extracted. This way even if number of parts will be changed for some reason this code most likely will work correctly.

pkg/cloudmeta/metadata.go
VAveryanov8: sure, thanks!

pkg/cloudmeta/gcp.go
VAveryanov8: yes, you convinced me :) I'm going to update this!

VAveryanov8: updated!

pkg/cloudmeta/gcp.go
```diff
@@ -35,16 +35,32 @@ func (gcp *gcpMetadata) Metadata(ctx context.Context) (InstanceMetadata, error)
...
+
+// The machine type for this VM. This value has the following format: projects/PROJECT_NUM/machineTypes/MACHINE_TYPE.
+// See https://cloud.google.com/compute/docs/metadata/predefined-metadata-keys#instance-metadata.
+func parseMachineTypeResponse(resp string) (string, error) {
+	errUnexpectedFormat := errors.Errorf("unexpected machineType response format: %s", resp)
+
+	parts := strings.SplitN(resp, "/", 4)
+	if len(parts) != 4 {
+		return "", errUnexpectedFormat
+	}
```
Michal-Leszczynski, line 59: Why would you use `SplitN` instead of `Split`?

In a really really rare case of changed response syntax, `SplitN` only validates that the number of slashes is not smaller than 4, but it won't detect if the number of slashes is greater than 4.

And this is kind of picky, but from my POV using `SplitN` instead of regular `Split` should be done when there is some benefit behind it (as from my POV it's more "complicated" function variant), but it is arguably worse here.

pkg/cloudmeta/gcp.go
```diff
@@ -35,16 +35,32 @@ func (gcp *gcpMetadata) Metadata(ctx context.Context) (InstanceMetadata, error)
...
+
+	if parts[2] != "machineType" {
+		return "", errUnexpectedFormat
 	}
```
Michal-Leszczynski, line 63: > // The machine type for this VM. This value has the following format: projects/PROJECT_NUM/machineTypes/MACHINE_TYPE.

Wouldn't it mean that we need to check for `machineTypes` instead for `machineType`?

pkg/cloudmeta/gcp.go
VAveryanov8: agree, Split would be better in that case! updated!

pkg/cloudmeta/gcp.go
VAveryanov8: :facepalm: thanks for catching this! updated!

Michal-Leszczynski (approved):
LGTM!

karol-kokoszka (approved):

Issue #3892, opened 2024-06-20
Use cached node info from configcache svc in other packages


Michal-Leszczynski:
Many packages (backup/repair/restore) needs to query node info.
Right now they are simply using the scylla client to do so, but since `configcache` service already stores node info (and more) and updates it periodically, other packages should take node info from `configcache` in order to avoid unnecessary API calls and simplify the code.

Michal-Leszczynski:
This could improve session setup as described in https://github.com/scylladb/scylla-manager/pull/3888#discussion_r1647178187.

PR #4166, opened 2024-12-13
Let other services use config cache

Michal-Leszczynski:
This is the first general step toward fixing #3892.
It allows for accessing config cache service from other services.
Actually using config cache service will be implemented in separate PRs.

Ref #3892

### DIFF
--- a/pkg/cmd/scylla-manager/server.go
+++ b/pkg/cmd/scylla-manager/server.go
@@ -102,6 +102,7 @@ func (s *server) makeServices(ctx context.Context) error {
 		metrics.NewRepairMetrics().MustRegister(),
 		s.clusterSvc.Client,
 		s.clusterSvc.GetSession,
+		s.configCacheSvc,
 		s.logger.Named("repair"),
 	)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


VAveryanov8 (approved):
LGTM :+1:

karol-kokoszka (approved):

Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 sorry guys, I added one more commit to the PR.
When trying to use the changes on other branch, it turned out that I also need to add a cluster to the cluster svc passed to config cache svc. Otherwise, config cache svc would never query this cluster config.

karol-kokoszka:
@Michal-Leszczynski I understand this PR just injects the service to other services that are going to use it, right ? 
The approach could be to do it in the PR that actually starts consuming this config-cache in other services.

karol-kokoszka:
But feel free to merge.

Michal-Leszczynski:
> The approach could be to do it in the PR that actually starts consuming this config-cache in other services.

Yea, but I already implemented a helper method for creating config cache svc in tests in this PR and it wasn't wasn't really working correctly (because of describes problem), so I wanted to keep it as a part of this PR.

Issue #4161, opened 2024-12-12
Extend --location to accept raw endpoint instead of just the provider


Michal-Leszczynski:
Right now the `--location` flag allows for specifying just the provider name, while the actual endpoint is always resolved on agent side:
```
  -L, --location [<dc>:]<provider>:<bucket>   A list of backup locations separated by a comma, specifies where to place the backup, the format is [<dc>:]<provider>:<bucket>.
                                              The '<dc>' parameter is optional it allows to specify location for a datacenter in a multi-dc setting, it must match Scylla nodes datacenter.
                                              The supported storage '<provider>'s are 'azure', 'gcs', 's3'.
                                              The 'bucket' parameter is a bucket name, it must be an alphanumeric string and **may contain a dash and or a dot, but other characters are forbidden**.
```

The resolving might consist of just checking hard-coded endpoint value in the `scylla-manager-agent.yaml` file (just for `s3`), or relying on rclone/SDKs to find the default endpoint for the machine (for all providers). In some cases this might not be flexible enough.

The Scylla object storage config identifies object storage by the endpoint. This means that we need to resolve provider into the endpoint when making a backup/restore with SM while using Scylla backup/restore API.
We can implement resolver simulating rclone/SDKs logic, but we shouldn't make it the only option, as it might not cover all the use cases.

Michal-Leszczynski:
Not needed anymore as we decided to use a [simple mechanism](https://github.com/scylladb/scylla-manager/pull/4381/commits/0b30a79138b482fb5b840a91c78f9856e1755ead) which matches `scylla-manager-agent.yaml` configuration for given provider against all `object_storage_endpoints` in `scylla.yaml`.

Issue #4163, opened 2024-12-13
Add a switch to choose between Scylla and Rclone API for backup/restore


Michal-Leszczynski:
Should we allow user to decide whether backup/restore tasks should use Rclone or Scylla API?

Having a way of falling back to Rclone (without rolling back to previous SM release, which is not always possible) when some issues regarding Scylla API or its usage by SM are encountered makes sense.
It would also allow for easier testing of cross Rclone/Scylla backup/restore.

So I would say that we should add a switch to choose between those APIs, and that the Scylla API should be the default choice (when possible).

Such switch could be implemented as a new restore/backup flags or as a config field in `scylla-manager.yaml` file.

I would say that it's easier for the user and the testing if we decide to go with the new flags.
The only drawback is that we are adding more and more flags to those commands, which might make them more difficult to configure for the users. On the other hand, in an ideal scenario, we don't expect users to use this flag at all. But if that's a case, maybe it's enough to just put it in the `scylla-manager.yaml` file, where it can still be configured on the really rare occasions, when it's needed?

What do you think?
cc: @karol-kokoszka @VAveryanov8 @regevran

regevran:
I don't think we need to keep both options - as the user has nothing to select from. It is implementation detail that, given the option to select, just confuses.

Having said that, if the same SM version needs to support both rclone and scylla - _one_ solution could be to make it configurable. A better solution (IMO) would be to query scylla for its ability to backup and restore - and when it can't - fallback to rclone. No user setup at all.

regevran:
> if there is a bug on SM or Scylla side

I am not familiar with defensive programming - finding solutions to problems that were not happened yet.
I think it would add more confusion and by itself may be a source of troubles.
E.g. someone someday changes the yaml and then experiences degradation in performance and we need to look for the reason. 
But again - if it is more convenient for SM to support older Scylla versions then, taking into consideration the risks - I am not objecting the yaml solution.

tzach:
To validate:
- Scylla API: ScyllaDB core interacts directly with S3
- Rclone API: Manager agents use Rclone to interact directly with S3

My 2c:
- User should not care if we are using rclone
- Best will be to convert to one solution ASAP. Supporting two solutions is a pain.
- As long as the core backup does not support all cloud, we must keep agent backup as an option.

karol-kokoszka:
We still need to support older versions of Scylla (correct me if I'm wrong), what means that the RClone support for backups and restores stays in Scylla Manager code.

[...]

PR #4169, opened 2024-12-16
Use Scylla API for backup

### LINKED ISSUE #4138: Handle SM backup task pause/resume with Scylla API
We should support the same SM task pause/resume functionality with native Scylla backup task.

### LINKED ISSUE #4141: Monitor Scylla backup API progress
Ref: https://github.com/scylladb/scylladb/issues/20653

### LINKED ISSUE #4143: Replace `/rclone/sync/movedir` with `/storage_service/backup` to use scylla streaming
Current effort on core side is expected to bring the feature where the data is send to AWS S3 (to be validated what about GCS and Azure) directly by Scylla server.
Scylla Manager is currently using RClone under the hood to send the data to S3.

This issue expects Scylla Manager to start using `/storage_service/backup` endpoint of Scylla server instead of the RClone one. https://docs.google.com/document/d/1d-nJnn5n2gqzRXvP5xpOis0fWxIzYkf-u9X1xrIAG_c/edit?tab=t.0#bookmark=id.9g3y3uw716bc

PR cannot be merged to master, we just want to keep it on feature branch until there is a green light that all corner cases are covered and we finally can get rid of RClone.

Michal-Leszczynski:
This PR starts using Scylla backup API in SM backup task!
It is mostly complete and can be tested, but there are 3 issues that were discovered during development:
- https://github.com/scylladb/scylla-manager/issues/4161 - right now SM tries its best to resolve the endpoint from `scylla-manager-agent.yaml` and `object-storage.yaml`. It should be enough for default setups and tests, but we will be on the safe side when we fix this issue. Take a look at bd0cf1ad67b1203451d318197d223d07c6c4bc16 for more info.
- https://github.com/scylladb/scylla-manager/issues/4181 - this can create issues in other places in the code and forced a workaround in the backup code.
- https://github.com/scylladb/scylla-manager/issues/4182 - this makes it really annoying to test Rclone/Scylla API against mixed integer/UUID sstables.

In terms of the general overview of this PR - the main objective was to fix replace the `/agent/rclone/sync/movedir` Rclone API with the `/storage_service/backup` Scylla API - nothing more.
Scylla API can be used when:
- node exposes Scylla backup API
- s3 is the used provider
- backup won't create versioned files

Checking whether Scylla API can be used is done separately per node/snapshot_dir.

Luckily, things like pause/resume/progress does not seem like they need additional work in the scope of this issue.

Also, for now Scylla versions which are supposed to support Scylla backup/restore API are:
- `master`
- `6.3`
- `2024.3`

Fixes #4143
Fixes #4138
Fixes #4141


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 so the idea is that the `ml/scylla-api` will be the branch for the scylla api milestone.
Please take a look!

VAveryanov8:
Nice! Looks good to me! 
Just a few questions/nits

karol-kokoszka (commented):
I will continue with the review, but leaving one comment now. Need to understand why SM must pass s3 endpoint. 
I feel this is the security threat (but maybe I'm wrong, I'm not security expert - convince me that it is not the threat).

pkg/cmd/agent/router.go
```diff
@@ -53,17 +58,166 @@ func promProxy(c agent.Config) http.Handler {
...
+// In such case, SM sends provider name as the "endpoint" query param,
+// which is resolved by agent to proper host name when forwarding request to Scylla.
+// Different "endpoint" query params are not resolved.
+//
+// Note that resolving "endpoint" query param in the proxy is just for the UX,
+// so it might not work correctly in all the cases.
+// In order to ensure correctness, "endpoint" should be specified directly by SM user
+// so that no resolving is needed.
+func objectStorageEndpointDirector(cfg agent.Config, logger log.Logger) func(r *http.Request) {
+	regex := regexp.MustCompile(`^/storage_service/backup$`)
```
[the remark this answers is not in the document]
Michal-Leszczynski: There is none:)
My though process was that in the next PR adding support for restore I would need to also check for the restore endpoint which is `/storage_service/restore` and I just thought that it would be nice to reuse their prefix.
But if you prefer the simpler approach, I can change that (although in general this best effort approach wouldn't probably be released - it's just for the testing now).

pkg/cmd/agent/router.go
```diff
@@ -53,17 +58,166 @@ func promProxy(c agent.Config) http.Handler {
...
+		if resolvedEndpoint != "" {
+			return resolvedEndpoint, nil
+		}
+
+		resolver := endpoints.DefaultResolver()
+		re, err := resolver.EndpointFor(s3Provider, region)
+		if err != nil {
+			return "", errors.Wrap(err, "resolve S3 endpoint for region "+region)
+		}
+
```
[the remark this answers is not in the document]
Michal-Leszczynski: Yeah, that would make sense:)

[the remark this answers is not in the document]
Michal-Leszczynski: done!

pkg/cmd/agent/router.go
```diff
@@ -53,17 +58,167 @@ func promProxy(c agent.Config) http.Handler {
...
+		resolvedHost, err := endpointToHostName(resolvedEndpoint)
+		if err != nil {
+			logger.Error(r.Context(), "Failed to convert endpoint to host name",
+				"endpoint", resolvedEndpoint,
+				"err", err)
+			return
+		}
+
```
karol-kokoszka, line 149: What happens when there is an error raised in the Directory that is expected to validate&modify the incoming request ? 
I understand that the error is just logged, but proceeds with the call to the scylla server with the current implementation.
This behavior means that the inproper request is proxied still to the scylla server.

It's much more efficient to stop processing the invalid request straight in the proxy instead of forwarding it.
To do that, you can introduce the middleware that is reponsible for validation + 
```
http.Error(w, "Missing required query parameter: endpoint", http.StatusBadRequest)
return
```
The middleware should evaluate the correct endpoint and save it to conext.
Then, the director just checks the context to see if there is something to change.

Director, can skip the validation 
This comment is only valid if you proceed with resolving endpoint in agent. The topic to discuss is that is can be safer/better to have the identifiers of endpoints in object storage and use it in sm configuration.

pkg/scyllaclient/client_agent.go
```diff
@@ -254,6 +254,29 @@ func (ni *NodeInfo) SupportsSafeDescribeSchemaWithInternals() (SafeDescribeMetho
...
+		return true, nil
+	}
+	// Check ENT
+	supports, err = scyllaversion.CheckConstraint(ni.ScyllaVersion, ">= 2024.3")
+	if err != nil {
+		return false, errors.Errorf("Unsupported Scylla version: %s", ni.ScyllaVersion)
+	}
+	return supports, nil
+}
+
```
karol-kokoszka, line 279: So, it's basing on the release version. 
It's confirmed that Scylla exposes wtih appi beginning with 6.3 and 2024.3 ? 
Is there any other possibility of checking the availablity of this endpoint ? Maybe simple HEAD request to the endpoint ?

pkg/service/backup/service.go
```diff
@@ -677,6 +677,11 @@ func (s *Service) Backup(ctx context.Context, clusterID, taskID, runID uuid.UUID
 		return errors.Wrap(err, "invalid cluster")
 	}
 
+	nodeConfig, err := s.configCache.ReadAll(clusterID)
+	if err != nil {
+		return errors.Wrap(err, "read all nodes config")
+	}
+
```
karol-kokoszka, line 684: Maybe it's safer and more accurate to force config cache to update cluster configuration first ? And then call to Read All.

pkg/cmd/agent/router.go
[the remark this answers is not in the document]
VAveryanov8: thanks for clarifying, we can leave it for testing :)

pkg/scyllaclient/client_agent.go
Michal-Leszczynski: It's not confirmed, we would need  to come back to this part after an actual Scylla release supports native backup/restore API.

pkg/service/backup/service_backup_integration_test.go
```diff
@@ -2625,3 +2644,74 @@ func TestBackupSkipSchema(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if !encounteredEnsured.Load() {
+		t.Fatalf("Expected SM to use %q API", ensuredPath)
+	}
+	if encounteredBlocked.Load() {
+		t.Fatalf("Expected SM not to use %q API", blockedPath)
+	}
+}
```
karol-kokoszka, line 2717: `latest-enterprise` has the Scylla API changes included already ? 
I'm asking because I wonder if the GH actions are already testing the scenario with Scylla API for backup.

pkg/service/backup/worker_deduplicate.go
```diff
@@ -134,48 +135,55 @@ func (w *worker) deduplicateUUIDSStables(remoteSSTables, localSSTables *sstableB
...
+		if !ok {
+			continue
+		}
+		// At this point analyzed SSTable ID is present in both local and remote dirs.
+		// Not being able to deduplicate it results in setting 'willCreateVersioned' to true.
 		crc32Idx := slices.IndexFunc(localBundle, func(fi fileInfo) bool {
 			return strings.HasSuffix(fi.Name, "Digest.crc32")
 		})
 		if crc32Idx == -1 {
+			willCreateVersioned = true
```
karol-kokoszka, line 159: Why it will create versioned if there is no crc32 file ? We don't know it. It will create versioned only if the checksum generated by Rclone are different.
But there is no option to validate it at this stage, so yeah it's possible better to assume it will.

pkg/service/backup/worker_scylla_upload.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
+	. "github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
+	"github.com/scylladb/scylla-manager/v3/swagger/gen/scylla/v1/models"
+)
+
+// Decides whether we should use Scylla backup API for uploading the files.
+func (w *worker) useScyllaBackupAPI(ctx context.Context, d snapshotDir, hi hostInfo) (bool, error) {
+	// Scylla backup API does not handle creation of versioned files.
+	if d.willCreateVersioned {
+		return false, nil
```
karol-kokoszka, line 20: I miss more debug information like why ScyllaBackupAPI is not going to be used.
My proposal is to either replace (bool, error) returns with (error) return, where error!=nil means not supported. Error message can be logged (info) as a reasoning then.
I see you log which API is going to be used, so RClone api log info can be extended with reasoning,.

pkg/service/backup/worker_upload.go
```diff
@@ -71,6 +71,14 @@ func (w *worker) uploadHost(ctx context.Context, h hostInfo) error {
 			err = errors.Wrap(w.deleteTableSnapshot(ctx, h, d), "delete table snapshot")
 		}()
 
+		if ok, err := w.useScyllaBackupAPI(ctx, d, h); err != nil {
+			return errors.Wrap(err, "check Scylla backup API support")
```
karol-kokoszka, line 75: Why return on error, and not failback to RClone API ?

schema/v3.5.0.cql
```diff
@@ -0,0 +1 @@
+ALTER TABLE backup_run_progress ADD scylla_task_id text;
```
karol-kokoszka, line 1: nit: Why new parameter is needed ? Cannot agent_job_id be used ? (it's not a complain, just the curiosity)

v3/swagger/agent.json
```diff
@@ -966,6 +966,34 @@
...
+          "default": {
+            "description": "Server error",
+            "schema": {
+              "$ref": "#/definitions/ErrorResponse"
+            },
+            "headers": {}
+          }
+        },
+        "security": []
+      }
```
karol-kokoszka, line 996: You have some commits from @VAveryanov8 PR here ?

karol-kokoszka (changes requested):
@Michal-Leszczynski Thanks for the PR! Looks nice. There are few comments to discuss.
The most important is to figure out what to do with these endpoints configured for object storage.
I'm setting "Request changes" to block it until it's confirmed if core can use IDs for example.

pkg/scyllaclient/client_agent.go
karol-kokoszka: Can we assume that if endpoint is available then it means scylla supports it ? If so, then I suggest to probe api with HEAD request.

pkg/cmd/agent/router.go
```diff
@@ -53,17 +58,166 @@ func promProxy(c agent.Config) http.Handler {
...
+				return "", errors.New("account is not set in Azure config")
+			}
+			endpoint := cfg.Azure.Endpoint
+			if endpoint == "" {
+				endpoint = defaultAzureEndpoint
+			}
+			resolvedEndpoint = cfg.Azure.Account + "." + endpoint
+		default:
+			// Endpoint has already been resolved on SM side
+			resolvedEndpoint = provider
```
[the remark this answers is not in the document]
Michal-Leszczynski: I think that we covered those points during the meeting, right?

pkg/scyllaclient/client_agent.go
Michal-Leszczynski: In general I like the idea of not relying on version checks, but proposed approach requires Scylla to handle HEAD requests and I don't think that's the case. At least I can't see them in the swagger definitions or swagger UI.

pkg/cmd/agent/router.go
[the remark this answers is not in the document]
karol-kokoszka: Right. These are comments from before the sync. Waiting for the output from the meeting then.

pkg/service/backup/worker_deduplicate.go
Michal-Leszczynski: That's exactly my though process:)

pkg/service/backup/worker_upload.go
Michal-Leszczynski: We could do it, but do we want to?
Because of https://github.com/scylladb/scylla-manager/issues/4163 it will be possible to control which API should be primarily used for the backup process.
The fallback without any error analysis can prolong the agony of a problematic backup (which might fail with Rclone API as well).
There is also a question of whether interrupting backup with Scylla and resuming it with Rclone is safe (although I don't know why wouldn't it be).

schema/v3.5.0.cql
Michal-Leszczynski: They have different types and it also allow for distinguishing which API was used.

pkg/service/restore/restore_integration_test.go
```diff
@@ -611,9 +611,6 @@ func TestRestoreTablesPreparationIntegration(t *testing.T) {
 		"rate_limit": []string{"88"},
 	})
 
-	Print("Validate state after backup")
-	validateState(h.srcCluster, "repair", true, 3, 88, pinnedCPU)
-
```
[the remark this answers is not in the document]
Michal-Leszczynski: When using Scylla backup API we don't need to alter transfers, rate limit, cpu pinning etc, so I deleted this check out of laziness. But I can bring it back for testing scenarios using Rclone API.

pkg/scyllaclient/client_agent.go
karol-kokoszka: OK, they don't support HEAD requests, at least this is what I see trying to `curl` for it from localhost.
GET responds OK, HEAD gives 404 (not found).

Then maybe, there is some configuration parameter available to GET which can answer whether the feature is available ?

pkg/service/backup/worker_upload.go
karol-kokoszka: Thanks for the context.
Let's not failback.

pkg/service/restore/restore_integration_test.go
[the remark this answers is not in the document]
karol-kokoszka: yes, please

pkg/scyllaclient/client_agent.go
Michal-Leszczynski: Not sure about this suggestion. Wouldn't it be difficult to distinguish potential errors coming from API not being exposed and connectivity issues?
Also, the initial idea was to check Scylla swagger definitions, but it's currently not possible due to https://github.com/scylladb/scylladb/issues/16424.

karol-kokoszka: If you have connectivity issue, then you got network level error. You won't get response with HTTP code.

Michal-Leszczynski: > Then maybe, there is some configuration parameter available to GET which can answer whether the feature is available ?

Could you elaborate?

karol-kokoszka: If core uses some configuration parameter related to the backup API in scylla that is exposed together with this API, there could be simple GET performed on this config param.
404 code means that the API is not available.

But, if the backup API means that there are only POST methods exposed that are changing the state then I think you need to stay with the current approach where you basically checks the versions number.

karol-kokoszka (approved):

pkg/service/backup/worker_scylla_upload.go
Michal-Leszczynski: I don't like mixing actual errors with checks, so I just added logging on failed checks.

Issue #4173, opened 2024-12-17
Flakiness in unit-test related to metadata


karol-kokoszka:
https://github.com/scylladb/scylla-manager/actions/runs/12372217257/job/34529936552

@VAveryanov8 please check

Issue #4172, opened 2024-12-17
TestPurgeIntegration is flaky


Michal-Leszczynski:
It looks like sometimes SM does not purge all manifests in TestPurgeIntegration:
```
2024-12-16T12:06:48.5510081Z --- Then: there should be 3 + 4 manifests
2024-12-16T12:06:48.5629775Z     service_backup_integration_test.go:1564: Expected 7 manifests (1 per each node) plus 4 generated, got 8 backup/meta/cluster/b9a74faf-dfef-498b-853f-943b81ca38cb/dc/dc1/node/03445f8d-a8f6-4397-861b-23ff71c1fb28/task_070e03d3-b12d-46c6-bb8c-a1f62183fbd2_tag_sm_20241215120647UTC_manifest.json.gz
2024-12-16T12:06:48.5633543Z         backup/meta/cluster/b9a74faf-dfef-498b-853f-943b81ca38cb/dc/dc1/node/03445f8d-a8f6-4397-861b-23ff71c1fb28/task_1f795853-fb1b-4c7a-8e80-928631431176_tag_sm_20241219120647UTC_manifest.json.gz
2024-12-16T12:06:48.5636397Z         backup/meta/cluster/b9a74faf-dfef-498b-853f-943b81ca38cb/dc/dc1/node/03445f8d-a8f6-4397-861b-23ff71c1fb28/task_2f2aaeb1-0922-4378-a217-8c7291df2014_tag_sm_20241216120647UTC_manifest.json.gz
2024-12-16T12:06:48.5638228Z         backup/meta/cluster/b9a74faf-dfef-498b-853f-943b81ca38cb/dc/dc1/node/03445f8d-a8f6-4397-861b-23ff71c1fb28/task_421ce0b2-3b0c-4719-8bb6-3b5259029f5e_tag_sm_20241215100647UTC_manifest.json.gz
2024-12-16T12:06:48.5640671Z         backup/meta/cluster/b9a74faf-dfef-498b-853f-943b81ca38cb/dc/dc1/node/03445f8d-a8f6-4397-861b-23ff71c1fb28/task_421ce0b2-3b0c-4719-8bb6-3b5259029f5e_tag_sm_20241216100647UTC_manifest.json.gz
[...]

PR #4175, opened 2024-12-17
fix(tests): increases delays in `pkg/cloudmeta` unit tests.

VAveryanov8:
This increases delay between 'fast' and 'slow' providers in metadata unit tests to reduce chance of failure.

Fixes: #4173

---

Please make sure that:


Michal-Leszczynski:
@VAveryanov8 please report the flakiness under https://github.com/scylladb/scylla-manager/issues/4172.

Michal-Leszczynski (approved):

Issue #4130, opened 2024-11-27
Extend the backup manifest


karol-kokoszka:
Include intro backup manifest rack and instance details.

karol-kokoszka:
> Let's also take https://github.com/scylladb/scylla-manager/issues/3916 into consideration.

We must add to the manifest file things like `cluster-id`, `node-id` (which is the `host-id`) and the `dc-name`. Basically everything which is used to identify the particular manifest files with the path.

Path is defined as follow
```
/backup/sst/cluster/<cluster ID>/dc/<data center name>/node/<node ID>
```

PR #4177, opened 2024-12-17
feat(backup): extends backup manifest with info needed for 1-to-1 restore. 

VAveryanov8:
This adds following data to the backup manifest:
##### General:
  **cluster_id**: uuid of the cluster
  **dc**: data center name
  **rack**: rack from the scylla configuration
  **node_id**:  id of the scylla node (equals to host id)
  **task_id**: uuid of the backup task
  **snapshot_tag**: snapshot tag 
  **shard_count**: number of shards in scylladb
  **cpu_count**: number of cpu available on scylla node
  **storage_size**: total size of the disk in bytes
##### Instance Details:
  **cloud_provider**: aws|gcp|azure or empty in case of on-premise
  **instance_type**: instance type, e.g. t2.nano or empty when on-premise

This also includes bug fix in cloudmeta.GetInstanceMetadata(ctx) - adds check for ctx cancellation. 
This also includes fixes in unit tests related to NodeInfo.

Fixes: https://github.com/scylladb/scylla-manager/issues/4130

---

Please make sure that:


VAveryanov8:
If we intend to populating [ManifestInfo](https://github.com/scylladb/scylla-manager/blob/acf312c7b813d9eec3fecb401e20722e470c8253/pkg/service/backup/backupspec/manifest.go#L23) from file content instead of file path and file name, then it worth to include `snapshot_id` and `task_id` into the manifest. 

However moving to populating ManifestInfo from file content looks like a relatively significant change considering that we need to preserve backward compatibility with  "older" manifests.

karol-kokoszka:
> If we intend to populating [ManifestInfo](https://github.com/scylladb/scylla-manager/blob/acf312c7b813d9eec3fecb401e20722e470c8253/pkg/service/backup/backupspec/manifest.go#L23) from file content instead of file path and file name, then it worth to include snapshot_id and task_id into the manifest.

> However moving to populating ManifestInfo from file content looks like a relatively significant change considering that we need to preserve backward compatibility with "older" manifests.

I'm not sure if I understand what do you mean ? We just want to add additional information to the manifest file without removing or changing anything.

You mean that if we want to include snapshot_id and task_id then it's a significant change ?

VAveryanov8:
> I'm not sure if I understand what do you mean ? We just want to add additional information to the manifest file without removing or changing anything.
> 
> You mean that if we want to include snapshot_id and task_id then it's a significant change ?

We briefly mentioned on a call, that we may want to simplify how ManifestInfo is populated if all needed info will be contained in a manifest file.
So I'm just pointing out if we want to do that - if will be a relatively significant change + snapshot_id and task_id are currently missing.

VAveryanov8:
I've updated pr - added snapshot_id and task_id, it's ready for review :eye:

VAveryanov8:
@Michal-Leszczynski @karol-kokoszka this pr is ready for review :eye:

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/scylladb/scylla-manager/v3
 go 1.23.2
 
 require (
+	cloud.google.com/go/compute/metadata v0.3.0
```
Michal-Leszczynski, line 6: Going with the `squash and merge` and mixing commits that update vendor and implement features is messy. In order to make it cleaner, we can either:
- make a separate PR for updating vendor
- don't use `squash and merge` - the owner of the PR would need to manually squash commits with some reasonable logic (e.g. separate vendor changes from feature implementations) and then use the `rebase and merge` option

pkg/cmd/agent/nodeinfo_linux.go
```diff
@@ -17,9 +18,16 @@ func (h *nodeInfoHandler) sysInfo(info *scyllaclient.NodeInfo) error {
 		return err
 	}
 
+	var statfs unix.Statfs_t
+	if err := unix.Statfs(info.DataDirectory, &statfs); err != nil {
+		return errors.Wrap(err, "statfs")
+	}
+	total := statfs.Blocks * uint64(statfs.Bsize)
```
Michal-Leszczynski, line 25: Not an expert on this matter, but from my POV using cloud provider sdk (e.g. [ec2 sdk](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#Client.DescribeVolumes)) should be safer and more convenient for the siren as well. Or perhaps do you think that it's completely unnecessary and would create the need to implement it separately for each provider in the `cloudmeta` pkg?

pkg/service/backup/worker_manifest.go
```diff
@@ -88,10 +92,60 @@ func (w *worker) createTemporaryManifest(h hostInfo, tokens []int64) ManifestInf
...
+	if err != nil {
+		// Metadata may not be available for several reasons:
+		// 1. running on-premise 2. disabled 3. smth went wrong with metadata server.
+		// As we cannot distiguish between this cases we can only log err and continue with backup.
+		w.Logger.Error(ctx, "Get instance metadata", "err", err)
+	}
+	result.CloudProvider = string(instanceMeta.CloudProvider)
+	result.InstanceType = instanceMeta.InstanceType
+
+	return result, nil
```
Michal-Leszczynski: Unfortunately, all of this code is executed on the SM side.
I guess that querying instance metadata should be done on the agent side?

Michal-Leszczynski (changes requested):

pkg/service/backup/worker_manifest.go
VAveryanov8: oh, I missed that, you're right!

pkg/cmd/agent/nodeinfo_linux.go
VAveryanov8: Do you think NodeInfo is a good place to [extend](https://github.com/scylladb/scylla-manager/blob/164ff2ce065cd106253fed40d2f6ffe883e73cc6/v3/swagger/gen/agent/models/node_info.go#L18) with `InstanceType`, `CloudProvider` information? Or it's better to have a separate call for them?

pkg/service/backup/worker_manifest.go
VAveryanov8: Do you think NodeInfo is a good place to [extend](https://github.com/scylladb/scylla-manager/blob/164ff2ce065cd106253fed40d2f6ffe883e73cc6/v3/swagger/gen/agent/models/node_info.go#L18) with InstanceType, CloudProvider information? Or it's better to have a separate call for them?

pkg/cmd/agent/nodeinfo_linux.go
VAveryanov8: good idea actually, but I'm worried that these calls will require additional permissions (e.g. DescribeVolumes) that service_account (or iam role) that attached to instance should have. And in case of on-premise we need to query file system anyway

go.mod
VAveryanov8: Yes, I can squash changes manually before merging and then merge without squashing.

pkg/cmd/agent/nodeinfo_linux.go
Michal-Leszczynski: We shouldn't worry with on-premise too much, as this effort is targeted for the cloud and other things won't work for on-premise as well (e.g. instance type). Having said that, I'm fine with the current approach, just wanted to note that there is another way of doing it, but if it doesn't seem better to you, there is no need for implementing it.

pkg/service/backup/worker_manifest.go
Michal-Leszczynski: I wouldn't mix them if it meant that we always try to query multiple providers when fetching NodeInfo, as it's used in multiple places in the code.
But I guess that it's safe to assume that the instance type won't change at runtime, so we can cache this value in agent memory and query it only once per agent re-start. The problem could be to distinguish cases of getting timeout on querying instance type vs querying it on on-prem.
Another approach could be to extend NodeInfo API call to include optional query param specifying that the instance details should also be included into the node info, but that's not that different from writing a separate endpoint for them.

VAveryanov8: I think I'll go with separate call just for the metadata

VAveryanov8:
TODO before merge
- [ ] Once #4186 is merged, update vendor to use `v3/swagger` from master branch. Currently this pr uses `v3/swagger` from the branch.
- [ ] Manually squash changes

pkg/scyllaclient/client_agent.go
```diff
@@ -262,3 +262,25 @@ func (c *Client) FreeOSMemory(ctx context.Context, host string) error {
...
+
+	meta, err := c.agentOps.Metadata(&p)
+	if err != nil {
+		return InstanceMetadata{}, errors.Wrap(err, "cloud metadata")
+	}
+
+	payload := meta.GetPayload()
+	if payload == nil {
+		return InstanceMetadata{}, nil
+	}
```
Michal-Leszczynski, line 280: I would say that the error checking should be enough (it should be reasonable to assume that agent does not hide errors), but if we are also validating that `payload != nil`, then we should return an error if the validation fails.

pkg/service/backup/backupspec/manifest.go
```diff
@@ -115,12 +115,29 @@ func (m *ManifestInfo) fileNameParser(v string) error {
...
+}
+
+// InstanceDetails extends backup manifest with additional instance details.
+// Mainly needed for 1-to-1 restore.
+type InstanceDetails struct {
+	CloudProvider string `json:"cloud_provider,omitempty"`
+	InstanceType  string `json:"instance_type,omitempty"`
+
+	ShardCount  int    `json:"shard_count"`
+	StorageSize uint64 `json:"storage_size"`
```
Michal-Leszczynski, line 140: I see that it makes sense to separate the cloud metadata fields to a separate struct, but IMHO a thing like shard count should be next to things like rack/tokens (shard count information might be useful for optimizations even outside of the 1-1 restore). A similar, but weaker argument could be made for total storage size.
Could we move those two directly into the `ManifestContent`?

pkg/service/backup/backupspec/manifest.go
```diff
@@ -115,12 +115,29 @@ func (m *ManifestInfo) fileNameParser(v string) error {
...
+	NodeID          string          `json:"node_id"`
+	DC              string          `json:"dc"`
+	TaskID          uuid.UUID       `json:"task_id"`
+	SnapshotTag     string          `json:"snapshot_tag"`
+	IP              string          `json:"ip"`
+	Size            int64           `json:"size"`
+	Tokens          []int64         `json:"tokens"`
+	Schema          string          `json:"schema"`
+	Rack            string          `json:"rack"`
+	InstanceDetails InstanceDetails `json:"instance_details"`
```
Michal-Leszczynski, line 130: Could we reorder those fields?
I feel like it would make sense to group and separate fields which are also present in the file path (with a proper comment).

pkg/service/backup/worker_manifest.go
```diff
@@ -88,10 +96,48 @@ func (w *worker) createTemporaryManifest(h hostInfo, tokens []int64) ManifestInf
...
+
+	shardCound, err := w.Client.ShardCount(ctx, host.IP)
+	if err != nil {
+		return InstanceDetails{}, errors.Wrap(err, "client.ShardCount")
+	}
+	result.ShardCount = int(shardCound)
+
+	nodeInfo, err := w.Client.NodeInfo(ctx, host.IP)
+	if err != nil {
+		return InstanceDetails{}, errors.Wrap(err, "client.NodeInfo")
```
Michal-Leszczynski, line 129: Perhaps shard count will be moved out of instance details, but could you leave a TODO comment (watch out for the linter!) here saying that this is the candidate for https://github.com/scylladb/scylla-manager/issues/3892 (but only after https://github.com/scylladb/scylla-manager/issues/4181 gets fixed...)?

Michal-Leszczynski (commented):
I just started thinking about re-tries and timeouts when fetching cloud metadata.
The problem is that we have them configured on both agent client (used by SM) and object storage client (used by the agent) and it seems problematic.

My first feeling is that we should drop any re-try and timeout handling on the agent side (as it acts as a advanced proxy) and rely on the configuration used by SM. I know that you were specifically asked to add the re-tries on the agent side, but I guess that we missed this problem before.

What are your opinions on this matter?
Maybe there is some other approach that solves this issue?

The current implementation of fetching cloud metadata just once at the start-up solves this problem, but it has other drawbacks mentioned in the comment. Maybe this is the best solution, but could you elaborate on other, possible approaches to this problem?

pkg/cmd/agent/metadata.go
```diff
@@ -0,0 +1,41 @@
...
+		logger.Error(ctx, "GetInstanceMetadata", "err", err)
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		render.Respond(w, r, metadataResponse{
+			CloudProvider: string(instanceMeta.CloudProvider),
+			InstanceType:  instanceMeta.InstanceType,
+		})
+	}, nil
+}
```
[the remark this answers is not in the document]
VAveryanov8: > we try to fetch the metadata only once at the start-up and we never re-try in case of failure

We have a retries in cloudmeta package. So I think if we failed to get metadata during X retries, then we never will be able to have a success with this call. We can increase number of retries ( I think right now it's around 3 for almost all providers), in order to have bigger chance of success :) 

> we hide any errors from SM 

Unfortunately we can't distinguish error that happened because metadata is not available (disabled or on-prem) from err that happened while actually fetching metadata. Considering that we have retries in  cloudmeta package we can assume that if error is returned from GetInstanceMetadata, then it means metadata is not available for us. 

> we use the instanceMetadata value even in case of an error 

Yes, taking into account that err indicate for us that metadata is not available, it's ok to simply reply with empty `instanceMeta`

pkg/scyllaclient/model.go
```diff
@@ -245,3 +245,9 @@ func ViewBuildStatusOrder() []ViewBuildStatus {
 func (s ViewBuildStatus) Index() int {
 	return slice.Index(ViewBuildStatusOrder(), s)
 }
+
+// InstanceMetadata defines node metadata needed for 1-to-1 restore.
+type InstanceMetadata struct {
+	CloudProvider string
+	InstanceType  string
+}
```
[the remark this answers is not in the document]
VAveryanov8: IMHO swagger models are for the transport level and here we have a package that encapsulates transport level details, so it's better not to expose internal details. Let me know if you have a different opinion about this.

pkg/cmd/agent/metadata.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/go-chi/render"
+	"github.com/pkg/errors"
+	"github.com/scylladb/go-log"
+	"github.com/scylladb/scylla-manager/v3/pkg/cloudmeta"
+)
+
+type metadataResponse struct {
+	CloudProvider string `json:"cloud_provider"`
+	InstanceType  string `json:"instance_type"`
+}
```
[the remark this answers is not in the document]
VAveryanov8: thanks, updated!

pkg/scyllaclient/client_agent.go
VAveryanov8: I'm scared of nil pointer dereference panic, so I've changed this code to return error :smile:

pkg/service/backup/backupspec/manifest.go
VAveryanov8: sure, updated!

pkg/service/backup/backupspec/manifest.go
VAveryanov8: done!

pkg/service/backup/worker_manifest.go
VAveryanov8: yep, added `VA_TODO:..` comment for that

pkg/service/backup/backupspec/manifest.go
```diff
@@ -115,12 +115,30 @@ func (m *ManifestInfo) fileNameParser(v string) error {
...
-	Tokens      []int64 `json:"tokens"`
-	Schema      string  `json:"schema"`
+	Version         string          `json:"version"`
+	ClusterName     string          `json:"cluster_name"`
+	IP              string          `json:"ip"`
+	Size            int64           `json:"size"`
+	Tokens          []int64         `json:"tokens"`
+	Schema          string          `json:"schema"`
+	Rack            string          `json:"rack"`
+	ShardCount      int             `json:"shard_count"`
```
karol-kokoszka, line 125: Could you please add number of CPUs next to shards as well ? 
Number of shards shows how many CPUs are taken by Scylla.
Number of CPU shows the total number of CPUs on the machine.

karol-kokoszka (changes requested):
Thanks !
Please fix the panic.

Additionaly, please harden some of the backup integration tests so that they will check the generated manifest content as well.

Backup integration tests are in pkg/service/backup/service_backup_integration_test.go.

pkg/cloudmeta/metadata.go
```diff
@@ -98,12 +98,16 @@ func (cloud *CloudMeta) GetInstanceMetadata(ctx context.Context) (InstanceMetada
...
-			mErr = multierr.Append(mErr, res.err)
-			continue
+		select {
+		case <-ctx.Done():
+			return InstanceMetadata{}, ctx.Err()
+		case res := <-results:
+			if res.err != nil {
+				mErr = multierr.Append(mErr, res.err)
+				continue
+			}
```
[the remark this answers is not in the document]
VAveryanov8:
```
		select {
		case <-ctx.Done():
			return InstanceMetadata{}, ctx.Err()
		case res := <-results:
			if res.err != nil {
				mErr = multierr.Append(mErr, res.err)
				continue
			}
```
In the above case I think it's ok to randomly return result or err if context was cancelled in the same time as result is available. But yeah, after thinking some time about it I prefer to change this code as you suggested, to avoid any confusions and it will more correct! Thanks! 

yes, you're right part with goroutine provider looks incorrect! 

The initial idea behind this code is that goroutine with provider shouldn't send any msg to `results` channel if context is cancelled. 
That's why we need another check for `ctx.Done()` when reading from `results`, so if  context was cancelled before any provider returned the result, we will not block forever on waiting for the results. 

Also I added this redundant context cancellation check in provider goroutine, so that only first provider will send a message to the `results` channel and other provider goroutines should simply return (because context will be cancelled once first msg is received from the channel). Probably there is nothing wrong with sending a message to a buffered channel when it will be never received from it, gc should handle such cases, but I've tried to avoid it anyway. 


This part should looks like this: 
```go
		go func(provider CloudMetadataProvider) {
			meta, err := cloud.runWithTimeout(ctx, provider)

                        select {
			case <-ctx.Done():
			     return
                        default:
                        }

			results <- msg{meta: meta, err: err}:
		}(provider)
```
and the rest is correct i think. But let me know if you think it's better to simply this part and do not check for context cancellation here at all.

pkg/cloudmeta/metadata_test.go
```diff
@@ -101,6 +102,31 @@ func TestGetInstanceMetadata(t *testing.T) {
...
+
+	if meta.CloudProvider != "" {
+		t.Fatalf("meta.CloudProvider should be empty, got %s", meta.CloudProvider)
+	}
+
+	if meta.InstanceType != "" {
+		t.Fatalf("meta.InstanceType should be empty, got %s", meta.InstanceType)
+	}
+}
+
```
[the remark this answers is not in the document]
VAveryanov8: yes, the purpose of this test is exactly to check if code correctly handles context cancellations (it wasn't before the change). 

Let's assume below code works always correctly : 
```
	results := make(chan msg, len(cloud.providers))
	for _, provider := range cloud.providers {
		go func(provider CloudMetadataProvider) {
			meta, err := cloud.runWithTimeout(ctx, provider)

			select {
			case <-ctx.Done():
				return
			case results <- msg{meta: meta, err: err}:
			}
		}(provider)
	}
```
If context is cancelled before any goroutine (provider) returned the result (meta or err), then no messages will be send to the results channel and below code will stuck forever: 

```
for range len(cloud.providers) {
		res := <-results  // this channel will never return any value if context is cancelled.
		if res.err != nil {
			mErr = multierr.Append(mErr, res.err)
			continue
		}
		return res.meta, nil
	}
```

So the idea of the test was to cancel context before any provider will return result and the simplest  scenario of that is to run GetInstanceMetadata function with already cancelled context :smile: But yeah, I can extend the test with cancelling context after some delay + suggested sanity check for goroutines number (cool idea btw!)

pkg/cmd/agent/metadata.go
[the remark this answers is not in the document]
VAveryanov8: The only concerned of doing it per request is the case when metadata is not available (disabled or on-premise). Then every backup request will be slowed down by 5 sec. Otherwise I would do it per request as well!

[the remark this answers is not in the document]
VAveryanov8: We can also lazily fetch metadata! This way metadata won't be fetched until first request and then the results are cached. So we shouldn't have any issue during service start up. 
I've added a commit with lazy loading, so I'd like to hear want do you think about it

pkg/cloudmeta/metadata.go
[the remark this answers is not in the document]
karol-kokoszka: It's ok, thanks !

pkg/cloudmeta/metadata_test.go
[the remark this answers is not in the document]
karol-kokoszka: Ok, let it be. I see you added the 50ms waiting < testProvider latency.

pkg/cmd/agent/metadata.go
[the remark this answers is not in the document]
karol-kokoszka: Lazy loading looks ok from my PoV :+1:

karol-kokoszka (approved):
:+1:  from my side, but please address the comment with manifest validation in integration tests + get the +1 from @Michal-Leszczynski

pkg/service/backup/backupspec/manifest.go
VAveryanov8: done!

pkg/cmd/agent/agent.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/scylladb/scylla-manager/v3/swagger/gen/agent/models"
 )
 
-func newAgentHandler(c agent.Config, rclone http.Handler, logger log.Logger) *chi.Mux {
+func newAgentHandler(c agent.Config, rclone http.Handler, cloudMeta http.HandlerFunc, logger log.Logger) *chi.Mux {
```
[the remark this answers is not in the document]
VAveryanov8: I followed existing pattern here indeed :smile: But I can see reasoning behind it - handler shouldn't have knowledge of how to init(create) it dependencies, this allows us easily  write tests for this handler, e.g.  https://github.com/scylladb/scylla-manager/pull/4177/files#diff-8f7e9dfb2db5ec6ff6b4fc4c0e99bf6d91dc47310606450273cdb296e8a2c278R82

pkg/cloudmeta/metadata.go
```diff
@@ -90,20 +90,32 @@ func (cloud *CloudMeta) GetInstanceMetadata(ctx context.Context) (InstanceMetada
 			select {
 			case <-ctx.Done():
 				return
-			case results <- msg{meta: meta, err: err}:
+			default:
 			}
+
+			results <- msg{meta: meta, err: err}
```
Michal-Leszczynski, line 96: I think that you can replace:
```go
select {
case <-ctx.Done():
	return
default:
}
```
with simple:
```go
if ctx.Err != nil {
	return
}
```

But apart from that, why do we even want to check for the context error here?
Can't we just get the metadata and push it to the chan?
If the call to meta failed because of context error, it will be reflected in the returned error.
We don't need to worry about subtle cases of when call to meta succeeded right before the context was cancelled.

pkg/cloudmeta/metadata.go
```diff
@@ -90,20 +90,32 @@ func (cloud *CloudMeta) GetInstanceMetadata(ctx context.Context) (InstanceMetada
...
+		select {
+		case <-ctx.Done():
+			return InstanceMetadata{}, ctx.Err()
+		case res := <-results:
+			// Additional context check just in case messages in results and in ctx.Done channels were available at the same time.
+			select {
+			case <-ctx.Done():
+				return InstanceMetadata{}, ctx.Err()
+			default:
+			}
```
Michal-Leszczynski, line 112: Same for simplifying the check for context error as above.

But apart from that, why do we worry about what you wrote in the comment.
Aren't we completely fine with that? What scenario does it improve?

pkg/cmd/agent/server.go
```diff
@@ -160,10 +160,14 @@ func (s *server) makeServers(ctx context.Context) error {
 	if err != nil {
 		return errors.Wrapf(err, "tls")
 	}
+	cloudMeta, err := newMetadataHandler(s.logger.Named("metadata"))
+	if err != nil {
+		return errors.Wrap(err, "newMetadataHandler")
+	}
```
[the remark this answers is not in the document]
VAveryanov8: yes, indeed, maybe it's not worth to failing the entire agent if aws config has some error in it :) 

done!

pkg/cloudmeta/metadata.go
VAveryanov8: thanks, I've simplified this check :) 

> We don't need to worry about subtle cases of when call to meta succeeded right before the context was cancelled.

GetInstanceMetadata function works this way:
1. Runs goroutines with aws, gcp, azure providers
2. Once first non-err result is received, this function returns. This return triggers context cancellation (see `defer cancel()` 
3. Other providers (all except first) will have cancelled context. 
4. We check for a canceled context in each provider to make a best effort to avoid sending messages to the results channel, as nobody will read them anymore.

So case when context is cancelled actually quite regular - each invocation of this func :)

pkg/cloudmeta/metadata.go
VAveryanov8: Thanks simplified! 

Well basically by doing additional check we give a priority to context cancellation. This way we can have predictable behavior when both context.Done and results channels have some messages in the same time.

pkg/cmd/agent/metadata.go
```diff
@@ -0,0 +1,61 @@
...
+		return metadata, nil
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		instanceMeta, err := lazyGetMetadata()
+		if err != nil {
+			// Metadata may not be available for several reasons:
+			// 1. running on-premise 2. disabled 3. smth went wrong with metadata server.
+			// As we cannot distinguish between these cases, we can only log err.
+			logger.Error(r.Context(), "GetInstanceMetadata", "err", err)
```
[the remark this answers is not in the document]
VAveryanov8: done!

pkg/cloudmeta/metadata.go
Michal-Leszczynski: Yeah, it's just that I don't see much value in this case:
> We check for a canceled context in each provider to make a best effort to avoid sending messages to the results channel, as nobody will read them anymore.

I expect that cases when the API call finishes and context is cancelled in the very very similar moment are really really rare, but even if they happen, we just send something on a channel and that's it.
I'm not saying that current behavior is bad, but rather that it might worry about situations which are not worthy to worry about. Having said that, I'm fine with the current state!

Michal-Leszczynski (approved):
LGTM!

VAveryanov8:
@karol-kokoszka @Michal-Leszczynski 

I've updated integration test with additional checks for new manifest fields. If you don't have any comments regarding it I can proceed with squashing and merging :smile:

karol-kokoszka (approved):

karol-kokoszka:
Yes, proceed.

go.mod
VAveryanov8: I've squashed changed in a way it seemed logical to me, but I'd appreciate a second pair of eyes  :)

Michal-Leszczynski: LGTM!

PR #4178, opened 2024-12-18
Healthcheck: correctly label nodes removed from metrics

### LINKED ISSUE #4017: Metrics from changed ip address nodes
Problem with scylla-manager.
He returns "scylla_manager_healthcheck_cql_status{} = -1" for scylladb nodes in k8s where the pod was restarted and the IP address was changed. In Database he have all correct ip address in table scylla-manager.cluster

SM   "version": "3.4.0-dev-0.20240822.0ad5692f0"

Michal-Leszczynski:
Commits a75df89c and 3cdd20dc added and started setting new labels (dc/rack) in healthcheck metrics.
Unfortunately, those labels weren't taken into consideration when removing metrics of removed clusters/nodes:
https://github.com/scylladb/scylla-manager/blob/cafa851a39fc4d1238b4d2827f11058dd81158a1/pkg/service/healthcheck/runner.go#L116-L121
This resulted in not removing those metrics until SM restart.

This commit fixes this issue and brings a small refactor to the way in which healthcheck labels are applied, so that it's more difficult to make such mistake in the future.

Fixes #4017

### DIFF
--- a/pkg/service/healthcheck/metrics.go
+++ b/pkg/service/healthcheck/metrics.go
@@ -8,14 +8,42 @@ import (
 )
 
 const (
-	clusterKey  = "cluster"
-	hostKey     = "host"
-	dcKey       = "dc"
-	pingTypeKey = "ping_type"
-	rackKey     = "rack"
+	clusterKey = "cluster"
+	dcKey      = "dc"
+	rackKey    = "rack"
+	hostKey    = "host"
 
 	metricBufferSize = 100
 )
+
+func labelNames() []string {
+	return []string{clusterKey, dcKey, rackKey, hostKey}
+}
+
+type labels struct {
+	cluster string
+	dc      string
+	rack    string
+	host    string
+}
+
+func newLabels(cluster, dc, rack, host string) labels {
+	return labels{
+		cluster: cluster,
+		dc:      dc,
+		rack:    rack,
+		host:    host,
+	}
+}
+
+func (ls labels) promLabels() prometheus.Labels {
+	return prometheus.Labels{
+		clusterKey: ls.cluster,
+		dcKey:      ls.dc,
+		rackKey:    ls.rack,
+		hostKey:    ls.host,
+	}
+}
 
 var (
 	cqlStatus = prometheus.NewGaugeVec(prometheus.GaugeOpts{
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
The fix was tested manually by decommissioning a node from a fresh testing cluster.

Before the fix:
image
After the fix:
image

Note that the metrics for the removed node will keep on showing `-1` until it's removed from the config cache svc, which can take up to 5 minutes, but I guess that's not really a problem. Before the fix, no matter how long I waited, the metrics weren't removed at all. After the fix it's clearly visible that after some time the metrics were removed.

Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 not sure about automated integration testing.
It would probably require some bash scripts using docker commands - I don't think that we have the infrastructure for that in place and that it would be too much hassle to properly implement it.

pkg/service/healthcheck/metrics.go
```diff
@@ -8,57 +8,85 @@ import (
...
 const (
-	clusterKey  = "cluster"
-	hostKey     = "host"
-	dcKey       = "dc"
-	pingTypeKey = "ping_type"
-	rackKey     = "rack"
+	clusterKey = "cluster"
+	dcKey      = "dc"
+	rackKey    = "rack"
+	hostKey    = "host"
```
VAveryanov8, line 14: just out of curiosity: Why ping_type label has been removed?

Michal-Leszczynski: It wasn't used anywhere in the code, so I deleted it.

I think that in the past we had a single metric for all types of healthchecks (REST/CQL/alternator) with a label defining its type. Now we have different metrics for different healthcheck types, but someone probably forgot to remove this ping_type key .

VAveryanov8: got it, thanks!

VAveryanov8 (approved):

karol-kokoszka (approved):
If there is a test for that in integration-tests, it would catch this bug before releasing scylla-manager with the bug.
But I agree it's too much effort to add it and to cover that scenario. Especially that we are not really testing metrics in e2e scenarios.

PR #4192, opened 2025-01-08
Use Scylla API for restore

Michal-Leszczynski:
WIP


schema/v3.5.0.cql
```diff
@@ -1,4 +1,27 @@
 ALTER TYPE restore_view ADD build_status text;
-ALTER TABLE restore_run_progress DROP skipped;
-ALTER TABLE restore_run_progress ADD restored bigint;
-ALTER TABLE backup_run_progress ADD scylla_task_id text;
\ No newline at end of file
+ALTER TABLE backup_run_progress ADD scylla_task_id text;
+
+DROP TABLE restore_run_progress;
```
VAveryanov8, line 4: As an alternative to dropping progress history we can try to do smth like: 
1. create temp table 
2. copy data from restore_run_progress to temp table
3. drop restore_run_progress
4. create restore_run_progress with new primary key
5. copy data from temp table to restore_run_progress
6. drop temp table 

It's an additional work, but then we can keep user history and we can safe some time trying to communicate why we need to drop table :)

pkg/service/restore/batch.go
```diff
@@ -9,12 +9,15 @@ import (
...
 	"github.com/pkg/errors"
 	. "github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
+	"github.com/scylladb/scylla-manager/v3/pkg/sstable"
 )
 
 // batchDispatcher is a tool for batching SSTables from
 // Workload across different hosts during restore.
 // It follows a few rules:
 //
+// - all SSTables within a batch have the same have the same batchType
```
VAveryanov8, line 19: typo

pkg/service/restore/batch.go
```diff
@@ -188,21 +204,36 @@ func (b batch) VersionedSize() int64 {
...
+
 // ValidateAllDispatched returns error if not all SSTables were dispatched.
 func (bd *batchDispatcher) ValidateAllDispatched() error {
 	bd.mu.Lock()
 	defer bd.mu.Unlock()
 
 	for i, rdp := range bd.workloadProgress.remoteDir {
-		if rdp.RemainingSize != 0 || len(rdp.RemainingSSTables) != 0 {
+		failed := rdp.RemainingSize != 0
+		for _, ssts := range rdp.RemainingSSTables {
```
VAveryanov8, line 228: if i understand correctly, we can have a situation when RemaningSize is 0, but there are still some RemaningSSTables? How this can happen?

schema/v3.5.0.cql
Michal-Leszczynski: If it was a backup/repair task, I would try harder to preserve task history, but I thought that dropping restore task history is fine for the users. Anyway, I will try to preserve it.

We even have some `pkg/schema/migrate` pkg.
I'm not familiar with it, but it has some functions for rewriting the table, so it should be rather easy!

pkg/service/restore/batch.go
Michal-Leszczynski: It shouldn't. I just wanted to keep on checking both size and sstables count just to be on the safe side (as it was done before).

To be honest, checking size seems less robust (maybe 0 size sstable is not realistic, but perhaps we might encounter some bug where Rclone reports 0 size by mistake).

I think that we can get rid of relying on remaining size and use sstable count instead.
How do you find this idea?

karol-kokoszka (approved):
Thanks @Michal-Leszczynski !
I have no comments. 

I see that latest scylla-enterprise:nightly-build supports backup/restore API and it's tested against all integration tests, what is v good.

pkg/service/restore/worker_scylla_restore.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+
+		task, err := w.client.ScyllaWaitTask(ctx, pr.Host, pr.ScyllaTaskID, int64(w.config.LongPollingTimeoutSeconds))
+		if err != nil {
+			return errors.Wrap(err, "wait for task")
+		}
+		w.scyllaUpdateProgress(ctx, pr, b, task)
+		switch scyllaclient.ScyllaTaskState(task.State) {
+		case scyllaclient.ScyllaTaskStateFailed:
+			return errors.Errorf("task error (%s): %s", pr.ScyllaTaskID, task.Error)
```
Michal-Leszczynski, line 115: I guess we shouldn't abort failed restore task (cc: @kreuzerkrieg)

pkg/service/restore/tablesdir_worker.go
```diff
@@ -252,9 +248,11 @@ func (w *tablesWorker) onLasStart(ctx context.Context, b batch, pr *RunProgress)
 func (w *tablesWorker) onLasEnd(ctx context.Context, b batch, pr *RunProgress) {
 	w.metrics.SetRestoreState(w.run.ClusterID, b.Location, w.target.SnapshotTag, pr.Host, metrics.RestoreStateIdle)
 	pr.setRestoreCompletedAt()
-	pr.Restored = pr.Downloaded + pr.VersionedProgress
+	pr.Restored = pr.Downloaded + pr.VersionedDownloaded
 	w.metrics.IncreaseRestoreStreamedBytes(w.run.ClusterID, pr.Host, b.Size)
-	w.metrics.IncreaseRestoreStreamDuration(w.run.ClusterID, pr.Host, timeSub(pr.RestoreStartedAt, pr.RestoreCompletedAt, timeutc.Now()))
+	w.metrics.IncreaseRestoreStreamDuration(w.run.ClusterID, pr.Host, timeSub(pr.DownloadCompletedAt, pr.RestoreCompletedAt, timeutc.Now()))
+	w.metrics.IncreaseRestoredBytes(w.run.ClusterID, pr.Host, b.Size)
+	w.metrics.IncreaseRestoreDuration(w.run.ClusterID, pr.Host, timeSub(pr.RestoreStartedAt, pr.RestoreCompletedAt, timeutc.Now()))
```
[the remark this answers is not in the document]
Michal-Leszczynski: In general, things like bandwidth and performance are better to be displayed as a part of metrics, not sctool.
So the restore metric was added as a counterpart to the already existing download and stream metrics, which are not really applicable to the native restore API. It's also done for the Rclone restore because this is a general effort of moving towards having a single metric for tracking restore progress and not dividing it into different stages which might be different for different Scylla/SM versions.

Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 I addressed/responded to the comments, please take a look!

pkg/schema/nopmigrate/nopmigrate.go
```diff
@@ -18,7 +21,155 @@ func nopCallback(context.Context, gocqlx.Session, migrate.CallbackEvent, string)
 func init() {
 	reg.Add(migrate.CallComment, "rewriteHealthCheck30", nopCallback)
 	reg.Add(migrate.CallComment, "setExistingTasksDeleted", nopCallback)
+	reg.Add(migrate.CallComment, "MigrateToRestoreRunProgressWithSortKeyWithScyllaTaskID", MigrateToRestoreRunProgressWithSortKeyWithScyllaTaskID)
 }
 
 // Callback is a sibling of migrate.Callback that registers all mandatory callbacks as nop.
 var Callback = reg.Callback
+
+// MigrateToRestoreRunProgressWithSortKeyWithScyllaTaskID needs to be implemented here because
```
VAveryanov8, line 30:
```suggestion
// MigrateToRestoreRunProgressWithSortKeyWithScyllaTaskID needs to be implemented here 
```

PR #4210, opened 2025-01-14
Kk/extract manifest manipulation to utils

karol-kokoszka:
This PR is used as a proof-of-concept for the idea of sharing the manifest manipulation part with the Scylla Cloud, so that SC can import the manifest dependency and just use it to understand snapshot backup content.

---

Please make sure that:


Michal-Leszczynski:
I know it's just a playground, but before going in this direction, we should carefully choose only the interesting subset `backupspec` instead of the whole pkg.

Issue #3829, opened 2024-04-29
Give possibility for restoring DC using mapping sourceDC -> destinationDC


karol-kokoszka:
https://github.com/scylladb/scylla-manager/issues/3871

Right now, there is no option in the Scylla Manager restore task to restore just a single data center (DC) from the backup location. This could lead to problematic situations, particularly when:

Encryption at Rest (EaR) is enabled,
Two DCs use different encryption keys,
Encryption keys are stored in different cloud regions, and
There is only one backup location available.
To address this, we would need to make the encryption keys multi-regional to facilitate the restoration process in such scenarios.

The `location` flag may not be very intuitive, as the `[dc]` part defines the destination DC, not the source DC data. We need to discuss during the manager planning to determine if a new flag specifying the source is necessary. 
If we can restore just a single DC, then we can restore DC by DC, avoiding the need to create multi-regional keys.

(cc: @tzach)

tzach:
> To address this, we would need to make the encryption keys multi-regional to facilitate the restoration process in such scenarios.

Agree, but how this is a Scylla Manager issue to fix?

karol-kokoszka:
We could potentially address the problem by allowing to restore just a single DC from the location bucket.
It's something what we doesn't support at the moment (possibly by a mistake).

rayakurl:
@tzach - we need a resolution. for now almost all sct tests are failing since they are multi DC. We will add a couple of pipelines for a single DC + encryption  but are are disabling the multi DC jobs as they constantly failing.  @mikliapko as discussed, please create a task for the new pipelines. and disable the multi DC ones for now. Thanks

karol-kokoszka:
**grooming notes**

The initial idea is to add new flag to the restore CLI, so that it's possible to define the origin DC from the backup location.
Then, data from this DC is going to be restored to specified destination.

@mikliapko SCT will have to be updated to test the scenario with restoring single DC.

Michal-Leszczynski:
> The initial idea is to add new flag to the restore CLI, so that it's possible to define the origin DC from the backup location.
> Then, data from this DC is going to be restored to specified destination.

After giving it some more thought, I wouldn't recommend adding it in such way.
The need for this feature raised from https://github.com/scylladb/scylla-manager/issues/3871, where it could be used to restore DC by DC.
This is problematic, as restore task does not only download and load&stream the data, but it also:
- disables and enables tombstone_gc 
- drops and creates views
- runs repair

[...]

PR #4213, opened 2025-01-16
feat(restore): adds --dc-mapping flag to restore command

VAveryanov8:
This adds support for `--dc-mapping` flag to restore command. It specifies mapping between DCs from the backup and DCs in the restored(target) cluster.  Only 1 use case is supported: 1-1 dc mapping. This means that squeezing (restore dc1 and dc2 into dc3) or extending (restore dc1 into dc1 and dc2) DCs is not supported when `--dc-mapping` is provided.
So the syntax is:
```
source_dc1=target_dc1,source_dc2=target_dc2
Where
     equal(=) is used to separate source   dc name and target dc name
     comma(,)  is used to separate multiple mappings
```
If `--dc-mapping` is not provided, then current behavior should be preserved - each node with access to DC can download it data. Also it's allowed to provide only subset of DCs, ignoring source dc or target (or both).
Only works with tables restoration (`--restore-tables=true`).

Fixes: https://github.com/scylladb/scylla-manager/issues/3829

---

Please make sure that:


Michal-Leszczynski:
In terms of syntax:
> "dc1,!dc2=>dc2"     - data from dc1 should be restored to dc2 DC. Ignoring dc2 from source cluster.

I find it confusing that it's possible to specify both restored and skipped DCs in the same mapping key.
It reads like "restore dc1 and not dc2 into dc2" instead of "don't restore dc2 and restore dc1 into dc2".
IMHO it should look like:
```
 "!dc2;dc1=>dc2"     - data from dc1 should be restored to dc2 DC. Ignoring dc2 from source cluster.
```
I would suggest to validate that positive and negative DC occurrences are not mixed in the mapping key and that negative DCs can't be mapped to anything, so that we avoid confusion and typing mistakes.

pkg/service/restore/batch.go
```diff
@@ -121,7 +122,9 @@ func newWorkloadProgress(workload Workload, locationHosts map[Location][]string)
 	hostDCAccess := make(map[string][]string)
 	for loc, hosts := range locationHosts {
 		for _, h := range hosts {
-			hostDCAccess[h] = append(hostDCAccess[h], locationDC[loc.StringWithoutDC()]...)
+			dcsInLoc := locationDC[loc.StringWithoutDC()]
+			hostAllDCs := hostDCs[h]
+			hostDCAccess[h] = append(hostDCAccess[h], strset.Intersection(strset.New(dcsInLoc...), strset.New(hostAllDCs...)).List()...)
```
Michal-Leszczynski, line 127: I don't think that we should just make an intersection here.
If someone specified that given DC should be restored by given DC, then all hosts from this DC should have access to the appropriate location. This should be done inside the `GetTarget` initial validation.

pkg/service/restore/model.go
```diff
@@ -33,9 +34,14 @@ type Target struct {
...
 	RestoreTables   bool       `json:"restore_tables,omitempty"`
 	Continue        bool       `json:"continue"`
+	DCMappings      DCMappings `json:"dc-mapping"`
 
 	// Cache for host with access to remote location
 	locationHosts map[Location][]string `json:"-"`
+	// Cache for host and their DC after applying DCMappings
+	hostDCs map[string][]string
+	// Cache for dcs that shouldn't be restored from location
+	ignoredSourceDC []string
```
Michal-Leszczynski, line 44: Wouldn't it be way simpler if we just had a single mapping `map[string][string]` of dst DC -> src DC list?
It could replace all `locationHosts, hostDCs, ignoredSourceDC`.
During batch dispatch we could/should have access to given hosts DC, so it would give us enough information for correctness.
Ignored DCs could be replaced with restored DCs which could be easily obtained from such mapping when needed for filtering.

The assumption is that we don't need to care for cases like: what if only a few hosts from this DC have access to this location. Even when making backup, the whole DC needs to be backed up to a single location and all hosts from given DC needs to have access to it.

pkg/service/restore/worker.go
```diff
@@ -161,12 +161,96 @@ func (w *worker) initTarget(ctx context.Context, properties json.RawMessage) err
 		} else {
 			w.logger.Info(ctx, "Found schema file")
 		}
+		return nil
+	}
+
+	// The only way to have a list of dcs from the source cluster is to
+	// get them from backup locations.
+	sourceDC, err := w.collectDCFromLocation(ctx, t.locationHosts)
```
Michal-Leszczynski, line 169: Correct, but we are already going through all of the manifests in the `GetTarget` (`initUnits`), so we should probably get all of the interesting data within a single iteration.

Makefile
```diff
@@ -115,6 +115,7 @@ INTEGRATION_TEST_ARGS := -cluster $(PUBLIC_NET)100 \
 -managed-cluster $(PUBLIC_NET)11,$(PUBLIC_NET)12,$(PUBLIC_NET)13,$(PUBLIC_NET)21,$(PUBLIC_NET)22,$(PUBLIC_NET)23 \
 -test-network $(PUBLIC_NET) \
 -managed-second-cluster $(PUBLIC_NET)31,$(PUBLIC_NET)32 \
+-managed-third-cluster $(PUBLIC_NET)41,$(PUBLIC_NET)42 \
```
Michal-Leszczynski, line 118: Having 3 clusters for the docker setup seems excessive.
You have written nice unit tests, so we don't need to cover all scenarios with integration tests.
Having said that, could we simply rename the DC in the second cluster (used just for the restore)?
We could even divide this cluster into 2 single nodes DCs if that could help.

Michal-Leszczynski:
@VAveryanov8 I made a general review, but I didn't dive into other details, because it would be better to first discuss the current comments (other things might not matter after that).

Michal-Leszczynski:
Also, we should deprecate the datacenter from the `--location` flag, as it was previously used to partially control the dc mapping. When dc mapping flag is present, it should be ignored. When it's not, it can still be respected, but in the implementation we would still just parse it into dc mapping on the SM side.

pkg/command/restore/dcmappings.go
```diff
@@ -0,0 +1,70 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+type dcMappings []dcMapping
+
+type dcMapping struct {
+	Source []string `json:"source"`
+	Target []string `json:"target"`
```
karol-kokoszka, line 15: The DC mapping should be single string to single string.
DC1 (of the source cluster) => DC2 (of the destination cluster) ; DC2 (of the source) => DC1 (of the target)

And that's it, no need for the complex logic.

karol-kokoszka (commented):
@VAveryanov8 Thanks for the extensive PR and for covering a lot of different scenarios, but we don't need such a complex logic.

It's enough to do the one to one mapping. Instead of allowing to merge multiple DCs in source or multiple DCs in target.

Let's discuss the goal of DC mapping on the call.

It's enough to have a possibility of:
- restoring just a single DC (that's why I don't really see the reason for introducing another flag that skips validation)
- explicitly defining that nodes from target DCX are expected to download data of source DCY and call to load & stream

karol-kokoszka:
@VAveryanov8 I see some recent pushes to this PR, is it ready for re-review ?

karol-kokoszka (commented):
Thanks !

@mikliapko There are SCT tests that were trying to restore multiDC cluster, but they were failing due to the problems with encryption at rest enabled on the cluster. 
Two datacenters were using different encryption keys.

This PR is expected to fix this problem. AFAIR, you changed these tests to singleDC later. Is it possible to validate this PR against the SCT tests with the multiDC restore with encryption at rest enabled ?

karol-kokoszka:
@VAveryanov8 Please rebase on master. You will have conflicts related to the new submodule, but they are suppose to be very easy to fix.

mikliapko:
> @mikliapko There are SCT tests that were trying to restore multiDC cluster, but they were failing due to the problems with encryption at rest enabled on the cluster. Two datacenters were using different encryption keys.

Yep, we can do it, just give me the SM build that can be used

karol-kokoszka:
@mikliapko Here is the manager-build triggered on the current branch https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-build/919/

You will have to edit the test to include `--dc-mapping source_dc1=>target_dc1;source_dc2=>target_dc2` into the sctool restore call.

pkg/service/restore/restore_integration_test.go
```diff
@@ -1072,3 +1072,89 @@ func TestRestoreTablesProgressIntegration(t *testing.T) {
...
+			"keyspace": ksFilter,
+			// Test if batching does not hang with
+			// limited parallel and location access.
+			"parallel":       1,
+			"snapshot_tag":   tag,
+			"restore_tables": true,
+			// DC Mapping is required
+			"dc_mapping": []map[string]string{
+				{"source": "dc1", "target": "dc1"},
+			},
```
Michal-Leszczynski, line 1138: Why `[]map[string]string`? Can't we use the `DCMappings` type?

pkg/service/restore/restore_integration_test.go
```diff
@@ -1072,3 +1072,89 @@ func TestRestoreTablesProgressIntegration(t *testing.T) {
 		}
 	}
 }
+
+func TestRestoreOnlyOneDCFromLocationIntegration(t *testing.T) {
```
Michal-Leszczynski, line 1076: So this test checks if we can restore a subset of DCs.
Isn't this whole PR more about ensuring that nodes from given DC restore data from given DC?
Restoring just a subset of DCs is kind of a byproduct of the `--dc-mapping` flag.

Maybe it would be beneficial to add a test which checks that (e.g. with the help of interceptors).
It would be difficult to do it in the context of `TestRestoreOnlyOneDCFromLocationIntegration`, because it requires restore into a cluster with multiple DCs.

testing/scylla/config/scylla.yaml
```diff
@@ -645,4 +645,4 @@ api_doc_dir: /usr/lib/scylla/api/api-doc/
 alternator_port: 8000
 alternator_write_isolation: only_rmw_uses_lwt
 alternator_enforce_authorization: true
-enable_ipv6_dns_lookup: true
\ No newline at end of file
+enable_ipv6_dns_lookup: true
```
Michal-Leszczynski, line 648: Why?

pkg/service/restore/service_restore_integration_test.go
```diff
@@ -772,7 +772,7 @@ func restoreWithAgentRestart(t *testing.T, target Target, keyspace string, loadC
 
 	// Recreate schema on destination cluster
 	if target.RestoreTables {
-		WriteDataSecondClusterSchema(t, dstSession, keyspace, 0, 0)
+		WriteData(t, dstSession, keyspace, 0)
```
Michal-Leszczynski, line 775: Why? The same for other similar changes.

pkg/service/restore/worker.go
```diff
@@ -45,37 +45,151 @@ type worker struct {
...
+
+func collectDCsFromManifests(manifests []*backupspec.ManifestInfo) []string {
+	dcs := strset.New()
+	for _, m := range manifests {
+		dcs.Add(m.DC)
+	}
+	return dcs.List()
+}
+
+// keep only manifests that have dc mapping. if --dc-mapping is not set it will return all manifests.
```
Michal-Leszczynski, line 172: Lower case comments

pkg/service/restore/worker.go
```diff
@@ -45,37 +45,151 @@ type worker struct {
...
+	if len(nodes) == 0 {
+		return nil, fmt.Errorf("no nodes with location %s access", loc)
+	}
+	return nodes, nil
+}
+
+// hostsByDC creates map of which hosts are responsible for which DC, also applies DCMappings if available.
+func hostsByDC(nodes scyllaclient.NodeStatusInfoSlice, targetDC2SourceDCMap map[string]string, locationDCs []string) map[string][]string {
+	dc2HostsMap := map[string][]string{}
+	// when --dc-mapping is not set all nodes with access to the location can handle all DCs from it
```
Michal-Leszczynski, line 139: This comment is misleading.
If `--dc-mapping` is not set, then we take location's DC into consideration, which results in passing `nodes` only from this DC into this function. Even if it is already taken care of before calling this function, we should rephrase this comment so that it does not create confusion.

pkg/service/restore/worker.go
```diff
@@ -162,12 +230,56 @@ func (w *worker) initTarget(ctx context.Context, properties json.RawMessage) err
...
+		targetDC := slices.Collect(maps.Keys(dcMap))
+		if err := w.validateDCMappings(t.DCMappings, sourceDC.List(), targetDC); err != nil {
+			w.logger.Debug(ctx,
+				"Validate dc mapping",
+				"source_dc", sourceDC,
+				"target_dc", targetDC,
+				"mappings", t.DCMappings,
+			)
+			return err
+		}
```
Michal-Leszczynski, line 250: So we log it only in case of error, but we don't log the error itself?
I guess we should either:
- wrap error with all interesting info and don't log anything
- log it not only in case of an error

pkg/service/restore/worker_test.go
```diff
@@ -0,0 +1,181 @@
...
+			},
+			locationDCs: []string{"source_dc1", "source_dc2"},
+
+			expected: map[string][]string{
+				"source_dc1": {"n1", "n2"},
+				"source_dc2": {"n3", "n4"},
+			},
+		},
+		{
+			name: "When --dc-mapping is provided will some DCs",
```
Michal-Leszczynski, line 124: `will` -> `with`?

pkg/service/restore/worker_test.go
```diff
@@ -0,0 +1,181 @@
...
+			name:     "sourceDC != targetDC, but with full mapping",
+			sourceDC: []string{"dc1"},
+			targetDC: []string{"dc2"},
+			dcMappings: []DCMapping{
+				{Source: "dc1", Target: "dc2"},
+			},
+			expectedErr: false,
+		},
+		{
+			name:     "source != target, but will full mapping, two dcs per cluster",
```
Michal-Leszczynski, line 30: `will` -> `with`?

pkg/command/restore/dcmappings.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/pkg/errors"
+)
+
+type dcMappings []dcMapping
+
+type dcMapping struct {
+	Source string `json:"source"`
+	Target string `json:"target"`
+}
```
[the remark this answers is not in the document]
Michal-Leszczynski: After going through this PR it seems like we indeed have a lot of code regarding parsing/testing this representation.

mikliapko:
> @mikliapko Here is the manager-build triggered on the current branch https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-build/919/
> 
> You will have to edit the test to include `--dc-mapping source_dc1=>target_dc1;source_dc2=>target_dc2` into the sctool restore call.

Testing with the provided build.

To make it work, dc mapping should be taken in quotes (see examples below):
```
ubuntu@manager-regression-adapt-fo-monitor-node-55ef36da-1:~$ sctool restore -c test --restore-tables --location s3:manager-backup-tests-us-east-1 --snapshot-tag sm_20250210120404UTC --dc-mapping eu-west-2scylla_node_west=>eu-west-2scylla_node_west;us-eastscylla_node_east=>us-eastscylla_node_east
Error: invalid argument "eu-west-2scylla_node_west=" for "--dc-mapping" flag: invalid syntax, mapping should be in a format of sourceDcs=>targetDcs, but got: eu-west-2scylla_node_west=

-bash: us-eastscylla_node_east=: command not found
ubuntu@manager-regression-adapt-fo-monitor-node-55ef36da-1:~$ sctool restore -c test --restore-tables --location s3:manager-backup-tests-us-east-1 --snapshot-tag sm_20250210120404UTC --dc-mapping "eu-west-2scylla_node_west=>eu-west-2scylla_node_west;us-eastscylla_node_east=>us-eastscylla_node_east"
restore/e3cc3b69-0c48-4109-ad78-fbe6257f6eb9
```

@VAveryanov8 @karol-kokoszka Is it inevitable in this case?

VAveryanov8:
> To make it work, dc mapping should be taken in quotes (see examples below):

For this particular syntax yes, quotes are required.

But the good news is that I'm gonna get rid of it toward more simple syntax using just equal(=) instead of array(=>) and comma(,) and instead of semi-column(;). Then I suppose it will be possible to use without quotes.

pkg/command/restore/dcmappings.go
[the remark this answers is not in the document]
VAveryanov8: Great finding! Using StringToString drastically simplified this PR!

testing/nodes_exec
```diff
@@ -21,4 +21,4 @@ for name in $(docker ps -f name=dc1_node -f name=dc2_node --format {{.Names}});
         echo "> ${name}"
         docker exec ${name} bash -c "$*"
     fi
-done
\ No newline at end of file
+done
```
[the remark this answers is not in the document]
VAveryanov8: setting of my editor I think, because I did some changes in this file and then removed them. By posix standard line should always end with newline character, that's why I decided to leave it like this

pkg/service/restore/service_restore_integration_test.go
VAveryanov8: reverted this change - it's an artifacts from previous version of this PR - when second cluster had dc named dc3. Here and in other parts of the code, we use WriteDataSecondClusterSchema to create second cluster schema in the first cluster which is confusing :)

pkg/service/restore/worker.go
VAveryanov8: removed this log, it's not needed anymore

pkg/service/restore/restore_integration_test.go
VAveryanov8: Good idea! I've added another tests with checks that nodes downloading tables only from corresponding DCs. I've leveraged info from restore_run_progress table to validate that - i think it's better then interceptors

pkg/service/restore/worker.go
```diff
@@ -45,37 +45,160 @@ type worker struct {
...
+	if len(nodes) == 0 {
+		return nil, fmt.Errorf("no nodes with location %s access", loc)
+	}
+	return nodes, nil
+}
+
+// hostsByDC creates map of which hosts are responsible for which DC, also applies DCMappings if available.
+func hostsByDC(nodes scyllaclient.NodeStatusInfoSlice, targetDC2SourceDCMap map[string]string, locationDCs []string) map[string][]string {
+	dc2HostsMap := map[string][]string{}
+	// When --dc-mapping is not set all nodes (or nodes from location.DC) with access to the location can handle all DCs (or location.DC) from it
```
Michal-Leszczynski, line 148: Is the second `(or location.DC)` needed?
Setting location's DC means that nodes from this DC will be restoring this location.
It doesn't mean that the only data center restored from this location will be this DC.

Also a small nit, this comment could be broken into multiple lines.

Michal-Leszczynski (commented):
The PR looks nice!
Just some question about the comment.

karol-kokoszka:
@mikliapko Here is the latest build from this branch https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-build/923/
It contains all recent changes.

Michal-Leszczynski (approved):

mikliapko:
> @mikliapko Here is the latest build from this branch https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-build/923/ It contains all recent changes.

Retesting here https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-sanity-test/69/

mikliapko:
> Retesting here https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-sanity-test/70/

Okay, the test is green

mikliapko:
> But does it fail on the master branch ?

I haven't run it there since it doesn't contain the fix for multiDC restore.
So, it's supposed to fail.

karol-kokoszka:
@VAveryanov8 feel free to merge it.

PR #4214, opened 2025-01-17
Swagger: scylla-manager, move towards just restored bytes

Michal-Leszczynski:
In overall/keyspace/table progress we were reporting both downloaded and restored bytes. There are two problems with it:

1. Users does not really care for the distinction between downloaded and restored bytes - they just want to know the amount of bytes that have been successfully restored.

2. With new native Scylla API restore, we no longer separate restoration into download/load&stream stages.

That's why we should move towards reporting just the amount of successfully restored bytes, and not keep on adding another fields to our API for each new type of restore.

This commit:
- Adds restored_bytes/duration to host progress, since it was lacking there. It does not deprecate the downloaded/streamed fields because they were just added in the previous minor release, so that would be confusing. Instead, for now we allow for host progress to contain detailed operation bandwidth information.
- Deprecates downloaded bytes in overall/keyspace/table progress, as we want to move towards observing just the restored bytes.

A general note is that we should use metrics for monitoring performance, not the sctool progress cmd, but since we already started going in this direction, we need to support it.

### DIFF
--- a/v3/swagger/gen/scylla-manager/models/restore_progress.go
+++ b/v3/swagger/gen/scylla-manager/models/restore_progress.go
@@ -23,7 +23,7 @@ type RestoreProgress struct {
 	// Format: date-time
 	CompletedAt *strfmt.DateTime `json:"completed_at,omitempty"`
 
-	// downloaded
+	// This field is DEPRECATED. Total bytes downloaded from table (included in restored)
 	Downloaded int64 `json:"downloaded,omitempty"`
 
 	// failed
--- a/v3/swagger/gen/scylla-manager/models/restore_host_progress.go
+++ b/v3/swagger/gen/scylla-manager/models/restore_host_progress.go
@@ -15,22 +15,28 @@ import (
 // swagger:model RestoreHostProgress
 type RestoreHostProgress struct {
 
-	// Total time spent by host on download in milliseconds
+	// Total time spent by host on download in milliseconds (included in restore_duration)
 	DownloadDuration int64 `json:"download_duration,omitempty"`
 
-	// Total bytes downloaded by host
+	// Total bytes downloaded by host (included in restored_bytes)
 	DownloadedBytes int64 `json:"downloaded_bytes,omitempty"`
 
 	// host
 	Host string `json:"host,omitempty"`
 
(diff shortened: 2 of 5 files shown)


v3/swagger/gen/scylla-manager/models/restore_progress.go
```diff
@@ -23,7 +23,7 @@ type RestoreProgress struct {
 	// Format: date-time
 	CompletedAt *strfmt.DateTime `json:"completed_at,omitempty"`
 
-	// downloaded
+	// This field is DEPRECATED. Total bytes downloaded from table (included in restored)
```
karol-kokoszka, line 26: I have a problem with deprecating it. 
Why it's deprecated ? We still support l&s restore handled with RClone.

What about clusters keeping the backup on GCS ?

karol-kokoszka (commented):
Why do you want to deprecate some of the fields ? 
Why not to introduce new fields (you actually did it) handling progress of the restore with Scylla API only ? 

`downloaded` properties cannot be deprecated yet. (even though the deprecation here is just a description).
Would be worth to extend the description of "restored_bytes" and "restore_duration" with an information that it's part of the restore handled by with ScyllaAPI.

v3/swagger/gen/scylla-manager/models/restore_progress.go
Michal-Leszczynski: We will support the restore with Rclone download+l&s, but we don't want to continue going in the direction of reporting each type of file progress in `sctool progress` (downloaded/l&s/restoredWithAPI/somethingNew bytes).
We want to focus on the thing that the user is interested in, so the amount of restored bytes or the general task progress, which is common for all restore types.
I deprecated it, but I still plan to fill this info - deprecation is just a note that we advise the user of this API to rely on the restored bytes, not downloaded bytes, which might be empty or removed in the future.

@karol-kokoszka @VAveryanov8 does this convince you?

karol-kokoszka: OK, let it be.

karol-kokoszka (approved):

PR #4225, opened 2025-01-23
Restore: rework progress aggregation

Michal-Leszczynski:
This PR does the following:
- reworks progress aggregation to rely on `iter.Seq` instead of direct DB access, which in turn allows for adding unit tests for progress aggregation
- fills the `RestoredBytes` and `RestoreDuration` introduced in  #4214

Note, the `RestoredBytes` and `RestoreDuration` need to be added to the `sctool progress` display as a part of a separate PR.


VAveryanov8 (approved):
Nice usage of iterators and tests!

Michal-Leszczynski:
Thanks @VAveryanov8!
I'm just wondering, which approach to iterator that can encounter an error do you prefer?

1. The one from this PR - iterator stores the error internally and it needs to be checked manually after iteration (similar to gocql iterator). It allows to hide the error handling from other parts of the code (might be good or bad).
2. Instead of `iter.Seq[K]`, use `iter.Seq2[K, error]` which allows for handling errors by the underlying code.

I think that the second is probably better in general, but the first approach seamed more tailored to this PR use case.

VAveryanov8:
> The one from this PR - iterator stores the error internally and it needs to be checked manually after iteration (similar to gocql iterator). It allows to hide the error handling from other parts of the code (might be good or bad).
Instead of iter.Seq[K], use iter.Seq2[K, error] which allows for handling errors by the underlying code.
I think that the second is probably better in general, but the first approach seamed more tailored to this PR use case.

Good question! I don't have  a strong opinion on this - iterators are quite new to the language and there are not that many projects that already using them. 
In general second approach is more flexible (err can be skipped and etc), but in this particular case I think what you did looks better.

Michal-Leszczynski:

karol-kokoszka (approved):

PR #4228, opened 2025-01-23
* feat(managerclient): make Items field in BackupListItems public for improved usability (#4227)

### LINKED ISSUE #4227: It is impossible to benefit from managerclient.ListBackups
It is not possible to retrieve a list of models.BackupListItem because it is private, which makes the method useless for managerclient users.

Image

aberry-21:
Fixed #4227


Michal-Leszczynski:
As per https://github.com/scylladb/scylla-manager/issues/4227#issuecomment-2621442857, I'm closing it for now.
Please let me know if there is a need for reopening it.

Issue #4201, opened 2025-01-13
Add validation stage to the 1-1 restore service


karol-kokoszka:
First step of the 1-1 restore performed by Scylla Manager is to validate that the destination cluster topology matches the source cluster topology.

Scylla Manager must:
- validate that all the nodes have an access to the backup location 
- validate that the mapping includes all the nodes
- validate that the source node belongs to the DC and rack defined by the mapping
- validate that the destination node belongs to the DC and rack defined by the mapping
- validate that the destination node has the equal amount of CPUs to the source node amount of CPUs
- validate that the destination node storage size is GE than the source node storage size
- validate that the destination node is the owner of the same tokens like the source node

This process must be parallelized per destination node (or source node).

PR #4230, opened 2025-01-27
feat(1-1-restore): validates if source and target clusters nodes are equal

VAveryanov8:
This adds a validation stage for 1-1-restore. The logic is as follows:
- Collect node information for the source cluster from backup manifests
- Collect node information for the target cluster from the Scylla API
- Apply node mappings to the source cluster nodes
- Compare each source node with its corresponding target node

Fixes: https://github.com/scylladb/scylla-manager/issues/4201

---

Please make sure that:


pkg/service/one2onerestore/worker_manifest.go
```diff
@@ -0,0 +1,44 @@
...
+	"strings"
+
+	. "github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
+
+	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
+)
+
+// getManifestInfo returns manifests with receiver's snapshot tag for all nodes in the location.
+func (w *worker) getManifestInfo(ctx context.Context, host, snapshotTag string, location Location) ([]*ManifestInfo, error) {
+	baseDir := path.Join("backup", string(MetaDirKind))
```
Michal-Leszczynski, line 18: Let's add this const path to the `backupspec` pkg.

pkg/service/one2onerestore/worker_validate.go
```diff
@@ -0,0 +1,247 @@
...
+
+		return nil
+	}, parallel.NopNotify)
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
+}
+
+func uniqHosts(locations []LocationInfo) []Host {
```
Michal-Leszczynski, line 172: This is a little strange to me - why can't we rely on the `nodesMapping` to get such information?
I think that it would simplify the flow of both `getTargetClusterNodeInfo` and `getSourceClusterNodeInfo` methods - there wouldn't be a need for matching hosts to manifests / locations from scratch.

pkg/service/one2onerestore/worker_validate.go
```diff
@@ -0,0 +1,247 @@
...
+	slices.SortFunc(targetNodeInfo, compareNodesInfo)
+
+	if err := clustersAreEqual(sourceNodeInfo, targetNodeInfo); err != nil {
+		return errors.Wrap(err, "clusters not equal")
+	}
+
+	return nil
+}
+
+func applyNodeMapping(sourceNodeInfo []nodeInfo, sourceMappings map[node]node) ([]nodeInfo, error) {
```
Michal-Leszczynski, line 43: What happens here and why do we need it?
Is is connected to the fact that src DCs and dst DCs might not be the same?

pkg/service/one2onerestore/worker_validate.go
```diff
@@ -0,0 +1,247 @@
...
+		if !ok {
+			return nil, errors.Errorf("mapping for source node (%v) is not found", source)
+		}
+		source.DC, source.Rack, source.HostID = target.DC, target.Rack, target.HostID
+		result = append(result, source)
+	}
+	return result, nil
+}
+
+func compareNodesInfo(a, b nodeInfo) int {
```
Michal-Leszczynski, line 59: nit: we could sort them just by the host ID.

pkg/service/one2onerestore/worker_validate.go
```diff
@@ -0,0 +1,247 @@
...
+		cmp.Compare(a.DC, b.DC),
+		cmp.Compare(a.Rack, b.Rack),
+		cmp.Compare(a.HostID, b.HostID),
+		cmp.Compare(a.CPUCount, b.CPUCount),
+		cmp.Compare(a.StorageSize, b.StorageSize),
+		slices.Compare(a.Tokens, b.Tokens),
+	)
+}
+
+func clustersAreEqual(sourceNodeInfo, targetNodeInfo []nodeInfo) error {
```
Michal-Leszczynski, line 70: Maybe let's rename it from `Equal` to something else, as we don't check the real equality here (e.g. storage size check). It's more about ensuring restore compatibility.

pkg/service/one2onerestore/worker_validate.go
```diff
@@ -0,0 +1,247 @@
...
+	if len(sourceNodeInfo) != len(targetNodeInfo) {
+		return errors.Errorf("clusters have different nodes count: source %d != target %d", len(sourceNodeInfo), len(targetNodeInfo))
+	}
+	for i, source := range sourceNodeInfo {
+		target := targetNodeInfo[i]
+		if source.DC != target.DC {
+			return errors.Errorf("source DC doesn't match target DC")
+		}
```
Michal-Leszczynski, line 78: So we can't restore into a different DC?
Is this expected, or is this something that will be implemented in the next PRs?

pkg/service/one2onerestore/worker_validate.go
VAveryanov8: As DC, Rack and HostID can have different values in source and target clusters,  but mapped one to one using provided `node-mapping`, we can replace all source DC, Rack and HostId with values from node-mapping and after that source nodes must be equal to target nodes.

pkg/service/one2onerestore/worker_validate.go
VAveryanov8: Up to this point nodes from source clusters must be already mapped to nodes from target cluster (see `applyNodeMapping(...)`). So if DC, Rack, HostId doesn't match it means that `node-mapping` was incorrect.

pkg/service/one2onerestore/worker_validate.go
VAveryanov8: In this particular case, I didn't use `nodesMapping`, because it's not validated yet :) So in theory it can has less node defined than actual cluster have. Also as `nodesMapping` doesn't have a host address, I wasn't sure how safe (or reliable) is to map nodes only by DC+HostID, as we don't yet have Rack information.

Michal-Leszczynski: Host ID alone should be good enough identifier.
But here, the only place where this function is used is:
```go
func (w *worker) getTargetClusterNodeInfo(ctx context.Context, locations []LocationInfo) ([]nodeInfo, error) {
	hosts := uniqHosts(locations) // <- the only usage
```
I think it's strange because we use backup location access info just for getting the hosts of the target cluster? In general, getting the node info of target cluster nodes does not need any information regarding backup location.

pkg/service/one2onerestore/worker_validate.go
Michal-Leszczynski: Hmmm, in general I don't like the fact that we have a correct node info constructed in some way, which we should be safe to assume to be const, but then we modify it for the sake of later comparison. Wouldn't it be better to encapsulate the mapping logic in the comparison function itself? Some function like:
```go
func ValidateCompatibleTopology(srcClususter, dstCluster []nodeInfo, mapping map[node]node)
```
or perhaps the `[]nodeInfo` could be replaced with `map[HostID]nodeInfo` since we don't care for the order.

This is just a subjective suggestion, so that's fine if you want to keep the current `applyNodeMapping` approach, maybe just add some comment to it.

pkg/service/one2onerestore/worker_validate.go
```diff
@@ -0,0 +1,247 @@
...
+		if source.CPUCount != target.CPUCount {
+			return errors.Errorf("source CPUCount doesn't match target CPUCount")
+		}
+		if source.StorageSize > target.StorageSize {
+			return errors.Errorf("source StorageSize greater than target StorageSize")
+		}
+		if !slices.Equal(source.Tokens, target.Tokens) {
+			return errors.Errorf("source Tokens doesn't match target Tokens")
+		}
```
Michal-Leszczynski, line 93: I'm not sure if we have any guarantee that the tokens are always sorted.
I guess it wouldn't harm to sort them here for extra safety (unless that's ensured by the API or in some other place in the code).

pkg/service/one2onerestore/worker.go
```diff
@@ -0,0 +1,89 @@
...
+	nodeStatus, err := w.client.Status(ctx)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "nodes status")
+	}
+	var (
+		allManifests  []*backupspec.ManifestInfo
+		nodesCountSet = strset.New()
+	)
+	for _, location := range target.Location {
+		// Ignore location.DC because all mappings should be specified via nodes-mapping file
```
Michal-Leszczynski, line 38: That's a good direction, but then we shouldn't allow user to specify location DC in the first place.
Or, we can still treat it as a hint if it simplifies the implementation, although I'm not sure of that.

pkg/service/one2onerestore/worker.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		allManifests = append(allManifests, manifests...)
+
+		for _, n := range nodes {
+			nodesCountSet.Add(n.HostID)
+		}
+	}
+
+	nodesWithAccessCount := len(nodesCountSet.List())
```
Michal-Leszczynski, line 55: nit: you can use `nodesCountSet.Size()`.

pkg/service/one2onerestore/worker_validate.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	if err := checkOne2OneRestoreCompatiblity(sourceNodeInfo, targetNodeInfo, nodeMappings); err != nil {
+		return errors.Wrap(err, "clusters not equal")
+	}
+
+	return nil
+}
+
+func checkOne2OneRestoreCompatiblity(sourceNodeInfo, targetNodeInfo []nodeValidationInfo, nodeMappings []nodeMapping) error {
```
Michal-Leszczynski, line 37: typo: `checkOne2OneRestoreCompatiblity` -> `checkOne2OneRestoreCompatibility`

pkg/service/one2onerestore/worker_validate.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	return sourceCluster, targetCluster, nil
+}
+
+type nodeValidationInfo struct {
+	DC          string
+	Rack        string
+	HostID      string
+	CPUCount    int64
```
Michal-Leszczynski, line 142: Shouldn't it be shard count instead of cpu count?

Michal-Leszczynski (commented):
The validation process looks way cleaner now:)

pkg/service/one2onerestore/model.go
```diff
@@ -28,10 +28,17 @@ type nodeMapping struct {
 
 type node struct {
 	DC     string `json:"dc"`
-	Rack   string `json:"rack"`
+	Rack   string `json:"rack_id"`
```
karol-kokoszka, line 31: It's a rack name not ID. 
Example: https://github.com/scylladb/scylladb/blob/master/conf/cassandra-rackdc.properties

pkg/service/one2onerestore/worker.go
```diff
@@ -0,0 +1,89 @@
...
+type worker struct {
+	managerSession gocqlx.Session
+
+	client         *scyllaclient.Client
+	clusterSession gocqlx.Session
+
+	logger log.Logger
+}
+
+// getManifestsAndHosts checks that each host in target cluster should have an access to at least one target location and fetches manifest info.
```
karol-kokoszka, line 27: This comment is a bit confusing.
Each node must have the access to the location where its SSTables are stored.

1-1 restore provides mapping between source node and the target node.

Location defines the DC. If the DC is empty then it means (or should mean) that all DCs are here.
See https://manager.docs.scylladb.com/stable/backup#meta

The method must check that all nodes that are expected to restore particular DC have an access to the location which keeps SSTables of this DC.
The nodes-mapping for 1-1 restore defines which node is going to restore data from which DC.

pkg/service/one2onerestore/worker.go
karol-kokoszka: Guys, why not to revert the logic here ?
Instead of iterating over locations, iterate over target nodes to check if they are able to access the location keeping their expected SSTables ? It must have the access.

Actually, please ensure that each node have the access to the corresponding location.

pkg/service/one2onerestore/worker.go
```diff
@@ -0,0 +1,89 @@
...
+
+	nodesWithAccessCount := len(nodesCountSet.List())
+
+	// If manifest count != nodes with access count means that 1-1 restore is not possible.
+	if len(allManifests) != nodesWithAccessCount || len(allManifests) != len(nodeStatus) {
+		return nil, nil, fmt.Errorf("manifest count (%d) != target nodes (%d)", len(allManifests), len(nodesCountSet.List()))
+	}
+
+	return allManifests, nodesToHosts(nodeStatus), nil
+}
```
karol-kokoszka, line 63: This logic will be completely different when you iterate over nodes instead of locations.

After ensuring that node has the access to the location, you just download corresponding manifest to check details like numer of shards and token ring.

pkg/service/one2onerestore/worker_validate.go
karol-kokoszka: Agree, it should be shards.
CPU defines amount of available CPUs on the machine.
Shards are defining how many CPUs are used by Scylla.

pkg/service/one2onerestore/worker_validate_test.go
```diff
@@ -0,0 +1,288 @@
...
+		name string
+
+		nodeMappings []nodeMapping
+		targetHosts  []Host
+		expected     map[string]Host
+		expectedErr  bool
+	}{
+		{
+			name: "All hosts have mappings",
+			nodeMappings: []nodeMapping{
```
karol-kokoszka, line 23: These tests are OK.
But you should verify it bit broader way.
Please keep `validateClusters` as the method you test. Most likely it will require to use integration tests instead of this ones.

karol-kokoszka (commented):
Few comments.
One is about changing the logic - to iterate over target nodes instead of locations. This is the most important for me.

pkg/service/one2onerestore/worker.go
VAveryanov8: > After ensuring that node has the access to the location, you just download corresponding manifest to check details like numer of shards and token ring.

That's exactly what is happening inside validateCluster method. 

A few word in general about my implementation and what was the reasoning behind it: 
In a nutshell we have a source cluster (backup) which nodes are represented by manifests file in the backup location(s) and a target cluster which nodes are actual live nodes in the cluster we want to restore data.

1. Get source cluster nodes and target  cluster nodes.  (getManifestsAndHosts)
2. Collect information needed to compare sources and target cluster nodes. (collectNodeValidationInfo) (here I iterate over nodes) 
3. Compare them using node mapping as rule how to match source cluster node and target cluster node. (checkOne2OneRestoreCompatibility) 

If validation has passed successfully, then I can use nodes info from step 1 farther in the code, as I know - it's valid for 1-1-restore and each node has exact match. 
Keeping the logic this way give us ability to keep validation logic in one place, without spreading it all across other parts of the code. 

> This logic will be completely different when you iterate over nodes instead of locations.

Here is how I see this logic
1. find node dc by looking at nodes mappings
2. find corresponding location by checking location.DC with DC from step 1
3. Download manifest content. Two steps actually list dir and then download, because manifest path contains taskID which we don't know (or do we?)  
4. collect additional node info (token ring and etc) 
5. compare nodes info with manifest content.

Without getting into the details, the main difference between to solutions, is that in first we collect all the info and then do the validation, in second, we validate nodes one by one. 

For me it's more a less the same, but if you prefer second over first,  let me know and i'm gonna change this pr

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/go-chi/render v1.0.0
 	github.com/go-openapi/runtime v0.28.0
 	github.com/go-openapi/strfmt v0.23.0
-	github.com/gocql/gocql v1.6.0
+	github.com/gocql/gocql v1.7.0
```
Michal-Leszczynski, line 14: I see that this version bump keeps on sneaking into PRs.
I don't see a reason against it, but let's make it a separate PR, so that's easier to review.

pkg/service/one2onerestore/worker.go
Michal-Leszczynski: > Two steps actually list dir and then download, because manifest path contains taskID which we don't know (or do we?)

We don't know it from the start, but it can be established just once at the beginning (by listing manifests) and then reused, but we can't escape listing all manifest dirs as it's needed to verify that manifest count equals nodes count.

go.mod
VAveryanov8: good catch. I'm going to rebase this branch and version bump should disappear. 
P.S. it doesn't make sense to bump this dependence, because in go.mod it's replaced by `scylladb/gocql`

pkg/service/one2onerestore/worker.go
Michal-Leszczynski: In general I like the idea of creating per node worker as it provides more concurrent design.
In the context of the validation, such worker could be initialized with all basic information about backup and target nodes like:
- target and backup host ID
- target IP
- target and backup cluster ID
- DC
- backup task ID
- snapshot tag
- backup location

If we assume that we get all location <-> DC mapping by user, the only missing piece is the task ID, but it can be established just once by the main workflow and passed to worker initialization (main workflow would also need to validate manifest count, but it wouldn't need to read any manifest).

But even if we don't get such mapping, we can first calculate it, and then move to the per node worker.
The only concern is that it might be more difficult to support this approach if we would like to make nodes mapping optional for the sake of user's convenience. Then, we couldn't provide backup DC and host ID from the start, but they would need to be filled in some other way.

With this design, there are two benefits:
- easier parallelization of the validation workflow - all steps like reading the manifest, fetching target info, validating, etc. could sequential in the context of a single node, but we could just use `parallel.Run` on all of them without any additional effort. The current implementation also do the time consuming things in parallel, but it requires more code handling the parallelism.
- easier body functions - there would be no need of handling slices or maps of some validation info and thinking whether we can be sure that they are at sync at given point. Validation functions would work just on the per node context with a single manifest and a single target node info.

Michal-Leszczynski: But I also agree that the current implementation does it's job, so this is more about personal preferences, so if the things from the previous comment didn't convince you, I'm not against the current state of implementation.

pkg/service/one2onerestore/worker.go
karol-kokoszka: The 1-1 restore, includes `nodes-mapping`. It explicitly tells which target node is expected to download what set of SSTables. The goal is to find which location keeps SSTables of the source-node and ensure that the target node can access this location.

The current comment says: **each host in target cluster should have an access to at least one target location**
This is not true, we must ensure that each target host have an access to the exact location that stores SSTables for the mapped source node.

> Do you mean that we should require user to specify the full location <-> DC mapping with the --location flag and not calculate it on our? I was under different impression when writing the https://github.com/scylladb/scylla-manager/pull/4230#discussion_r1937014861 about ignoring location DC (resolved).

We ARE requiring user to specifiy `nodes-mapping`. Node is part of on of the DC. Scylla Manager must understand which location (from the given locations) keeps the DC that is owning the source node.
To understand the `full location` <-> `DC mapping` , or even rather `location` <-> `nodes mapping` it's enough to check `meta` directory https://manager.docs.scylladb.com/stable/backup#meta from all bacup locations.

> I think that making 1-1-restore as easy to use is important in general, but it does not need to happen in the first iteration targeting mainly Scylla Cloud as a user.

> On the other hand, the code matching locations to DCs on SM side is already a part of this PR, so does it make sense to delete it? What do you guys think?

We don't care about the ease of use for end user much. There is still automation required to deliver cloned cluster. This automation creates the output which must involve the mapping <source-node> -> <target-node>. 
We care about the compatibility with the automation tool.

karol-kokoszka: OK, this method is just expected to get list of hosts from the target cluster and the list of manifests from the snapshot.

It adds simple validation to compare if the number of target hosts is equal to number of manifests.

It's fine.

But still... I think the comment is misleading.
getManifestsAndHosts -> getAllSnapshotManifestsAndAllTargetHosts would be more meaningful for me.

pkg/service/one2onerestore/worker.go
karol-kokoszka: It's not the validation stage. My fauly. It's fine.

pkg/service/one2onerestore/worker.go
karol-kokoszka: @VAveryanov8 are you checking somewhere in the code if the target node has the access to the expected location ? I mean the location that keeps SSTables that are expected to be downloaded to this node ?
It should be part of validation stage.
Otherwise, it may happen that you go to the copy stage, but one of the target nodes is not able to get its SSTables.

karol-kokoszka: > In general I like the idea of creating per node worker as it provides more concurrent design.
In the context of the validation, such worker could be initialized with all basic information about backup and target nodes like:

> target and backup host ID
target IP
target and backup cluster ID
DC
backup task ID
snapshot tag
backup location
If we assume that we get all location <-> DC mapping by user, the only missing piece is the task ID, but it can be established just once by the main workflow and passed to worker initialization (main workflow would also need to validate manifest count, but it wouldn't need to read any manifest).

> But even if we don't get such mapping, we can first calculate it, and then move to the per node worker.
The only concern is that it might be more difficult to support this approach if we would like to make nodes mapping optional for the sake of user's convenience. Then, we couldn't provide backup DC and host ID from the start, but they would need to be filled in some other way.

> With this design, there are two benefits:

> easier parallelization of the validation workflow - all steps like reading the manifest, fetching target info, validating, etc. could sequential in the context of a single node, but we could just use parallel.Run on all of them without any additional effort. The current implementation also do the time consuming things in parallel, but it requires more code handling the parallelism.
easier body functions - there would be no need of handling slices or maps of some validation info and thinking whether we can be sure that they are at sync at given point. Validation functions would work just on the per node context with a single manifest and a single target node info.
EDIT: Note that such design would probably fit very well into the later restore data stages as well.

That's the idea of 1-1 restore to do it per node in parallel. After all SSTables are copied to the corresponding node, then it's a matter of calling to refresh per all tables that are included in the backup location in this snapshot for the current node.
It should/must be done in parallel per node.

@VAveryanov8 @Michal-Leszczynski let's sync on monday. It may be faster way of discussing it.

I put in the design doc -> `There must be one worker / goroutine per destination node copying and refreshing SSTables from the corresponding node.`

karol-kokoszka:
These are the only concerns I have right now:
- https://github.com/scylladb/scylla-manager/pull/4230#discussion_r1946518741
- https://github.com/scylladb/scylla-manager/pull/4230#discussion_r1946508109

Other than that, it's fine.

VAveryanov8:
@Michal-Leszczynski @karol-kokoszka I've made changes to validation stage according to your comments, so please take a look one more time - it should be easier now :) 

additionally updated unit and integrational tests.

pkg/service/one2onerestore/model.go
Michal-Leszczynski: But now it shows that this PR is actually changing it from `rack_id` to `rack` (instead of not touching it at all).
I guess that's because of having multiple PRs rebased on one another.

VAveryanov8: It only shows changes in `pkg/command/one2onerestore`, because i've renamed `rack_id` to `rack` in the scope of this PR

pkg/service/one2onerestore/worker.go
```diff
@@ -0,0 +1,81 @@
...
+		manifests, err := w.getManifestInfo(ctx, nodeAddr, target.SnapshotTag, target.SourceClusterID, location)
+		if err != nil {
+			return nil, nil, err
+		}
+		allManifests = append(allManifests, manifests...)
+	}
+	return allManifests, nodesToHosts(nodeStatus), nil
+}
+
+func findCorrespondingNode(nodeStatus scyllaclient.NodeStatusInfoSlice, locationDC string, nodeMappings []nodeMapping) (addr string, err error) {
```
Michal-Leszczynski, line 45: `findCorrespondingNode` is rather vague. Corresponding to what?

pkg/service/one2onerestore/worker_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+package one2onerestore
+
+import (
+	"testing"
+
+	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
+)
+
+func TestFindCorresponingNode(t *testing.T) {
```
Michal-Leszczynski, line 11: `TestFindCorresponingNode` -> `TestFindCorrespondingNode`

pkg/service/one2onerestore/worker_validate_integration_test.go
```diff
@@ -0,0 +1,341 @@
...
+				t.Fatalf("Expected err, but got nil")
+			}
+			if !tc.expecterErr && err != nil {
+				t.Fatalf("Unexpected err: %v", err)
+			}
+		})
+	}
+}
+
+func testNewWorker(t *testing.T) (*worker, *testutils.HackableRoundTripper) {
```
Michal-Leszczynski, line 235: Shouldn't it be `newTestWorker`?

pkg/service/one2onerestore/worker_validate_integration_test.go
```diff
@@ -0,0 +1,341 @@
...
+
+	nodeMappings := getNodeMappings(t, w.client)
+
+	testCases := []struct {
+		name                 string
+		hostsProvider        func() []Host
+		manifestsProvider    func() []*backupspec.ManifestInfo
+		nodeMappingsProvider func() []nodeMapping
+		setIntereptor        func()
+		expecterErr          bool
```
Michal-Leszczynski, line 64: typos

pkg/service/one2onerestore/worker_validate_integration_test.go
```diff
@@ -0,0 +1,341 @@
...
+				t.Fatalf("Unexpected err: %v", err)
+			}
+		})
+	}
+}
+
+func testNewWorker(t *testing.T) (*worker, *testutils.HackableRoundTripper) {
+	t.Helper()
+	hrt := testutils.NewHackableRoundTripper(scyllaclient.DefaultTransport())
+	cfg := scyllaclient.TestConfig(testconfig.ManagedSecondClusterHosts(), testutils.AgentAuthToken())
```
Michal-Leszczynski, line 238: Why second cluster? Wouldn't it be more interesting to test it on a multi-dc setup?

pkg/service/one2onerestore/worker_validate_integration_test.go
```diff
@@ -0,0 +1,341 @@
...
+	return backup.Config{
+		DiskSpaceFreeMinPercent:   5,
+		LongPollingTimeoutSeconds: 1,
+		AgeMax:                    24 * time.Hour,
+	}
+}
+
+func runBackup(t *testing.T, backupSvc *backup.Service, clusterID uuid.UUID, props map[string]any) string {
+	t.Helper()
+	testutils.Printf("Run backup with properties: %v", props)
```
Michal-Leszczynski, line 291: Does `testutils.Printf` have any advantages over regular `t.Logf`?

pkg/service/one2onerestore/worker_validate_integration_test.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	pr, err := backupSvc.GetProgress(ctx, clusterID, taskID, runID)
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "get progress"))
+	}
+	return pr.SnapshotTag
+}
+
+func getNodeMappings(t *testing.T, client *scyllaclient.Client) []nodeMapping {
```
Michal-Leszczynski, line 318: It wold be good to signalize (either in a comment or function name) that this gets the identity mapping.

pkg/service/one2onerestore/worker_validate_integration_test.go
```diff
@@ -0,0 +1,341 @@
...
+			},
+			expecterErr: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.setIntereptor != nil {
+				tc.setIntereptor()
+			}
```
Michal-Leszczynski, line 222: We should always set interceptor to nil if it's not specified in the test case.
Otherwise, test case ordering could influence test results.
It would also allow for not setting nil interceptors directly in the test cases.

Michal-Leszczynski (commented):
Just some small nits!

pkg/service/one2onerestore/service.go
```diff
@@ -25,7 +25,7 @@ type Servicer interface {
 	Runner() Runner
 }
 
-// Service for the 1-1-restore.
+// Service for the 1-2-1 restore.
```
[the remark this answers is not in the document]
VAveryanov8: thanks, I haven't noticed it - I think it happened during rebase :)

pkg/service/one2onerestore/worker_validate_integration_test.go
VAveryanov8: yep, make sense, updated to multi-dc cluster

pkg/service/one2onerestore/worker_validate_integration_test.go
VAveryanov8: I don't think there is any advantage, just saw the usage of it in other integration tests, so decided to reuse for consistency.

pkg/service/one2onerestore/worker_validate_integration_test.go
Michal-Leszczynski: `setIntereptor` -> `setInterceptor`

Michal-Leszczynski (commented):
Just one last typo!!!
https://github.com/scylladb/scylla-manager/pull/4230#discussion_r1951220905

VAveryanov8:
> there is a mismatch in the token ring shape of the node w.Client.Tokens(ctx, h.IP)

done: https://github.com/scylladb/scylla-manager/blob/b7ceb0989e8d19eb5bceefa8aca42d4e0534ba6f/pkg/service/one2onerestore/worker_validate_integration_test.go#L204 


> there is a mismatch between number of DCs

hm, interesting case, but everything else is correct, e.g. number of nodes, each node is present in the node mappings and token ring is matching? This is not validated by current implementation. 

If we need such validation then it could be check just by looking at the nodes mapping, this could be done on the sctool side and such task shouldn't be scheduled.

Michal-Leszczynski (approved):

VAveryanov8:
> We can have some basic validation on sctool side, but it's always better to have the whole validation on the SM side as it's possible to schedule 1-1-restore directly via API - which might even be the case for the siren automation.

That's a good point, then I'm going to extend this pr a little bit :)

VAveryanov8:
@karol-kokoszka @Michal-Leszczynski I've pushed one more commit to this pr with nodes mapping validation - https://github.com/scylladb/scylla-manager/pull/4230/commits/45c479b133378330923c303fcf35566850ac5a5d.

pkg/service/one2onerestore/model.go
```diff
@@ -49,9 +49,41 @@ func (t *Target) validateProperties() error {
...
+	)
 
+	for _, nodeMapping := range nodesMapping {
+		s, t := nodeMapping.Source, nodeMapping.Target
+
+		sourceDCCount[s.DC]++
+		targetDCCount[t.DC]++
+
+		sourceRackCount[s.DC+s.Rack]++
+		targetRackCount[t.DC+t.Rack]++
```
Michal-Leszczynski, line 78: Let's add some simple struct like:
```go
type dcRack struct {
  dc string
  rack string
}
```
instead of concatenating strings, as it's always better to be on the safer side.

pkg/service/one2onerestore/model.go
```diff
@@ -49,9 +49,41 @@ func (t *Target) validateProperties() error {
...
+
+		if sourceDCCount[s.DC] != targetDCCount[t.DC] {
+			return errors.Errorf("source and target clusters has different number of nodes per DC")
+		}
+
+		if sourceRackCount[s.DC+s.Rack] != targetRackCount[t.DC+t.Rack] {
+			return errors.Errorf("source and target clusters has different number of nodes per Rack")
+		}
```
Michal-Leszczynski, line 86: I would say that what we are trying to validate here is that mapping between src and dst DCs is 1-1.
So it's not just about having the same number of nodes per DC, e.g.:
- dc1:nodeX -> dc3:nodeX
- dc1:nodeX -> dc4:nodeX
- dc2:nodeX -> dc4:nodeX

where nodeX is some new unique node each time. Here we have the same number of nodes per DC (at least in my understanding), but we don't have the 1-1 mapping, which is important.

The validation code should work, because it checks the DC size after each iteration (although it isn't obvious, because the code also kind of focuses on the DC size, not the 1-1 mapping), but the error message should focus on the 1-1 mapping, not just the DC size.

The same goes for rack error message.

pkg/service/one2onerestore/model.go
Michal-Leszczynski: Also, even though we probably do it explicitly or implicitly in other places in the code, we should validate as much contextless properties of node mapping in `validateNodesMapping` as possible. This means that we should also validate the 1-1 nodes mapping here as well (not just the DCs and racks).

pkg/service/one2onerestore/model.go
karol-kokoszka: Indeed, @VAveryanov8 you can extend your test with 
```
		{
			name: "test",
			nodesMapping: []nodeMapping{
				{Source: node{DC: "dc1", Rack: "rack1", HostID: "h1"}, Target: node{DC: "dc2", Rack: "rack1", HostID: "host1"}},
				{Source: node{DC: "dc1", Rack: "rack2", HostID: "h2"}, Target: node{DC: "dc2", Rack: "rack2", HostID: "host3"}},
				{Source: node{DC: "dc1", Rack: "rack2", HostID: "h3"}, Target: node{DC: "dc2", Rack: "rack1", HostID: "host2"}},
			},
			expectedErr: true,
		},
```

node from source_rack1 that has only 1 node is mapped to host from target rack that has 2 nodes.
It passes current validation, but shouldn't.

pkg/service/one2onerestore/model.go
VAveryanov8: updated!

pkg/service/one2onerestore/model.go
```diff
@@ -55,35 +55,72 @@ func (t *Target) validateProperties() error {
...
+	return nil
+}
+
+func checkHostMapping(hostMap map[string]struct{}, hostID string) error {
+	if _, ok := hostMap[hostID]; !ok {
+		hostMap[hostID] = struct{}{}
+		return nil
+	}
+	return errors.Errorf("host is already mapped: %s", hostID)
+}
```
Michal-Leszczynski, line 126: This comment is a nit and as we discussed, feel free to resolve it on your own.

nit: using generic function like:
```go
func checkOneTwoOneMapping[K ordered](m map[K]K, source, target K, allowMultiple bool) error {
	mapped, ok := m[source]
	if !ok {
		m[source] = target
		return nil
	}
	if mapped != target {
		return errors.Errorf("%v is mapped to both %v and %v", source, target, mapped)
	}
        if !allowMultiple {
                return errors.Errof("%v is mapped to %v multiple times", source, target)
        }
	return nil
}
```

The benefit of this approach is that we can have just one checking function, we don't need to expose `dcRack` type, and we avoid checking 1-1 dc mapping with `dcRack` type, which is a little hacky.
On the downside, the error display probably look clunkier, but I haven't tried it.

Michal-Leszczynski (approved):

Issue #4203, opened 2025-01-13
Add refresh SSTables stage


karol-kokoszka:
After the corresponding SSTables are copied to the destination node, there is a need to call Scylla server to refresh given keyspace and column_family.

https://github.com/scylladb/scylla-manager/blob/b501137e8681f6ee8f29bbeadf7558011c585543/v3/swagger/scylla_v1.json#L11841

This must be done per node.
In the context of node, per keyspace and per column_family.

Issue #4202, opened 2025-01-13
Add copy-data stage


karol-kokoszka:
After validation, Scylla Manager must copy SSTables of the source node to the corresponding /keyspace/column_family/upload data directory on the destination node.

If the corresponding /keyspace/column_family/upload data directory doesn't exist, it means that schema is not restored.
Service must stop the restore with the meaningful error message.

To download files from the backup location, SM is using RClone.
It's OK to copy all the data from the source_node node/keyspace/column_family location with one call and then monitor the rclone job progress.

PR #4235, opened 2025-01-31
feat(1-1-restore): adds copy data stage + loadSStables call

VAveryanov8:
This adds copy data stage for the 1-1-restore with simple integration test (create backup, truncate tables, restore).
Data is copied by each node, table by table and after copy data dir job loadSStables is called.

Refs: #4202 #4203 

P.S. For now I've added 1-1-restore integration test under  `Test other, smaller packages` check as it small enough (run takes about 30s) and doesn't cover a lot of use cases.

Also this pr includes full support for `--keyspace` parameter (meaning you can filter tabels, use `!` and etc), but it will be changed in separate PR accordingly to #4253 

---

Please make sure that:

### DIFF
--- a/.github/cfg/integration-test-core.yaml
+++ b/.github/cfg/integration-test-core.yaml
@@ -136,4 +136,4 @@ jobs:
         run: make pkg-integration-test PKG=./pkg/schema/migrate
 
       - name: Run 1-1-restore tests
-        run: make pkg-integration-test PKG=./pkg/service/one2onerestore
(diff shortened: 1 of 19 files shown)


karol-kokoszka:
I'll proceed with the review, when the previous parts are merged to feature-branch.

VAveryanov8:
@karol-kokoszka @Michal-Leszczynski previous parts are in the feature branch, so this one is ready for review!

PR #4238, opened 2025-02-03
Set Scylla user task TTL when using native Scylla API

### LINKED ISSUE #4158: Set Scylla task TTL
The information about Scylla tasks (e.g. backup and restore) is stored on Scylla node for a limited amount of time after they are finished.
Querying status of a task which finished over TTL seconds ago results in an error.

For backup:
It's kind of fine if we loose a successful task progress, as we can always start a new one and since [SSTables are being removed from the snapshot dir as they are being uploaded](https://github.com/scylladb/scylladb/issues/21799), we don't perform any redundant work (although we need to worry about progress calculation).

For restore:
We are unlucky if we loose a successful task progress, as we would basically need to re-restore the whole batch from scratch.

The current Scylla implementation is that the internal tasks and [user tasks have different TTL values](https://github.com/scylladb/scylladb/pull/21505). Note that setting TTL value means that all tasks started after the change will use the set value, but it's not possible to set per task TTL value directly.

Tasks scheduled by SM are qualified as regular user tasks - the default value for their TTL is 1h.
SM should always set the TTL value even when it's 1h - we can't be sure that the value hasn't been changed from the default.

The question are:
- what value should we set - is 1h enough?
- should backup and restore use different values?
- should those values be configurable?

cc: @karol-kokoszka @VAveryanov8 @regevran

Michal-Leszczynski:
Fixes #4158

### DIFF
--- a/pkg/scyllaclient/client_scylla.go
+++ b/pkg/scyllaclient/client_scylla.go
@@ -1263,6 +1263,57 @@ func (c *Client) ScyllaAbortTask(ctx context.Context, host, id string) error {
 	return err
 }
 
+// ManagerTaskTTLSeconds defines the Scylla user task TTL
+// that SM should use for the Scylla tasks it schedules.
+// We do it in order to minimize the chance that Scylla task finished
+// right after we timed out on long polling its progress.
+// (5 minutes is generous for such use case).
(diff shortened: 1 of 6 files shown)


VAveryanov8 (approved):
LGTM

Issue #4242, opened 2025-02-06
Move to UBI based container images for Scylla Manager containers


ylebi:
To achieve OpenShift certification for the Scylla-Operator, our Scylla Manager container images need to be based on Red Hat's UBI images.

cc: @mykaul @mflendrich

mflendrich:
The scope of this issue is that the Scylla Manager container passes all the criteria in sections 2.1 (image content requirements) and 2.2 (image policy requirements) [the certification policy guide](https://docs.redhat.com/en/documentation/red_hat_software_certification/2024/html-single/red_hat_openshift_software_certification_policy_guide/index#assembly-requirements-for-container-images_openshift-sw-cert-policy-introduction).

In human terms, this means:
- moving to an UBI-based image
- adding the image labels in the dockerfile as described in the policy
- verifying using the preflight tool as described in https://github.com/scylladb/scylla-pkg/issues/4858#issuecomment-2621040083

mflendrich:
I think this shouldn't be closed yet, because the dockerfile after #4244 is still missing some stuff required for certification (at least the labels).

cc @karol-kokoszka @Michal-Leszczynski @yaronkaikov

avikivity:
Need to check if there are any restrictions on those UBI things.

mflendrich:
OpenShift certification tested with the latest dev: `scylladb/scylla-manager:3.5.0-dev-0.20250330.da2730904`.

The `BasedOnUbi` test passes.
`HasRequiredLabel` and `HasLicense` not yet.

```
time="2025-03-31T08:57:05+02:00" level=info msg="certification library version" version="1.12.1 <commit: 1adc5822eb992c19c97ce16b83b0158406c4ada0>"
time="2025-03-31T08:57:06+02:00" level=info msg="running checks for scylladb/scylla-manager:3.5.0-dev-0.20250330.da2730904 for platform arm64"
time="2025-03-31T08:57:06+02:00" level=info msg="target image" image="scylladb/scylla-manager:3.5.0-dev-0.20250330.da2730904"
time="2025-03-31T08:57:38+02:00" level=info msg="check completed" check=HasLicense result=FAILED
time="2025-03-31T08:57:38+02:00" level=info msg="check completed" check=HasUniqueTag result=PASSED
time="2025-03-31T08:57:38+02:00" level=info msg="check completed" check=LayerCountAcceptable result=PASSED
[...]

PR #4244, opened 2025-02-06
feat(manager docker) switch to be based on ubi9-minimal

yaronkaikov:
To achieve OpenShift certification for the Scylla-Operator, our Scylla Manager container images must be based on Red Hat's UBI images.

Fixes: https://github.com/scylladb/scylla-manager/issues/4242

### DIFF
--- a/dist/docker/scylla-manager-agent.dockerfile
+++ b/dist/docker/scylla-manager-agent.dockerfile
@@ -1,15 +1,13 @@
-FROM ubuntu
-ARG ARCH=amd64
+FROM docker.io/redhat/ubi9-minimal:latest
+ARG ARCH=x86_64
 
-RUN apt-get update && \
-    apt-get -y upgrade && \
-    apt-get install -y --no-install-recommends ca-certificates && \
-    apt-get autoremove -y && \
-    apt-get clean && \
+RUN microdnf update && \
+    microdnf -y upgrade && \
+    microdnf install -y ca-certificates && \
     rm -rf /var/lib/apt/lists/*
 
-COPY release/scylla-manager-agent*$ARCH.deb /
-RUN dpkg -i scylla-manager-agent*$ARCH.deb && rm /scylla-manager-agent*.deb
+COPY release/scylla-manager-agent*$ARCH.rpm /
+RUN rpm -ivh scylla-manager-agent*$ARCH.rpm && rm /scylla-manager-agent*.rpm
 
 USER scylla-manager
 ENV HOME=/var/lib/scylla-manager/
--- a/dist/.goreleaser-docker.yaml
+++ b/dist/.goreleaser-docker.yaml
@@ -29,8 +29,8 @@ dockers:
       - docker/scylla-manager.yaml
       - release/
     build_flag_templates:
-      - "--build-arg=ARCH=arm64"
-      - "--platform=linux/arm64"
+      - "--build-arg=ARCH=aarch64"
+      - "--platform=linux/aarch64"
 
   - ids:
     use: docker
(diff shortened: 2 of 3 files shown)


karol-kokoszka (approved):

Michal-Leszczynski:
@yaronkaikov does this PR fix #4242? I'm asking in the context of this [comment](https://github.com/scylladb/scylla-manager/issues/4242#issuecomment-2640136745).

mflendrich:
This PR isn't enough to close #4242 - the preflight check won't pass, at least the labels are missing.

yaronkaikov:
> @yaronkaikov does this PR fix #4242? I'm asking in the context of this [comment](https://github.com/scylladb/scylla-manager/issues/4242#issuecomment-2640136745).

probably not, but we can send a follow-up patches to make it right. Just not sure when

mflendrich:
follow-up sounds good :+1:. Good to see manager on UBI, however the entire list (at least the labels) is a blocking requirement for openshift certification as well. So we'll need to get it scheduled somehow.

@yaronkaikov could you please reopen #4242 so that we don't lose track of the tail end?
I don't have reopen rights on the repo. Thanks :pray:

yaronkaikov:
I re-opened it

dist/docker/scylla-manager-agent.dockerfile
```diff
@@ -1,15 +1,13 @@
-FROM ubuntu
-ARG ARCH=amd64
+FROM docker.io/redhat/ubi9-minimal:latest
```
mykaul, line 1: Should have used quay.io really.

It's unclear why we use latest, but we can review this decision later.

dist/docker/scylla-manager-agent.dockerfile
```diff
@@ -1,15 +1,13 @@
...
 
-RUN apt-get update && \
-    apt-get -y upgrade && \
-    apt-get install -y --no-install-recommends ca-certificates && \
-    apt-get autoremove -y && \
-    apt-get clean && \
+RUN microdnf update && \
+    microdnf -y upgrade && \
+    microdnf install -y ca-certificates && \
     rm -rf /var/lib/apt/lists/*
```
mykaul, line 7: Is this the way to clean microdnf residue?

dist/docker/scylla-manager-agent.dockerfile
```diff
@@ -1,15 +1,13 @@
...
 
-COPY release/scylla-manager-agent*$ARCH.deb /
-RUN dpkg -i scylla-manager-agent*$ARCH.deb && rm /scylla-manager-agent*.deb
+COPY release/scylla-manager-agent*$ARCH.rpm /
```
mykaul, line 9: This creates another layer, which I assume we don't squash later?

mykaul:
@yaronkaikov apologies for not reviewing this, but I have left some comments. I believe we need a follow-up to this work - shall I open another issue?

dist/docker/scylla-manager-agent.dockerfile
mykaul: In fact, we should use registry.access.redhat.com/ubi9/ubi:latest

yaronkaikov:
> @yaronkaikov apologies for not reviewing this, but I have left some comments. I believe we need a follow-up to this work - shall I open another issue?

I have re-opened the original issue since we are missing some labels for the certification, please add relevant info there, or open a new one.

dist/docker/scylla-manager-agent.dockerfile
mykaul: This was not addressed. Please do.

dist/docker/scylla-manager-agent.dockerfile
mykaul: ?

ylebi: https://github.com/scylladb/scylla-operator-images/blob/master/base/ubi/9.5-minimal/Dockerfile#L8

PR #4245, opened 2025-02-06
goreleaser: ignore backupspec/* tags

karol-kokoszka:
As per title name. 

Tag for submodule "backupspec" is added what creates the problem with goreleaser in one of the jenkins pipeline.
This PR is to just ignore this tag by goreleaser.

---

Please make sure that:


karol-kokoszka:
Build problem is fixed with this change https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-build/917

Issue #4253, opened 2025-02-13
Add "--keyspaces" flag to onetoonerestore


karol-kokoszka:
Introduce (but NOT copy) the --keyspace flag to onetoonerestore https://manager.docs.scylladb.com/stable/sctool#k-keyspace
The keyspace flag will work only on keyspace level. Ex "ks1,ks2".

PR #4256, opened 2025-02-17
feat(1-1-restore): reduces --keyspace param features to keyspace level filtering only

VAveryanov8:
This reduces functionality of `--keyspace` parameter to support only:
 - keyspace level filtering. Table level filtering (system.table_name) is forbidden.
 - only include filtering. Excluding(!) is forbidden.
 - wildcards (\*) are forbidden. The only exception is when the only filter that is provided is a single character (*), meaning that all keyspaces should be included into restore process - it's a default value if `--keyspace` is not provided.

Refs: #4253

---

Please make sure that:

### DIFF
--- a/pkg/service/one2onerestore/model.go
+++ b/pkg/service/one2onerestore/model.go
@@ -3,6 +3,8 @@
 package one2onerestore
 
 import (
+	"strings"
+
 	"github.com/pkg/errors"
 	. "github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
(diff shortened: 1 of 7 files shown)


pkg/service/one2onerestore/model.go
```diff
@@ -51,12 +53,36 @@ func (t *Target) validateProperties() error {
...
+// Also exclude operations (!) are forbidden as well.
+func validateKeyspaceFilter(keyspaces []string) error {
+	// default value, it's ok to have a wildcard(*) in that case.
+	if len(keyspaces) == 1 && keyspaces[0] == "*" {
+		return nil
+	}
+	for _, filter := range keyspaces {
+		if strings.Contains(filter, ".") {
+			return errors.Errorf("only keyspace level filtering is allowed, but table is provided: %s", filter)
+		}
```
Michal-Leszczynski, line 75: Unfortunately, alternator tables can have dots in their names ([dynamodb docs](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html)).

On the other hand, this is the reason why SM can't restore their schema, as CQL does not allow for it.
But schema can be recreated manually by the user/support and the data restoration should work fine for alternator tables as well.

This means that dots handling should take alternator tables into consideration.

This could mean that it would be easier to support full `--keyspace` flag syntax, but just verify that its usage does not lead to partial keyspace restoration.

pkg/service/one2onerestore/model_test.go
```diff
@@ -121,3 +121,60 @@ func TestValidateNodesMapping(t *testing.T) {
...
+func TestValidateKeyspaceFilter(t *testing.T) {
+	testCases := []struct {
+		name        string
+		keyspace    []string
+		expectedErr string
+	}{
+		{
+			name:     "Only keyspace level filtering",
+			keyspace: []string{"hello", "system"},
+		},
```
Michal-Leszczynski, line 134: We need to add a test for alternator tables (e.g. `Tab_le-With1.da_sh2-aNd.d33ot.-`).

pkg/service/one2onerestore/model.go
VAveryanov8: good point! 
we don't care about dots in the table names, because that's the idea - allow only keyspace level filter, but as far as I understand from the Alternator docs - alternator keyspace can have dots too, because it's basically `alternator_${table_name}` - that's the problem :) 

The simple fix which will allow only filtering on keyspace level and not relying on dots would be getting all keyspaces names from the cluster and comparing them with what was provided in `--keyspace` flag. This way we will not allow to filter on anything but only keyspaces that are part of the cluster (which is ok, because schema should already be restored).  What do you think about this fix?

Michal-Leszczynski: Sounds good!

VAveryanov8:
jfyi: I've force-pushed rebase on `1-1-restore-feature-branch`

Michal-Leszczynski (approved):

Issue #4121, opened 2024-11-22
Re-structure SM modules


Michal-Leszczynski:
Right now we have 4 modules in our SM repo:
- github.com/scylladb/scylla-manager/v3 - core SM module
- github.com/scylladb/scylla-manager/v3/pkg/managerclient - wrapper around SM client generated from swagger and used by sctool
- github.com/scylladb/scylla-manager/v3/pkg/util - util functions used by both managerclient and core SM modules
- github.com/scylladb/scylla-manager/v3/swagger - swagger definitions for Scylla, SM, SM-agnet and generated client

This module division is annoying to work with. For example, when extending task progress display in sctool, one must:
- make PR with swagger change
- make PR updating deps in managerclient module
- make PR with managerclient changes
- make PR updating deps in SM core module
- make PR with core SM changes

Which creates a lot of steps and is prone to conflicts when multiple people are working on SM sub-modules.
Not to mention that forgetting to update dependencies might be easy to overlook and it could cause problems with releases.

Except for this module structure being annoying to work with, it also does not make much sense logically:
- why would we export managerclient module to siren/operator if it's just a simple wrapper of swagger generated files
- there is no need for separate util module
- swagger module contains mixed swaggers definitions from different projects - Scylla swagger is completely external to SM, SM swagger is internal to SM, but it should be available for other projects (siren/operator), Agent swagger is completely internal to SM

Proposed solution:
1. Change swagger module to contain just SM swagger definitions and generated code
2. Merge SM-agent swagger into SM core module
3. Switch from maintaining Scylla swagger in our repo to importing it from [Scylla golang swagger repo](https://github.com/scylladb/scylladb-swagger-go-client) (created and used by the operator team)
4. Merge managerclient module into SM core module - siren/cloud should be fine with that, as they just need the SM swagger generated files, which can be found in the SM swagger module - although it would require some small changes on their side (perhaps we can do it for them).
5. Merge util module into SM core module

This would leave us with just 2 modules:
- github.com/scylladb/scylla-manager/v3 - core SM module
- github.com/scylladb/scylla-manager/v3/swagger - swagger definitions and generated files for SM swagger

Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 I made some changes to the approach discussed during our meeting, but that's because I learned a little bit more about how swagger definitions are managed by the operator team.
Please let me know if you are fine with this approach!

VAveryanov8:
[...]

PR #4260, opened 2025-02-19
feat(1-1-restore): sets tombstone_gc mode to repair 

VAveryanov8:
This sets tombstone_gc mode to 'repair' to avoid data resurrection.
If tables mode is already 'repair', 'disabled' or 'immediate', then it won't be changed.

I'm not 100% sure if it's a safe to leave 'immediate' unchanged, but from the docs seems ok. 

P.S.
Also this adds skipping of local system tables - the same way we have in
regular restore service.

P.S.2 : )
I think it may turn out that we need to set tombstone gc mode to 'repair' for views as well. I've created https://github.com/scylladb/scylla-manager/issues/4261 to investigate this.

---

Please make sure that:

### DIFF
--- a/pkg/service/one2onerestore/worker_tgc.go
+++ b/pkg/service/one2onerestore/worker_tgc.go
@@ -9,12 +9,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/scylladb/go-set/strset"
-	"github.com/scylladb/scylla-manager/v3/pkg/service/backup/backupspec"
-	"github.com/scylladb/scylla-manager/v3/pkg/util/retry"
-	"github.com/scylladb/scylla-manager/v3/pkg/util/timeutc"
 
-	"github.com/scylladb/gocqlx/v2"
 	"github.com/scylladb/gocqlx/v2/qb"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/service/one2onerestore/worker_tgc.go
```diff
@@ -0,0 +1,181 @@
...
+	type backupTable struct {
+		keyspace, table string
+	}
+
+	tablesToRestore := map[backupTable]struct{}{}
+
+	for _, wl := range workload {
+		err := wl.manifestContent.ForEachIndexIter(keyspaceFilter, func(fm backupspec.FilesMeta) {
+			tablesToRestore[backupTable{keyspace: fm.Keyspace, table: fm.Table}] = struct{}{}
+		})
```
Michal-Leszczynski, line 44: We iterate over index (which requires reading from a file) just to get table names more than once.
Can't we extend worker with restored tables initialized just once?

pkg/service/one2onerestore/worker_tgc.go
```diff
@@ -0,0 +1,181 @@
...
+			"wait", wait,
+		)
+	}
+
+	return alterSchemaRetryWrapper(ctx, op, notify)
+}
+
+func alterTableTombstoneGCStmt(keyspace, table string, mode tombstoneGCMode) string {
+	return fmt.Sprintf(`ALTER TABLE %q.%q WITH tombstone_gc = {'mode': '%s'}`, keyspace, table, mode)
+}
```
Michal-Leszczynski, line 123: Could we move get/set tombstone gc functions to query pkg?

pkg/service/one2onerestore/worker_tgc.go
```diff
@@ -0,0 +1,181 @@
...
+
+func (w *worker) awaitSchemaAgreement(ctx context.Context, clusterSession gocqlx.Session) {
```
Michal-Leszczynski, line 125: I don't think that it's needed here. This is copied from the regular restore, which was copied from the backup, where it made sense to await schema agreement before backing up schema sstables. Here we are even using some additional wrappers over schema altering methods in case they take more time because of many recent schema changes.

Michal-Leszczynski: Not to mention that this implementation probably requires single host session to work properly, as the two queries might be routed to different hosts.

Michal-Leszczynski (commented):
Looks good, just a few comments!

pkg/service/one2onerestore/worker_tgc.go
VAveryanov8: As query pkg lives under separate go module, it would be easier to make this change as part of (or after ) https://github.com/scylladb/scylla-manager/issues/4121.

pkg/service/one2onerestore/worker_tgc.go
VAveryanov8: I loaded everything into a memory and now thinking how big could be all []FileMeta.[]Files? Have you every had any issues with it? I guess, ManifestContent uses files under the hood for a reason? What do you think?

Michal-Leszczynski (approved):
Looks good!

PR #4266, opened 2025-02-25
Refactor repair tests

Michal-Leszczynski:
This is a ground work for introducing new tablet repair API into repair service tests.
It mainly:
- separates interceptor request parsing from the test code, so that it's easier to extend it with new tablet repair endpoints
- tries not to rely on tablet table ranges count, as the new tablet repair API will repair an entire table with a single call
- improves vnode/tablet keyspace creation

### DIFF
--- a/pkg/service/repair/helper_integration_test.go
+++ b/pkg/service/repair/helper_integration_test.go
@@ -1,0 +1,80 @@
+// Copyright (C) 2025 ScyllaDB
+
+//go:build all || integration
+// +build all integration
+
+package repair_test
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"testing"
+
+	"github.com/scylladb/go-log"
+	"github.com/scylladb/gocqlx/v2"
+	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
+	. "github.com/scylladb/scylla-manager/v3/pkg/testutils"
+	. "github.com/scylladb/scylla-manager/v3/pkg/testutils/db"
+	. "github.com/scylladb/scylla-manager/v3/pkg/testutils/testconfig"
+)
+
+// Read only, should be used for checking testing environment
+// like Scylla version or tablets.
+var globalNodeInfo *scyllaclient.NodeInfo
+
+// Used to fill globalNodeInfo before running the tests.
+func TestMain(m *testing.M) {
+	if !flag.Parsed() {
+		flag.Parse()
+	}
+
+	config := scyllaclient.TestConfig(ManagedClusterHosts(), AgentAuthToken())
+
+	logger := log.NewDevelopment().Named("Setup")
+	c, err := scyllaclient.NewClient(config, logger)
+	if err != nil {
+		logger.Fatal(context.Background(), "Failed to create client", "error", err)
+	}
+
+	globalNodeInfo, err = c.AnyNodeInfo(context.Background())
+	if err != nil {
+		logger.Fatal(context.Background(), "Failed to get global node info", "error", err)
+	}
+
+	os.Exit(m.Run())
+}
+
+// Creates vnode keyspace.
+func createVnodeKeyspace(t *testing.T, session gocqlx.Session, keyspace string, rf1, rf2 int) {
+	stmt := "CREATE KEYSPACE %s WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': %d, 'dc2': %d}"
+	if globalNodeInfo.EnableTablets {
+		stmt += " AND tablets = {'enabled': false}"
+	}
+	ExecStmt(t, session, fmt.Sprintf(stmt, keyspace, rf1, rf2))
+}
+
+// Creates tablet keyspace or skips the test if that's not possible.
+func createTabletKeyspace(t *testing.T, session gocqlx.Session, keyspace string, rf1, rf2, tablets int) {
+	if !globalNodeInfo.EnableTablets {
+		t.Skip("Test requires tablets enabled in order to create tablet keyspace")
+	}
+	stmt := "CREATE KEYSPACE %s WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': %d, 'dc2': %d} AND tablets = {'enabled': true, 'initial': %d}"
+	ExecStmt(t, session, fmt.Sprintf(stmt, keyspace, rf1, rf2, tablets))
+}
+
+// Creates keyspace with default replication type (vnode or tablets).
+func createDefaultKeyspace(t *testing.T, session gocqlx.Session, keyspace string, rf1, rf2, tablets int) {
+	if globalNodeInfo.EnableTablets {
+		stmt := "CREATE KEYSPACE %s WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': %d, 'dc2': %d} AND tablets = {'enabled': true, 'initial': %d}"
+		ExecStmt(t, session, fmt.Sprintf(stmt, keyspace, rf1, rf2, tablets))
+		return
+	}
+	stmt := "CREATE KEYSPACE %s WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': %d, 'dc2': %d}"
+	ExecStmt(t, session, fmt.Sprintf(stmt, keyspace, rf1, rf2))
+}
+
+func dropKeyspace(t *testing.T, session gocqlx.Session, keyspace string) {
+	ExecStmt(t, session, fmt.Sprintf("DROP KEYSPACE IF EXISTS %q", keyspace))
+}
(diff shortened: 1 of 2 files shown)


pkg/service/repair/helper_integration_test.go
```diff
@@ -0,0 +1,467 @@
...
+	return strings.HasPrefix(req.URL.Path, oldRepairStatusPathPrefix) && req.Method == http.MethodGet
+}
+
+func isRepairStatusReq(req *http.Request) bool {
+	return isOldRepairStatusReq(req)
+}
+
+func isKillRepairReq(req *http.Request) bool {
+	return strings.HasPrefix(req.URL.Path, killRepairPathPrefix) && req.Method == http.MethodPost
+}
```
karol-kokoszka, line 155: I feel (it's subjective) this is boilerplate code.
Besides, it hides the logic.
Repair in SM bases on API calls to Scylla server. I don't see a reason for hiding the exact endpoint that is going to be intercepted.
It must be (for readability) explicitly stated in the code of the test that want's to intercept something. Otherwise, it's tricky to understand what the test is trying to do without jumping between the helper functions.

pkg/service/repair/helper_integration_test.go
```diff
@@ -0,0 +1,467 @@
...
+
+func newRepairSchedReq(t *testing.T, req *http.Request) (repairSchedReq, bool) {
+	if isOldRepairSchedReq(req) {
+		return newOldRepairSchedReq(t, req), true
+	}
+	return repairSchedReq{}, false
+}
+
+func newOldRepairSchedReq(t *testing.T, req *http.Request) repairSchedReq {
```
karol-kokoszka, line 164: Looking at the function body and the function name I have a little dissonance.
The name suggests I'm creating OldRepairSchedReq, but it returns repairSchedReq.
It suggests that it creates something new (constructor), but in fact it just parses data from the http.Request.

I assume this is intended to just parse POST /storage_service/repair_async

Could you please explicitly define the function repsonsibility ? E.g. "parsePostRepairAsyncEndpointPayload` ?

pkg/service/repair/helper_integration_test.go
```diff
@@ -0,0 +1,467 @@
...
+
+func isRepairStatusReq(req *http.Request) bool {
+	return isOldRepairStatusReq(req)
+}
+
+func isKillRepairReq(req *http.Request) bool {
+	return strings.HasPrefix(req.URL.Path, killRepairPathPrefix) && req.Method == http.MethodPost
+}
+
+func newRepairSchedReq(t *testing.T, req *http.Request) (repairSchedReq, bool) {
```
karol-kokoszka, line 157: Q: What is the meaning of the `repairSchedReq{}, false` return ? What does it indicate ? What information it holds ?

pkg/service/repair/helper_integration_test.go
```diff
@@ -0,0 +1,467 @@
...
+}
+
+func newRepairStatusReq(t *testing.T, req *http.Request) (repairStatusReq, bool) {
+	if isOldRepairStatusReq(req) {
+		return newOldRepairStatusReq(t, req), true
+	}
+	return repairStatusReq{}, false
+}
+
+func newOldRepairStatusReq(t *testing.T, req *http.Request) repairStatusReq {
```
karol-kokoszka, line 201: I already left the comment under newOldRepairSchedReq which is valid to this function as well.

pkg/service/repair/helper_integration_test.go
```diff
@@ -0,0 +1,467 @@
...
+	}
+	if status.id == "" {
+		t.Error("Not fully initialized old repair status req")
+		return repairStatusReq{}
+	}
+
+	return status
+}
+
+func newRepairSchedResp(t *testing.T, resp *http.Response) (repairSchedResp, bool) {
```
karol-kokoszka, line 219: What does it mean if this function returns falsE ?

pkg/service/repair/helper_integration_test.go
```diff
@@ -0,0 +1,467 @@
...
+			return resp, nil
+		}
+		if body, ok := mockRepairStatusRespBody(t, req, status); ok {
+			resp := httpx.MakeResponse(req, http.StatusOK)
+			resp.Body = body
+			return resp, nil
+		}
+		return nil, nil
+	})
+}
```
karol-kokoszka, line 346: I have a bit of the problem with this function.
The name states (at least that's how I interpret it):
- I'm creating the interceptor that filters out only RepairStatusEndpoint calls 

I don't know what it does after intercepting it.
I must go to the function body to understand that it's not only intercepting repair_status but repair_async as well.

Why is that ? What is the rationale for that implementation ?

pkg/service/repair/helper_integration_test.go
```diff
@@ -0,0 +1,467 @@
...
+
+func repairHoldInterceptor(t *testing.T, ctx context.Context, after int64) http.RoundTripper {
```
karol-kokoszka, line 348: What is the meaning of `after int64` ? 
I must carefully read function body to understand it.
Either add the comment or change the function name.

Or please just explain what you wanted to achieve with that function.

karol-kokoszka (commented):
I just left couple of comments.
The logic is kind of complicated and requires test reader to jump into the helper functions to understand what do they really do.

Ideally, the test should explicitly show which API calls are intercepted (like "storage_service/repair_async" etc).

--------------------------

> separates interceptor request parsing from the test code, so that it's easier to extend it with new tablet repair endpoints

Possibly yes, but it's not so obvious for the code reader.

> tries not to rely on tablet table ranges count, as the new tablet repair API will repair an entire table with a single call

Can you explain in few sentences what does it mean ? I understand that there is no need for ranges count, but please elaborate on how do you try to not rely on it.

> improves vnode/tablet keyspace creation

That is cool and useful.

pkg/service/repair/helper_integration_test.go
Michal-Leszczynski: I first made the implementation for both types of repair, but then decided to separate it into different PRs because:
- it makes the PRs smaller
- it's easier to see that I didn't "cheat" when refactoring tests just for the tablet repair API
- the general refactor wasn't blocked by the fact that we were not yet testing the not yet released Scylla exposing the tablet repair API

That's why functions `isRepairSchedReq` and `isRepairStatusReq` look useless. They are supposed to look like:
```go
func isRepairSchedReq(req *http.Request) bool {
	return isOldRepairSchedReq(req) || isTabletRepairSchedReq(req)
}
```

In general, the parsing code and interceptors need to check the endpoint in many places, so that's why those functions saved me from writing long lines of code like:
```go
	if (strings.HasPrefix(req.URL.Path, oldRepairSchedPathPrefix) && req.Method == http.MethodPost) ||
              (strings.HasPrefix(req.URL.Path, tabletRepairSchedPathPrefix) && req.Method == http.MethodPost) {}
```

pkg/service/repair/helper_integration_test.go
Michal-Leszczynski: I will update it with the comment. In order to simplify interceptors code, this function can be used on any request. If it's used on a request scheduling a repair, it will return parsed request and true. If it's used on any other request, it will return `repairSchedReq{}, false`.

pkg/service/repair/helper_integration_test.go
Michal-Leszczynski: >  Repair in SM bases on API calls to Scylla server. I don't see a reason for hiding the exact endpoint that is going to be intercepted.

We have lots of tests which use interceptors. I wanted those tests to run against tablet repair API as well. That's why I decoupled exact endpoint parsing to helper functions. It's because those tests don't care for the exact endpoint, but just for some general properties shared by different repair endpoints (e.g. keyspace/table).

> It must be (for readability) explicitly stated in the code of the test that want's to intercept something. Otherwise, it's tricky to understand what the test is trying to do without jumping between the helper functions.

That's true, but that should be the case. First thing is that the interceptor needs to be set in the test code (it's not some strange default). Second is that most of this PR's work is focused on decoupled parsing (which is easily extended for tablet API parsing in the next PR), which does not alter the request at all, it just gathers some parsed and useful information about repair request.

Michal-Leszczynski: You can see the extended tablet repair API parsing here: https://github.com/scylladb/scylla-manager/pull/4274/commits/f64c2840c9fc91c017e5c305dd8a382eb11ac23d. Please let me know what you think after reading my comments.

pkg/service/repair/helper_integration_test.go
Michal-Leszczynski: I will explain it with a comment. That's not a new function - it has been updated to use new parsing approach and moved to the helper file containing all predefined interceptors. `after` describes after how many calls to repair status should this function start blocking.

karol-kokoszka (approved):

Michal-Leszczynski:
@karol-kokoszka if you are fine with the way in which the comments were addressed, could you resolve them?

karol-kokoszka:
@Michal-Leszczynski It's ok, feel free to merge.

PR #4269, opened 2025-02-25
fix(config-cache): use netip.Addr to unify the IPv6

### LINKED ISSUE #4181: Don't identify host by IP string in config cache svc
As a follow up to #4166, I wanted to use config cache svc in restore svc in order to improve querying node info.

The problem is that config cache svc identifies node by IP string. Apart from the problem that node IP might change over time and we should handle it gracefully, the IP string itself is not a unique representation for given IP, e.g. `2001:db9:200:0:0:0:0:32` equals `2001:0DB9:200::32`.

In my case, some tests [failed](https://github.com/scylladb/scylla-manager/pull/4169/checks) when SM wanted to get node info:
```
2024-12-17T14:50:42.0601679Z     helper_integration_test.go:206: run backup: upload: 2001:db9:200:0:0:0:0:32: system_distributed_everywhere.cdc_generation_descriptions_v2: check Scylla backup API support: no config for node 2001:db9:200:0:0:0:0:32
```

This could also create other problems, as unfortunately SM code was written with the assumption that node can be identified by string IP.

karol-kokoszka:
Fixes #4181

This PR uses netip.Addr to try to parse the host parameter (that can be both, IP or DNS).
If it is able to parse the IP, it keeps this parsed IP (as string) as a key.
If it's not able to parse the IP, original string is kept as a key.

---

Please make sure that:

### DIFF
--- a/pkg/service/configcache/service.go
+++ b/pkg/service/configcache/service.go
@@ -4,6 +4,7 @@ package configcache
 
 import (
 	"context"
+	"net/netip"
 	"sync"
 	"time"
 
@@ -80,7 +81,18 @@ func (svc *Service) Read(clusterID uuid.UUID, host string) (NodeConfig, error) {
 		return emptyConfig, err
 	}
 
-	rawHostConfig, ok := clusterConfig.Load(host)
+	hostKey := host
+	// Remove leading '[' and trailing ']' if present.
+	// netip.ParseAddr(bracketedIP) leaves brackets.
+	// Need to remove it.
+	if host != "" && host[0] == '[' && host[len(host)-1] == ']' {
+		host = host[1 : len(host)-1]
+	}
+	if addr, err := netip.ParseAddr(host); err == nil {
+		hostKey = addr.String()
+	}
+
+	rawHostConfig, ok := clusterConfig.Load(hostKey)
 	if !ok {
 		return emptyConfig, ErrNoHostConfig
 	}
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski (commented):
What about some simple test checking two different string representations of the same IPV6 address?

karol-kokoszka:
> What about some simple test checking two different string representations of the same IPV6 address?

Yes, good idea. Let me add one.

VAveryanov8 (approved):
:+1:

karol-kokoszka:
@Michal-Leszczynski test is added.

pkg/service/configcache/service.go
```diff
@@ -80,7 +81,18 @@ func (svc *Service) Read(clusterID uuid.UUID, host string) (NodeConfig, error) {
...
+	hostKey := host
+	// Remove leading '[' and trailing ']' if present.
+	// netip.ParseAddr(bracketedIP) leaves brackets.
+	// Need to remove it.
+	if host != "" && host[0] == '[' && host[len(host)-1] == ']' {
+		host = host[1 : len(host)-1]
+	}
+	if addr, err := netip.ParseAddr(host); err == nil {
+		hostKey = addr.String()
+	}
```
Michal-Leszczynski, line 93: So we still need to be cautious about different IP `string` representations and we do some parsing manually.
Wouldn't it be safer to change the map key type from `string` to `netip.Addr` which is also comparable?

karol-kokoszka: The host IP is not really configurable. It's possible to set agent's port in scylla-manager-agent.yaml, but not possible to set it per host in scylla-manager.yam.

Issue #4273, opened 2025-02-26
Decide on tablet load balancing during repair


Michal-Leszczynski:
In terms of enabling/disabling tablet load balancing during repair, there are a few cases that need to be considered.

Before Scylla 2025.1, SM was disabling tablet load balancing during repair so that there is no tablet migration between multiple calls to repair API. Otherwise, it could happen that some tablet migrated from not yet repaired replica set to an already repaired replica set and thus avoid being repaired. 

With Scylla 2025.1.0 we have native tablet repair API, but it does not allow for partial tablet table repair (e.g. when using --dc, --host, --ignore-down-hosts SM flags). The idea was to keep on using the old repair API for such cases and use the new tablet repair API only for fully repaired tables. When using the old repair API, SM would still need to disable tablet load balancing for the time of partial repair of tablet table.

With Scylla 2025.1.1 we will have tablet repair API which will support partial table repair, so SM would never use the old repair API for repairing tablet tables and it wouldn't need to stop tablet load balancing during repair at all.

First problem is that the old repair API might not be safe (https://github.com/scylladb/scylladb/issues/23008).

Second problem is that when aiming for high disk utilization (~90%), stopping tablet load balancing during repair (which takes a lot of time), might really harm the scale out procedure (cc: @mykaul).

For Scylla 2025.1.1 our those problems will disappear, as SM will always be using tablet repair API and it won't disable tablet load balancing during repair.

On the other hand, it needs to be decided what to do for the 2025.1.0 release and if older releases need any adjustments to the repair procedure.

Ref: use new tablet repair API by SM https://github.com/scylladb/scylla-manager/issues/4188
Ref: initial SM and tablets integration https://github.com/scylladb/scylla-manager/issues/3516

cc: @tzach @karol-kokoszka @bhalevy @asias

asias:
I vote for backporting https://github.com/scylladb/scylladb/pull/22621 to 2025.1.0.  It is merged for a while. The risk is low, the filter is not used by default.  Yet, the benefit is huge for SM implementation.

Michal-Leszczynski:
> I vote for backporting https://github.com/scylladb/scylladb/pull/22621 to 2025.1.0. It is merged for a while. The risk is low, the filter is not used by default. Yet, the benefit is huge for SM implementation.

This would make the tablet balancing problems during repair disappear for 2025.1, which is nice.
@mykaul @tzach @bhalevy can we do it?

mykaul:
[...]

PR #4274, opened 2025-02-26
Tablet repair API

### LINKED ISSUE #4188: Integrate tablet repair scheduler
In scylla commit 0d2583600d1325f2064a0d5d776bcf50660a5a42 (Merge 'Add tablet repair scheduler support' from Asias He), the tablet repair scheduler is implemented. A new tablet repair api is added. With this new api, the request of the repairs will be scheduler by scylla core along with other tablet tasks, e.g., migration, rebuild. There is no need for the management tool to schedule repair tasks on different nodes any more. One can run this api to repair tablets of a given table on any of the nodes.  

Note: Currently, node and dc selection are not supported yet. We might support it later, there is a PR: https://github.com/scylladb/scylladb/pull/21985. 

```
     {        
         "path":"/storage_service/tablets/repair",
         "operations":[
            {
               "nickname":"repair_tablet",
               "method":"POST",
               "summary":"Repair a tablet",
               "type":"void",
               "produces":[
                  "application/json"
               ],
               "parameters":[
                  {
                     "name":"ks",
                     "description":"Keyspace name to repair",
                     "required":true,
                     "allowMultiple":false,
                     "type":"string",
                     "paramType":"query"
                  },
                  {
                     "name":"table",
                     "description":"Table name to repair",
                     "required":true,
                     "allowMultiple":false,
                     "type":"string",
                     "paramType":"query"
                  },
                  {
                     "name":"tokens",
                     "description":"Tokens owned by the tablets to repair. Multiple tokens can be provided using a comma-separated list. When set to the special word 'all', all tablets will be repaired",
                     "required":true,
                     "allowMultiple":false,
                     "type":"string",
                     "paramType":"query"
                  }
               ]
            }
         ]
      },
```

Michal-Leszczynski:
This PR adds support for tablet repair API in Scylla 2025.1.0:
- for Scylla < 2025.1.0 - repairing tablet keyspaces stops tablet load balancing and uses old repair API
- for Scylla == 2025.1.0 - SM never stops tablet load balancing, it uses tablet repair API for repairing entire tablet table and old repair API for repairing partial tablet table
- for Scylla >= 2025.1.1 - SM never stops tablet load balancing and always uses tablet repair API which is extended with host filtering capabilities (not yet implemented)

See #4273 for the decision process about not stopping tablet load balancing.

Fixes #4188

### DIFF
--- a/pkg/scyllaclient/client_scylla.go
+++ b/pkg/scyllaclient/client_scylla.go
@@ -591,6 +591,27 @@ func ReplicaHash(replicaSet []string) uint64 {
 	return hash.Sum64()
 }
 
+// TabletRepair schedules Scylla repair tablet table task and returns its ID.
+// The whole table will be repaired on all nodes with just a single task.
+// The host is only needed so that we know which node should be queried
+// for the task status.
+func (c *Client) TabletRepair(ctx context.Context, keyspace, table, host string) (string, error) {
+	const allTablets = "all"
+	dontAwaitCompletion := "false"
+	p := operations.StorageServiceTabletsRepairPostParams{
+		Context:         forceHost(ctx, host),
+		Ks:              keyspace,
+		Table:           table,
+		Tokens:          allTablets,
+		AwaitCompletion: &dontAwaitCompletion,
+	}
+	resp, err := c.scyllaOps.StorageServiceTabletsRepairPost(&p)
+	if err != nil {
+		return "", err
+	}
+	return resp.GetPayload().TabletTaskID, nil
+}
+
 // Repair invokes async repair and returns the repair command ID.
 func (c *Client) Repair(ctx context.Context, keyspace, table, master string, replicaSet []string, ranges []TokenRange, intensity int, smallTableOpt bool) (int32, error) {
 	dr := dumpRanges(ranges)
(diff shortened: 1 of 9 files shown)


pkg/scyllaclient/client_scylla.go
```diff
@@ -1159,6 +1180,75 @@ func (c *Client) RaftReadBarrier(ctx context.Context, host, groupID string) erro
...
+	ctx = withShouldRetryHandler(ctx, scyllaWaitTaskShouldRetryHandler)
+	ctx = forceHost(ctx, host)
+	ctx = noTimeout(ctx)
+	p := &operations.TaskManagerWaitTaskTaskIDGetParams{
+		Context: ctx,
+		TaskID:  id,
+	}
+	if longPollingSeconds > 0 {
+		p.SetTimeout(&longPollingSeconds)
+	}
```
karol-kokoszka, line 1218: The long polling timeout is handled by the server. 
If the task is not done withing "longPollingSeconds" period, then scylla responds with "RUNNING", right ?
At the same time, we don't set timeout on the API call.
It means that the timeout must be defined on Scylla client that we use to do the call.
And it is defined in `pkg/scyllaclient/config.go TimeoutConfig`.

The default long-polling is 10 seconds, client timeout is 30 seconds.

If the client times out, after several backoffs, the repair ends with error.
What is actually expected.

Please validate if my understanding is correct.
Overall it's desired behavior.

karol-kokoszka (commented):
no comments, overall is ok, but i understand other PRs must be merged first.

pkg/scyllaclient/client_scylla.go
Michal-Leszczynski: > The long polling timeout is handled by the server.

Yes

> If the task is not done withing "longPollingSeconds" period, then scylla responds with "RUNNING", right ?

It responds with error and `http.StatusRequestTimeout`, so it's different from the old repair API.

> At the same time, we don't set timeout on the API call.
It means that the timeout must be defined on Scylla client that we use to do the call.

You are right, not setting the timeout at all (on SM side) is too optimistic. We should set it according to the logic from other long polling calls:
```go
// When using long polling, wait duration starts only when node receives the
// request.
// longPollingTimeout is calculating timeout duration needed for request to
// reach node so context is not canceled before response is received.
func (c *Client) longPollingTimeout(waitSeconds int) time.Duration {
	return time.Second*time.Duration(waitSeconds) + c.config.Timeout
}
```

> If the client times out, after several backoffs, the repair ends with error.
What is actually expected

So we are not doing re-tries on Scylla side timeouts, as they are expected, and we just want to keep on long polling the status. We only re-try on other errors. If we set timeout on SM side and it times out, we should use the re-try mechanism. But we don't have a total timeout for all combined calls for getting finished repair status. So if Scylla keeps on time outing on its Side, SM will keep on querying the repair status. In my opinion it's fine, as we can't have any realistic (and useful) timeout for repairing an entire tablet table without implementing some logic based on dataset size, tablet count, cluster topology.

karol-kokoszka: >> If the task is not done withing "longPollingSeconds" period, then scylla responds with "RUNNING", right ?

> It responds with error and http.StatusRequestTimeout, so it's different from the old repair API.

Got it. I see you coupled `ScyllaWaitTask` with `ScyllaTaskProgress` which is called when first is timed out...
OK, It's fine.

pkg/service/repair/generator.go
```diff
@@ -125,11 +132,15 @@ func (g *generator) Run(ctx context.Context) (err error) {
...
 	for _, ksp := range g.plan.Keyspaces {
-		// Disable tablet migration when repairing tablet table.
-		// Without that it could be possible that some tablet "escapes" being
-		// a repaired by migrating from not yet repaired token range to already repaired one.
-		if err := g.ringDescriber.ControlTabletLoadBalancing(ctx, g.ringDescriber.IsTabletKeyspace(ksp.Keyspace)); err != nil {
-			return errors.Wrapf(err, "control tablet load balancing")
+		if !g.plan.apiSupport.tabletRepairNoHostFiltering {
+			// Disable tablet migration when repairing tablet table.
+			// Without that it could be possible that some tablet "escapes" being
+			// a repaired by migrating from not yet repaired token range to already repaired one.
```
mykaul, line 138:
```suggestion
			// repaired by migrating from not yet repaired token range to already repaired one.
```

Michal-Leszczynski:
@karol-kokoszka This PR is ready (I fixed some remaining tests). The implementation still follows PR description.

VAveryanov8 (approved):

Michal-Leszczynski:
@karol-kokoszka repair nemesis is running: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/nemesis-mgmt-repair-aws/

karol-kokoszka:
@Michal-Leszczynski @mikliapko There are a couple of runs of these jenkins job. Which one to validate ?

mikliapko:
> @Michal-Leszczynski @mikliapko There are a couple of runs of these jenkins job. Which one to validate ?

We started with the analysis of run 3
https://argus.scylladb.com/tests/scylla-cluster-tests/76bc8dc7-eb4a-43be-8bea-8dc74a391953

It does scale out from 3 to 6 nodes and triggers repair right after it.

PR #4278, opened 2025-02-28
Don't create views on tablet tables

### LINKED ISSUE #4271: Don't create views on tablet tables in tests
Because of https://github.com/scylladb/scylladb/issues/21832, out integration tests are [failing](https://github.com/scylladb/scylla-manager/actions/runs/13541774481/job/37844252469?pr=4270) against Scylla 2025.1.0.
We should skip all tests which create views on tablet tables.

cc: @karol-kokoszka @VAveryanov8

Michal-Leszczynski:
Fixes #4271

### DIFF
--- a/pkg/service/backup/service_backup_integration_test.go
+++ b/pkg/service/backup/service_backup_integration_test.go
@@ -2556,6 +2556,13 @@ func TestBackupViewsIntegration(t *testing.T) {
 
 	Print("Given: table with MV and SI")
 	WriteData(t, clusterSession, testKeyspace, 1, testTable)
+
+	// It's not possible to create views on tablet keyspaces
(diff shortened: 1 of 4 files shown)


VAveryanov8 (approved):

Michal-Leszczynski:
It looks like I missed some places - I will merge this PR with #4270 so we have some validation.

karol-kokoszka (commented):
The tests are failing on create-schema on older versions of Scylla.
Why is that ?

Michal-Leszczynski:
> The tests are failing on create-schema on older versions of Scylla.
Why is that ?

Create schema tests also include the get target tests and I forgot to update them.
It should be good now.

karol-kokoszka (approved):

PR #4291, opened 2025-03-06
fix(consul): use prometheus_address|port from scylla.yaml for consul endpoint

### LINKED ISSUE #4300: Consul endpoint always returns node address instead of `prometheus_address` from scylla.yaml
Ref https://github.com/scylladb/scylla-enterprise/issues/5234

It appears that consul endpoint in Scylla Manager always returns node IP address instead of the one configured in `prometheus_address` in scylla.yaml.
If someone decide to use consul for discovering scylla nodes in monitoring stack + uses different `prometheus_address` than the `listen_address`, prometheus-server will not be able to discover nodes.

API is defined here https://github.com/scylladb/scylla-manager/blob/9dc2b7e4da11447d602b4031c494e0b8bbc718b4/pkg/restapi/consul.go#L28-L34

karol-kokoszka:
fixes #4300 

This PR:
- extends test docker environment so that Scylla's prometheus_address is overwritten and belongs to network A, but Scylla's listen address belongs to network B
- prometheus server is started in network A
- `prometheus.yaml` is updated to scrap Scylla Manager's consul endpoint instead of hardcoded targets.
- the test validating consul endpoint is updated to reflect the changes 
- consul endpoint is fixed, so that it returns prometheus_address instead of listen_address. The latter is still returned if `prometheus_address` is set to `0.0.0.0`

Example:
image

192.168.200.11 node was introduced here with --prometheus-address 192.168.200.11 value => that's why it's in red.
Prometheus is in the network 192.168.100.0/24.
Nodes (listen-address) are in the network 192.168.200.0/24.

Prometheus server is not able to reach 192.168.200.0/24.
Consul endpoint gives a list of prometheus addresses of all the nodes.

---

Please make sure that:


karol-kokoszka:
@Michal-Leszczynski @VAveryanov8 this PR is ready for the review !

pkg/testutils/testconfig/testconfig.go
```diff
@@ -52,6 +53,17 @@ func IPFromTestNet(hostIPending string) string {
 	return *flagTestNet + hostIPending
 }
 
+// IPFromSecondTestNet returns IP from the host network:
+//
+//	IPFromTestNet("11") -> 192.168.100.11
+//	IPFromTestNet("11") -> 2001:0DB9:100::11.
```
Michal-Leszczynski, line 59: What is this formatting?

pkg/service/cluster/service.go
```diff
@@ -667,10 +669,27 @@ func (s *Service) ListNodes(ctx context.Context, clusterID uuid.UUID) ([]Node, e
...
+			if err != nil {
+				return nil, errors.Wrapf(err, "node info call of %s", h)
+			}
+			promAddr := ni.PrometheusAddress
+			if temp, err := netip.ParseAddr(ni.PrometheusAddress); err == nil {
+				promAddr = temp.String()
+			}
+			if promAddr == "" || promAddr == "0.0.0.0" || promAddr == net.ParseIP("0:0:0:0:0:0:0:0").String() {
+				promAddr = h
+			}
```
Michal-Leszczynski, line 682: It's better to use already implemented checks like https://pkg.go.dev/net#IP.IsUnspecified

pkg/service/cluster/service.go
```diff
@@ -667,10 +669,27 @@ func (s *Service) ListNodes(ctx context.Context, clusterID uuid.UUID) ([]Node, e
...
 				s.logger.Error(ctx, "Failed to get number of shards", "error", err)
 			}
+			ni, err := client.NodeInfo(ctx, h)
+			if err != nil {
+				return nil, errors.Wrapf(err, "node info call of %s", h)
+			}
+			promAddr := ni.PrometheusAddress
+			if temp, err := netip.ParseAddr(ni.PrometheusAddress); err == nil {
+				promAddr = temp.String()
+			}
```
Michal-Leszczynski, line 679: Why is this step important?

Michal-Leszczynski (commented):
In general LGTM, just a few comments!

VAveryanov8 (commented):
Cool PR! Just a few comments/questions

pkg/testutils/testconfig/testconfig.go
karol-kokoszka: Good catch, this is copy paste comment from IPFromTestNet method comment.
The examples just show the output for given parameter.

Updated.

pkg/service/cluster/service.go
karol-kokoszka: makes sense.
updated

pkg/service/cluster/service.go
karol-kokoszka: ni.PrometheusAddress is a string, so I assume in the code that it may hold everything

testing/docker-compose.yaml
```diff
@@ -174,10 +174,17 @@ services:
 
   prometheus:
     image: prom/prometheus:${PROMETHEUS_VERSION}
```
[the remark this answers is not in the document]
karol-kokoszka: it's here https://github.com/scylladb/scylla-manager/blob/5b8b401fad9273b75e891c6c2f94ec70933e46e0/testing/.env#L18

testing/docker-compose-ipv6.yaml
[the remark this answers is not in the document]
karol-kokoszka: definitely

testing/docker-compose.yaml
```diff
@@ -174,10 +174,17 @@ services:
...
+    command:
+      - '--config.file=/etc/prometheus/prometheus.yml'
+      - '--storage.tsdb.path=/prometheus'
+      - '--web.console.libraries=/usr/share/prometheus/console_libraries'
+      - '--web.console.templates=/usr/share/prometheus/consoles'
+      - '--log.level=debug'
+      - '--log.format=json'
```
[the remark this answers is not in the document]
karol-kokoszka: it's leftover from my debugging
thnkas !

VAveryanov8 (approved):
:+1:

pkg/service/cluster/service.go
```diff
@@ -667,10 +668,31 @@ func (s *Service) ListNodes(ctx context.Context, clusterID uuid.UUID) ([]Node, e
...
+
+			promAddr := h
+			if ni.PrometheusAddress != "" {
+				promAddr = ni.PrometheusAddress
+				if temp, err := netip.ParseAddr(ni.PrometheusAddress); err == nil {
+					if !temp.IsUnspecified() {
+						promAddr = temp.String()
+					}
+				}
+			}
```
Michal-Leszczynski, line 684: I still don't get why do we bother with parsing.

If `ni.PrometheusAddress` is set, the we  set `promAddr` to it.
Then we verify if the address is correct and specified - if that's the case, we basically set `promAddr` to the `ni.PrometheusAddress` yet again.
Why do we do that?
Do we care for the address to be represented in some deterministic way and that's why we parse and print it here?
More importantly, shouldn't we reset `promAddr` to `h` if the check does not succeed?

pkg/testutils/testconfig/testconfig.go
Michal-Leszczynski: To be honest I meant strange tabulation and dot at the end, but it's a nit so I'm not going to stall this PR because of that.

pkg/service/cluster/service.go
karol-kokoszka: Just for safety.
The only check before, is to validate that the string is not empty.
But, as it is the string, it may contain whatever text.

karol-kokoszka: > Do we care for the address to be represented in some deterministic way?

Possibly yes — netip.ParseAddr().String() will normalize the address, which can be useful if you later use it as a key or in metrics. But this is not obvious without a comment.

> Shouldn't we reset promAddr to h if the check does not succeed?

Yes, that would be safer if you want to avoid storing incorrect Prometheus addresses. Currently, it uses ni.PrometheusAddress even if it's unspecified or unparseable.

Thanks. That was a very good finding.

Michal-Leszczynski (approved):

PR #4301, opened 2025-03-11
tablet repair API filtering

Michal-Leszczynski:


pkg/service/repair/service.go
```diff
@@ -176,6 +175,49 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
 	return t, nil
 }
 
+func validateIgnoreDownNodes(t Target, status scyllaclient.NodeStatusInfoSlice) error {
+	// Ensure that either there are no down nodes in repaired DCs,
+	// or that --ignore-down-nodes is set.
```
VAveryanov8, line 180:
```suggestion
	// or that --ignore-down-hosts is set.
```

pkg/service/repair/service.go
```diff
@@ -176,6 +175,49 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
...
+	downNodesInDCs := status.Datacenter(t.DC).Down().Hosts()
+	if len(downNodesInDCs) == 0 || len(t.IgnoreHosts) > 0 {
+		return nil
+	}
+	return errors.Errorf("repairing DCs with down nodes won't be successful: %v. "+
+		"Please either exclude down nodes from being repaired with --ignore-down-nodes flag, "+
```
VAveryanov8, line 186:
```suggestion
		"Please either exclude down nodes from being repaired with --ignore-down-hosts flag, "+
```

pkg/service/repair/service.go
```diff
@@ -176,6 +175,49 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
...
+}
+
+func validateHost(t Target, p *plan, dcMap map[string][]string) error {
+	// Nothing to validate - no --host flag
+	if t.Host == "" {
+		return nil
+	}
+	// Ensure Host is not ignored
+	if slices.Contains(t.IgnoreHosts, t.Host) {
+		return errors.New("host can't have status down")
```
VAveryanov8, line 197: nit: some comment that IgnoreHosts actually contains list of hosts that are down  might be helpful - I had to check how t.IgnoreHosts is populated in order to understand  what's going on here - if host is in the ignoreHosts slice we return error about hosts status down :)

VAveryanov8 (approved):
:+1:

Michal-Leszczynski:
I guess we can merge it with the hand picked Scylla version and change it to `scylla-nightly:latest` in #4306.

pkg/service/repair/generator.go
```diff
@@ -125,19 +125,20 @@ func (g *generator) Run(ctx context.Context) (err error) {
...
+		if !g.plan.apiSupport.tabletRepair {
 			// Disable tablet migration when repairing tablet table.
 			// Without that it could be possible that some tablet "escapes" being
 			// repaired by migrating from not yet repaired token range to already repaired one.
-			// We don't need to do it when tablet repair API is available
-			// (even if it does not allow for host filtering).
+			// We don't need to do it when tablet repair API is available.
 			if err := g.ringDescriber.ControlTabletLoadBalancing(ctx, g.ringDescriber.IsTabletKeyspace(ksp.Keyspace)); err != nil {
 				return errors.Wrapf(err, "control tablet load balancing")
 			}
```
karol-kokoszka, line 144: My understanding is:
- per every ks, if Scylla API doesn't support tablet's API then 
   - KS is tablet based => keep the load-balancing enabled
   - KS is vnodes based => disable load-balancing
  
It it opposite to the comment:

> 			// Disable tablet migration when repairing tablet table.
			// Without that it could be possible that some tablet "escapes" being
			// repaired by migrating from not yet repaired token range to already repaired one.
			// We don't need to do it when tablet repair API is available.

Expected ? Or my understanding is wrong ?

Michal-Leszczynski: It looks like there is missing `!` when checking if keyspace is tablet replicated, and unfortunately it is not introduced by this PR, but it was like this from the very beginning....
The good thing is that tablet's tombstone gc mode repair protects us from things like data resurrection, but this is still a big issue that wasn't caught by any test.
I will fix it in this PR and add some tests for it as well.

Michal-Leszczynski:
@VAveryanov8 could you take one more look at this PR? I added two commits since your review.

pkg/service/repair/service_repair_integration_test.go
```diff
@@ -2237,25 +2231,38 @@ func TestServiceRepairIntegration(t *testing.T) {
...
-				singleCall: false,
+				ks:            vnodeKs,
+				tab:           []string{t1, t2},
+				api:           repairAsyncEndpoint,
+				singleCall:    false,
+				loadBalancing: true,
 			},
 		}
 
+		if ok, err := globalNodeInfo.SupportsTabletRepair(); err != nil {
```
VAveryanov8, line 2256: nit: 

```go
func supportsTabletRepair(t *testing.T) bool {
   ok, err := globalNodeInfo.SupportsTabletRepair()
   if err != nil {
      t.Fatal(err)
   }
   return ok
}

....

if !supportsTableRepair(t) {
	// For this Scylla version tablet tables
	// are still repaired with the old repair API.
	testCases[0].api = repairAsyncEndpoint
	testCases[0].singleCall = false
	testCases[0].loadBalancing = false
}

```

VAveryanov8:
Looks good to me :+1:

PR #4315, opened 2025-03-26
Scyllaclient: increase timeout on disabling tablet load balancing

### LINKED ISSUE #4313: Increase timeout on disabling tablet load balancing
SM disables tablet load balancing when:
- repairing tablet tables for Scylla older than 2025.1
- taking snapshot during backup

SM should also disable it during restore (#4243).

The problem is that SM uses the default (30s) timeout for this call, but it actually might require much more time due to currently running migrations. In such case, SM uses retries, but in some scenario that's not enough and it results in the following error:
```
Status:		ERROR
Cause:		control tablet load balancing: giving up after 6 attempts: after 30s: context deadline exceeded
control post repair tablet load balancing: giving up after 6 attempts: after 30s: context deadline exceeded
```

Retries are not ideal, because they don't prevent migrations from being started during intervals between the calls.
Because of that, the timeout should be increased.

Ref #4243

Michal-Leszczynski:
Fixes #4313


Michal-Leszczynski:
cc: @mikliapko

karol-kokoszka (approved):

PR #4316, opened 2025-03-27
Move to AGPL

### LINKED ISSUE #4229: Move Scylla Manager to AGPL
> The source available Scylla Manager will move to AGPL and the closed source Kubernetes multi-region operator will be merged with the main Apache-licensed ScyllaDB Kubernetes operator.

https://www.scylladb.com/2024/12/18/why-were-moving-to-a-source-available-license/

As ScyllaDB Enterprise and ScyllaDB Open Source move to Source Avilable, we can open source Scylla Manager as AGPL.

- Update the license to AGPL  and use it for the next Manager minor release
- Remove the 5 node limits, but only for Enterprise 2025.1 release and later.

Michal-Leszczynski:
It's worth to note that the old SM licenses are still available under:
- older release branches
- https://www.scylladb.com/2024/12/18/why-were-moving-to-a-source-available-license/
- https://www.scylladb.com/scylla-source-available-license/

Fixes #4229

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,62 +1,661 @@
-Scylla SOURCE AVAILABLE LICENSE (SSAL) AGREEMENT Last Update: June 30, 2021
+                    GNU AFFERO GENERAL PUBLIC LICENSE
+                       Version 3, 19 November 2007
 
-This Agreement sets forth the terms on which the Licensor makes available the Software. BY INSTALLING, DOWNLOADING, ACCESSING, USING OR DISTRIBUTING ANY OF THE SOFTWARE, YOU AGREE TO THE TERMS AND CONDITIONS OF THIS AGREEMENT. IF YOU DO NOT AGREE TO SUCH TERMS AND CONDITIONS, YOU MUST NOT USE THE SOFTWARE. 
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-If you are receiving the Software on behalf of a legal entity, you represent and warrant that you have the actual authority to agree to the terms and conditions of this agreement on behalf of such entity. 
+                            Preamble
 
-The terms below have the meanings set forth below for purposes of this Agreement: 
+  The GNU Affero General Public License is a free, copyleft license for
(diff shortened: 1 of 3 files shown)


tzach:
Let's wait to merge this PR; at least users start using 2025.1

Michal-Leszczynski:
> Let's wait to merge this PR; at least users start using 2025.1

@tzach  my understanding was that we would like to release SM 3.5.0 (compatible with 2025.1.0) even before 2025.1, so that users are not delayed because of the lack of SM support. If we are going to do it like that, then it would be good to change the license on the SM 3.5.0 release without waiting for the 2025.1 to settle in.

cc: @mykaul

mykaul:
> > Let's wait to merge this PR; at least users start using 2025.1
> 
> @tzach my understanding was that we would like to release SM 3.5.0 (compatible with 2025.1.0) even before 2025.1, so that users are not delayed because of the lack of SM support. If we are going to do it like that, then it would be good to change the license on the SM 3.5.0 release without waiting for the 2025.1 to settle in.
> 
> cc: @mykaul

Indeed. We need a release to accompany 2025.1.0 which will be released today/tomorrow.

tzach:
> Indeed. We need a release to accompany 2025.1.0 which will be released today/tomorrow.

3.5 with 2025.1 support is urgent
AGPL change is not urgent

Are they coupled?

Michal-Leszczynski:
> Are they coupled?

@tzach It's just a question of which SM release should start using AGPL license.

I understand that AGPL is less limiting than the current SM license, so changing it shouldn't break anything for the user.
It still doesn't mean that we should change it in a patch release - it should be minor at least.

So if we won't include it in the SM 3.5.0, the next opportunity would be SM 3.6.0. 
I just thought that it would be good to update license in the release which supports Scylla with updated license, but if you think that's not urgent, then we can update the license later on.

tzach (approved):

Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 Please take a look and advise if that's the right time/release/method to update the license!

karol-kokoszka (approved):

karol-kokoszka:
@annastuchlik @tzach 
What about this page ? https://www.scylladb.com/scylla-source-available-license/
It states the SSAL license for the Manager.

Who maintains this page ?

PR #4318, opened 2025-03-27
Update compatibility matrix

### LINKED ISSUE #4311: docs: Issue on page Compatibility Matrix
I would like to report an issue on page https://manager.docs.scylladb.com/branch-3.4/restore/compatibility-matrix

### Problem
The compatibility table does not include 3.4 within version 3.4

###  Suggest a fix

Add 3.4 compatibility information.

Michal-Leszczynski:
Fixes #4311


docs/source/compatibility-matrix.rst
```diff
@@ -12,6 +12,12 @@ The following table shows which version of Scylla Manager restore task supports
    * - ScyllaDB Manager Version
      - ScyllaDB Open Source Version
      - ScyllaDB Enterprise Version
+   * - 3.5
+     - 5.4, 6.0, 6.1, 6.2
```
tzach, line 16: No need to support old version of ScyllaDB here. 6.2 is good enough.
We want to encourage user to upgrade to latest - 6.2 and 2025.1 right after

karol-kokoszka (approved):

PR #4322, opened 2025-03-30
DRAFT: Verify that manager and agent versions match

atpugtihsrah:
---

Please make sure that:


atpugtihsrah:
@Michal-Leszczynski Could you please review this PR?

atpugtihsrah:
Sanity Check was failing because there was no comment on the exported method I created. It looks like workflows have to be approved each time a new commit is made.

atpugtihsrah:
@Michal-Leszczynski Could you please review the PR when you have some time?

atpugtihsrah:
Hi @karol-kokoszka , thanks for the feedback. I'll make some changes. 

> Potentially, the check could be done at the beginning of the particular task execution.

If the versions don't match, then should the task execution be cancelled/stopped/disallowed?

Issue #4280, opened 2025-03-03
Attend to mykaul's comments from #4244 (Move Manager containers to be UBI based)


mykaul:
Multiple nitpicks and not so nitpicks around the Dockerfile.
1. If it's available without auth., use Red Hat's registry for UBI images, nothing else.
hint:
```
ykaul@ykaul:~/github/scylla-cluster-tests$ podman pull ubi9
Resolved "ubi9" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull registry.access.redhat.com/ubi9:latest...
Getting image source signatures
Checking if image destination supports signatures
Copying blob 3d1c86608284 done   | 
Copying blob de21b519acd7 [=====>--------------------------------] 13.0MiB / 84.4MiB | 33.5 MiB/s
```

2. Make sure you don't add extra layers, if you don't squash the image later.
3. Cleanup of minidnf is via `microdnf clean all`

Hope this is all.

mykaul:
Ref https://github.com/scylladb/scylla-manager/pull/4244#issuecomment-2695377028

PR #4327, opened 2025-04-01
docker: fix Docker image to pass OpenShift Certification

syuu1228:
To make docker image ready for Red Hat OpenShift Software Certification,
we need to fix following requirements which mentioned on the document of
the certification:
 - HasRequiredLabel: we need to define labels which describes the image
 - HasLicense: we need to add license file to /licenses

(described at:
https://docs.redhat.com/en/documentation/red_hat_software_certification/2025/html/red_hat_openstack_certification_policy_guide/index)

With this patch, openshift-preflight will returns following result:
  "Preflight result: PASSED"

related #4242

### DIFF
--- a/dist/.goreleaser-docker.yaml
+++ b/dist/.goreleaser-docker.yaml
@@ -16,6 +16,15 @@ dockers:
     extra_files:
       - docker/scylla-manager.yaml
       - release
+      - license/
+    build_flag_templates:
+      - "--label=name=scylla-manager"
+      - "--label=maintainer=Michał Jan Matczuk <michal@scylladb.com>"
+      - "--label=vendor=ScyllaDB, Inc."
+      - "--label=version={{.Version}}"
+      - "--label=release=1"
+      - "--label=summary=Scylla Manager"
+      - "--label=description=Scylla Manager"
 
   - ids:
     use: docker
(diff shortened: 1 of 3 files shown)


yaronkaikov (approved):

yaronkaikov:
@syuu1228 let's also address in this PR (can be a different commit)  https://github.com/scylladb/scylla-manager/issues/4280, i would change the ref for https://github.com/scylladb/scylla-manager/issues/4242 to `Fixes` since it should close it 

Thank you for your fast response on this one

Michal-Leszczynski:
> @syuu1228 let's also address in this PR (can be a different commit) https://github.com/scylladb/scylla-manager/issues/4280, i would change the ref for https://github.com/scylladb/scylla-manager/issues/4242 to Fixes since it should close it

@yaronkaikov @syuu1228 can we expect #4280 to be fixed today?

Michal-Leszczynski (approved):
I see that we are still using `Michał Jan Matczuk <michal@scylladb.com>` in a few places. I will change it in a separate PR after this one gets merged.

cc: @karol-kokoszka

Michal-Leszczynski:
@yaronkaikov @syuu1228 I'm merging this PR for now (as we want to get ready for the release), so please address #4280 in a new PR. Thanks for your help!

Issue #4329, opened 2025-04-01
Manager's query to perform 'cqlping' is inefficient and causes a scan instead of a local query


mykaul:
Very similar to changes we've recently made in the drivers (see https://github.com/scylladb/java-driver/issues/282 and https://github.com/scylladb/python-driver/issues/418 the query to system.local should be with the WHERE close to prevent a scan.

PR #4330, opened 2025-04-01
cqlping: perform a query and not a scan

mykaul:
Very similar to changes we've recently made in the drivers (see scylladb/java-driver#282 and scylladb/python-driver#418 , the query to system.local should be with the WHERE close to prevent a scan and just issue a local query.

Fixes: https://github.com/scylladb/scylla-manager/issues/4329

---

Please make sure that:

### DIFF
--- a/docs/source/health-check.rst
+++ b/docs/source/health-check.rst
@@ -141,7 +141,7 @@ CQL query health check
 
 You may specify CQL ``username`` and ``password`` flags when adding cluster to Scylla Manager using :ref:`sctool cluster add <cluster-add>` command.
 It's also possible to add or change that using :ref:`sctool cluster update <cluster-update>` command.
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
Test failure due to flake #4334.

Michal-Leszczynski:
@mykaul should we backport this fix to `branch-3.5`, or can it wait for SM 3.5.1?
Is it suspected to cause troubles with healthcheck for some customers?

mykaul:
> @mykaul should we backport this fix to `branch-3.5`, or can it wait for SM 3.5.1? Is it suspected to cause troubles with healthcheck for some customers?

I took the conservative approach here. We've had this for years, it's not urgent. But we need to see that tests cover cqlping functionality (do they?) - I did not even compile it.

mykaul:
@Michal-Leszczynski - any idea what this CI failure is about?

Michal-Leszczynski:
The failure is due to a flake: https://github.com/scylladb/scylla-manager/issues/4334, so no problems with this PR.

PR #4332, opened 2025-04-01
Docs: update latest version to 3.5

Michal-Leszczynski:


Michal-Leszczynski:
@karol-kokoszka This PR is needed for SM 3.5 docs to be visible.

karol-kokoszka (approved):

PR #4335, opened 2025-04-01
docker: small fix-ups for UBI9 containers

### LINKED ISSUE #4280: Attend to mykaul's comments from #4244 (Move Manager containers to be UBI based)

syuu1228:
Do following fix-ups pointed out on #4280:

- Use Red Hat registry for UBI image
- Minimize layers
- Run cleanup of microdnf

fixes #4280

### DIFF
--- a/dist/docker/scylla-manager-agent.dockerfile
+++ b/dist/docker/scylla-manager-agent.dockerfile
@@ -1,14 +1,15 @@
-FROM docker.io/redhat/ubi9-minimal:latest
+FROM registry.access.redhat.com/ubi9-minimal:latest
 ARG ARCH=x86_64
+
+COPY release/scylla-manager-agent*$ARCH.rpm /
+COPY license/LICENSE.* /licenses/
 
 RUN microdnf -y update && \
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski (approved):
LGTM, but let's wait for @yaronkaikov review as well

PR #4336, opened 2025-04-01
Update ScyllaDB versions used in Github actions

mykaul:
1. Removed 2023.1.x
2. Update 6.2.0 to 6.2.3
3. Update 2024.1.11 to 2024.1.15
4. Update 2025.1.0-rc4 to 2025.1.0 GA

---

Please make sure that:
- Code is split to commits that address a single change
- Commit messages are informative
- Commit titles have module prefix
- Commit titles have issue nr. suffix

Issue #4243, opened 2025-02-06
Disable tablet balancing during restore


Michal-Leszczynski:
As discussed in https://github.com/scylladb/scylladb/issues/22707, until load&stream is synchronized with tablet balancing on Scylla side, we need to disable it on the SM side. Otherwise, it's possible that a node streams sstable to another node which no longer owns it because of the tablet migration running in parallel.

karol-kokoszka:
Connected to https://github.com/scylladb/scylla-manager/issues/4275.
The issue above, must be clarified first.

Michal-Leszczynski:
@bhalevy @mitso23 closing this as not planned, as I understand that we don't want to do it for neither native tablet restore and tablet aware restore.

PR #4340, opened 2025-04-02
Update tested scylla versions

### LINKED ISSUE #4334: Test TestClientActiveRepairsIntegration is flaky
Example flake: https://github.com/scylladb/scylla-manager/actions/runs/14192086043/job/39758853118

Michal-Leszczynski:
This PR updates tested Scylla versions and also changes tested configurations.
Among other things, it started testing 2025.1 with tablets and IPV6.
This combination discovered a problem with IPV6 parsing, so this PR also partially deals with #4342, but it does not fixes it completely. It also addresses some flakiness in the tests (see commit descriptions).

It also added a new package called util2 in place of the original util package which migrated to a separate submodule. The migration of util was a mistake, so in order not to waist time on creating multiple PRs with submodule changes and verion bumps, I decided to add new functionality, where it is supposed to be. After #4121 is fixed, util and util2 should be merged into a single package. 

The failure in restore tests is caused by #4243, which needs to be addressed quickly.

Fixes #4334
Supersedes #4336.

### DIFF
--- a/pkg/scyllaclient/client_scylla.go
+++ b/pkg/scyllaclient/client_scylla.go
@@ -9,8 +9,10 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"net/netip"
 	"regexp"
 	"runtime"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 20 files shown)


pkg/util2/slices/slices.go
```diff
@@ -0,0 +1,38 @@
+// Copyright (C) 2025 ScyllaDB
+
+package slices
```
karol-kokoszka, line 3: Could you please unit-test this lib ? 
With some basic / not extensive tests.

pkg/util2/maps/maps.go
```diff
@@ -0,0 +1,47 @@
+// Copyright (C) 2025 ScyllaDB
+
+package maps
```
karol-kokoszka, line 3: Could you please unit-test this lib ? 
With some basic / not extensive tests.

karol-kokoszka (commented):
@Michal-Leszczynski Just very general question. WDYT about changing the client for Scylla the way that "Addr" is always of netip.Addr type instead of string ? I understand that swagger defines it as string, but we are not generating client out of the swagger.

/pkg/scyllaclient/client_scylla.go is our wrapper over swagger generated code.

VAveryanov8 (approved):
:+1:

karol-kokoszka (approved):
:+1:

PR #4357, opened 2025-04-23
TMP: scylla api

Michal-Leszczynski:
TMP


Michal-Leszczynski:
@karol-kokoszka it would be good to merge native scylla API support to master in some time.
This PR contains all the changes from `ml/scylla-api` + changes around new scylla object storage endpoints configuration.
This PR is ready except for failing ipv6 tests which for now are assumed to fail because of scylla inability to handle ipv6 object storage endpoints. 
The question is whether we should first merge this PR and later fix https://github.com/scylladb/scylla-manager/issues/4163, or is adding the API switch a prerequisite for merging those changes to master.

PR #4360, opened 2025-04-25
docs: update dependencies

dgarcia360:
This is a mandatory dependency update to resolve a critical Dependabot alert. For more details, see the [Dependabot alerts](https://docs.github.com/en/code-security/dependabot/dependabot-alerts/viewing-and-updating-dependabot-alerts).

---

Please make sure that:


Michal-Leszczynski:
@dgarcia360 why are we updating just the `poetry.lock` file?
It might or might not be a good idea to commit it to the repo in general, but we should also update the dependencies which generate it, right? So shouldn't this PR also update the `pyproject.toml`?

dgarcia360:
@Michal-Leszczynski

> It might or might not be a good idea to commit it to the repo in general

https://python-poetry.org/docs/basic-usage/#as-an-application-developer

> but we should also update the dependencies which generate it, right? So shouldn't this PR also update the pyproject.toml?

Not necessarily. In this case, we’re updating transitive dependencies (those that are required by the dependencies listed in `pyproject.toml`). For example, the theme we use depends on a version range of the `sphinx-notfound-page` package. 

Running `make update` refreshes the `poetry.lock` file to use the latest compatible versions within those defined ranges.

Michal-Leszczynski (approved):

PR #4361, opened 2025-04-28
fix(testing): fix testing setup of prometheus

VAveryanov8:
This adds scrape config for scylla-manager service itself.

---

Please make sure that:

### DIFF
--- a/testing/prometheus/prometheus-ipv6.yaml
+++ b/testing/prometheus/prometheus-ipv6.yaml
@@ -7,3 +7,6 @@ scrape_configs:
       - server: '2001:0DB9:100::2'
         scheme: 'http'
         services: ['scylla']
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
> It works for the IPv4 (make sure you've restarted container with prometheus), but I wasn't able to verify it against IPv6 and that's why this PR is in draft

Could you explain how does it work with IPV4? I tried to populate metrics by `make pkg-integration-test` and manually starting SM server, adding cluster, executing repair, but none of those approaches populated anything for the prometheus server at `192.168.100.98:9090`. When starting SM server manually, I am able to scrape its metrics manually at `localhost:5090/metrics`.

Michal-Leszczynski:
okay, I don't think I ever used `make run-server`, so it makes sense now, thanks!

Issue #4254, opened 2025-02-13
Validate if there are any tablet based keyspaces


karol-kokoszka:
... and fail the process if there are any, and they are not filtered out by the "--keyspaces" flag.

VAveryanov8:
The way I see how we can check if keyspace is tablet based or not is following: 
1. Check the value of scylla.yaml `enable_tablets`. https://opensource.docs.scylladb.com/stable/reference#confval-enable_tablets 
2. If it is false - then there is not way to create tablet based keyspace - we good to go with 1-1-restore. 
3. if it's set to true, every keyspace by default is tablet based, but there is an option to opt out - `CREATE KEYSPACE .... WITH tablets = {'enabled': false}  `, so we need to see how the keyspace was created. Unfortunately `system_schema.keyspace` and `scylladb/gocql` doesn't expose this option, so the only way I found is to check (parse) the output of `DESCRIBE SCHEMA WITH INTERNALS`. 

@Michal-Leszczynski @karol-kokoszka what do you think? Maybe you know the better way of checking if keyspace is tablet based or not?

Michal-Leszczynski:
When running the `1-1-restore` we already have the required schema in the cluster, so we can just query Scylla API for the information about table replication type:
https://github.com/scylladb/scylla-manager/blob/af2379317f84faf9304959cf20457f4864b4ec99/pkg/scyllaclient/client_scylla.go#L268-L306

VAveryanov8:
> When running the 1-1-restore we already have the required schema in the cluster, so we can just query Scylla API for the information about table replication type:

Thanks, that's exactly what i was searching for! :) 

> But in general we have a problem with recreating keyspace replication type from DESCRIBE SCHEMA WITH INTERNALS.
> See https://github.com/scylladb/scylladb/issues/22866 for more info.
> This issue could happen only when we use different tablet configuration in scylla.yaml in the backup and restore destination cluster.

good finding!

karol-kokoszka:
Please add this check to the validation stage, where SM verifies the source-destination cluster topology.

PR #4365, opened 2025-04-29
feat(1-1-restore): fail validation if tablet-based keyspaces are present

VAveryanov8:
This adds additional validation to check that there are no tablet based keyspaces in the cluster, because 1-1-restore can work only with vnodes.

Refs: #4254

---

Please make sure that:

### DIFF
--- a/pkg/service/one2onerestore/worker.go
+++ b/pkg/service/one2onerestore/worker.go
@@ -226,6 +226,8 @@ func skipRestorePatterns(ctx context.Context, client *scyllaclient.Client, sessi
 			skip = append(skip, ks)
 		}
 	}
+	// See https://github.com/scylladb/scylla-enterprise/issues/4168
+	skip = append(skip, "system_replicated_keys")
 
 	// Skip outdated tables.
 	// Note that even though system_auth is not used in Scylla 6.0,
--- a/pkg/service/one2onerestore/worker.go
+++ b/pkg/service/one2onerestore/worker.go
@@ -15,6 +15,7 @@ import (
 	"github.com/scylladb/gocqlx/v2"
 	"github.com/scylladb/scylla-manager/backupspec"
 	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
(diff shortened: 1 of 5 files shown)


VAveryanov8:
Tests with `tablets: none` are failing and it's hard to do something with them (other than to ignore), because this PR relies on tablets support.

I think we can simply ignore them,because it looks like they are no longer exists in the master branch :)

karol-kokoszka (approved):

pkg/service/one2onerestore/worker.go
```diff
@@ -305,3 +314,17 @@ func isRestoreAuthAndServiceLevelsFromSStablesSupported(ctx context.Context, cli
...
+	if err != nil {
+		return errors.Wrap(err, "new keyspace filter")
+	}
+	for _, ks := range tabletKeyspaces {
+		if filter.Check(ks, "") {
+			return errors.Errorf("1-1-restore doesn't support tablet based replication. Keyspace: %s", ks)
+		}
+	}
+	return nil
+}
```
Michal-Leszczynski, line 332: 1-1 keyspace filter is just a list of keyspace without any patters (except for a simple `*`). I would consider using simple `slices.Contains` func in order not to confuse the reader with the keyspace filters used in other SM tasks.

Michal-Leszczynski:
> Tests with tablets: none are failing and it's hard to do something with them (other than to ignore), because this PR relies on tablets support.
I think we can simply ignore them,because it looks like they are no longer exists in the master branch :)

Can't we just `t.Skip("explanation")` them when tablets aren't supported?

pkg/service/one2onerestore/worker.go
VAveryanov8: When this function is invoked keyspace filter also includes 'skip list' which actually have some patterns :) 

https://github.com/scylladb/scylla-manager/pull/4365/files#diff-2cea9c21226ddc87135b86e914571186f4d82e6eeb0bd0a304df4ceb0e1e4b1aR52-R57

PR #4366, opened 2025-04-29
fix(repair_test): decrease node start flakiness

Michal-Leszczynski:
It's important to wait for all nodes to perceive all nodes as live before moving forward with the test.

Example flakiness: https://github.com/scylladb/scylla-manager/actions/runs/14728318840/job/41336270422?pr=4364


VAveryanov8 (approved):
:rocket:

PR #4376, opened 2025-05-06
feat(sctool): support for 1-1-restore progress rendering

VAveryanov8:
This adds support of 1-1-restore progress to `sctool progress`.

---

Please make sure that:

### DIFF
--- a/pkg/command/progress/cmd.go
+++ b/pkg/command/progress/cmd.go
@@ -68,6 +68,7 @@ var supportedTaskTypes = strset.New(
 	managerclient.RestoreTask,
 	managerclient.RepairTask,
 	managerclient.ValidateBackupTask,
+	managerclient.One2OneRestoreTask,
 )
 
 func (cmd *command) run(args []string) error {
@@ -99,6 +100,8 @@ func (cmd *command) run(args []string) error {
 		return cmd.renderRestoreProgress(task)
 	case managerclient.ValidateBackupTask:
 		return cmd.renderValidateBackupProgress(task)
+	case managerclient.One2OneRestoreTask:
+		return cmd.renderOne2OneRestoreProgress(task)
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -30,10 +30,10 @@ require (
 	github.com/scylladb/go-reflectx v1.0.1
 	github.com/scylladb/go-set v1.0.2
 	github.com/scylladb/gocqlx/v2 v2.8.0
-	github.com/scylladb/scylla-manager/backupspec v1.0.1
-	github.com/scylladb/scylla-manager/v3/pkg/managerclient v0.0.0-20250414145630-57597921df36
+	github.com/scylladb/scylla-manager/backupspec v1.0.2
+	github.com/scylladb/scylla-manager/v3/pkg/managerclient v0.0.0-20250506140140-b1b310d3a2e7
 	github.com/scylladb/scylla-manager/v3/pkg/util v0.0.0-20250310104639-18257b71bb48
-	github.com/scylladb/scylla-manager/v3/swagger v0.0.0-20250310104639-18257b71bb48
+	github.com/scylladb/scylla-manager/v3/swagger v0.0.0-20250502123045-ee6af14e22db
```
Michal-Leszczynski, line 36: Can we bump the version to the same newest commit from master for all submodules?
I know that they haven't been changed, but it makes it easier to avoid problems with strange version mismatches.

VAveryanov8: the problem is that the sub modules (managerclient in particular) are not in master yet, but only in feature branch :)

Michal-Leszczynski: ah yes......
So the merge to master will need to be partitioned into merge per changed submodule and corresponding version bumps, right?

VAveryanov8: yes, we can't simply merge feature branch into master, so probably it will be done in chunks of commits (separate PR!?) :)

Michal-Leszczynski (approved):

Issue #4375, opened 2025-05-05
Provide "unpin-agent-cpu" flag to 1-1 restore


karol-kokoszka:
Expected exactly the same behavior as in L&S restore - https://manager.docs.scylladb.com/stable/sctool/restore.html#unpin-agent-cpu

It's part of speeding up the download time effort.

PR #4380, opened 2025-05-12
feat(1-1-restore): add unpin-agent-cpu flag 

VAveryanov8:
This adds support of unpin-agent-cpu flag to 1-1-restore command.
Unpinning agent CPU should help with agent download speed.

Refs: https://github.com/scylladb/scylla-manager/issues/4375

---

Please make sure that:

### DIFF
--- a/pkg/command/one2onerestore/res.yaml
+++ b/pkg/command/one2onerestore/res.yaml
@@ -31,4 +31,3 @@ dry-run: |
 
 unpin-agent-cpu: |
   Defines if ScyllaDB Manager Agent should be unpinned from CPUs during restore.
-  This might significantly improve download speed at the cost of decreasing streaming speed. 
--- a/pkg/command/one2onerestore/res.yaml
+++ b/pkg/command/one2onerestore/res.yaml
@@ -28,3 +28,7 @@ nodes-mapping:
 dry-run: |
   Validates and displays restore information without actually running the restore.
   This allows you to display what will happen should the restore run with the parameters you set.
+
+unpin-agent-cpu: |
+  Defines if ScyllaDB Manager Agent should be unpinned from CPUs during restore.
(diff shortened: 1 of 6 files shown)


pkg/command/one2onerestore/res.yaml
```diff
@@ -28,3 +28,7 @@ nodes-mapping:
 dry-run: |
   Validates and displays restore information without actually running the restore.
   This allows you to display what will happen should the restore run with the parameters you set.
+
+unpin-agent-cpu: |
+  Defines if ScyllaDB Manager Agent should be unpinned from CPUs during restore.
+  This might significantly improve download speed at the cost of decreasing streaming speed.
```
Michal-Leszczynski, line 34: Streaming speed in 1-1-restore?

VAveryanov8: I thought it will affect load&stream, because there will be less cpu power for it? Maybe the exact wording is not correct and it's better to either delete it completely or just mention sstable loading.

Michal-Leszczynski: But we don't have load&stream in 1-1-restore, right?

VAveryanov8: you right! My question is more about load sstables api call in the 1-1-restore context - I'm genuinely curios if it's gonna be slower or not when agent cpu is unpinned? And if yes, is it worth to mentioned that in the flag description

Michal-Leszczynski: I guess we should benchmark it.
It might be that there is no drawback to setting this flag, so we should always do it and maybe not even expose a flag for controlling this behavior.

VAveryanov8: make sense!
anyway this "`at the cost of decreasing streaming speed`" is not needed here, so I'm gonna delete it.

PR #4381, opened 2025-05-12
Merge native backup

### LINKED ISSUE #4138: Handle SM backup task pause/resume with Scylla API

### LINKED ISSUE #4141: Monitor Scylla backup API progress

### LINKED ISSUE #4143: Replace `/rclone/sync/movedir` with `/storage_service/backup` to use scylla streaming

### LINKED ISSUE #4384: Set Scylla task TTL for native backup

### LINKED ISSUE #4386: Add api switch for native backup

### LINKED ISSUE #4388: Upload all tables in parallel with native backup API
SM uploads tables sequentially in the context of a single node.
Uploading all tables at the same time with rclone puts too much pressure on the agent/node.

There is no such limitation with the native backup api.
Uploading all tables in parallel should even make it possible to achieve better controlled performance.
Moreover, it enables scylla to prioritize uploading snapshot-ed sstables which have already been removed from the data dir, which should allow to decrease disk space utilization faster. 

Note that it's still possible that some tables are uploaded with rclone and some with native backup api (e.g. when the upload results in creation of versioned files - even though it is a really rare case). In such cases, SM will first upload in parallel all tables which can be uploaded with native backup api, and then upload the remaining  tables sequentially with rclone api.

Michal-Leszczynski:
This PR contains subset of changes from `ml/scylla-api-2` needed in order to support native backup api.
The main change comes with replacing rclone `sync/movedir` api with native scylla `storage_service/backup` api when uploading snapshot-ed files into s3. This PR also makes some changes regarding configuring SM test env to support the native scylla backup api. The most interesting part from user perspective is the new `sctool backup --method` flag (previously `--api-hint`), which is documented as below:
```
   - name: method
      default_value: auto
      usage: |-
        Specify API used for uploading files.
        This flag can be set to either:
        - 'auto' means that native backup API will be used when possible, otherwise Rclone API will be used.
        - 'rclone' means that only Rclone API will be used.
        - 'native' means that only native API will be used.
        ScyllaDB Manager can use either Rclone API or native Scylla API for uploading snapshot-ed files.
        In both cases, it requires to configure ScyllaDB Manager Agent 'scylla-manager-agent.yaml' config.
        In case of native backup API, it also requires to configure 'object_storage_endpoints' in 'scylla.yaml'.
```

Initially it was supposed to be set to `rclone` by default, but I decided to go with `auto` because:
- we still want to have it set to `auto` before release
- it increases the scope of native backup api tests
- changing the default value is not just changing a single string - it also requires changes in the tests, which takes some minor time

Fixes #4143
Fixes #4138
Fixes #4141
Fixes #4384
Fixes #4386
Fixes #4388


Michal-Leszczynski:
Build: https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-build/986/artifact/00-Build.txt

Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 Please take a look. This is important in terms of leveraging the new native backup api from Scylla 2025.2.

VAveryanov8 (approved):
LGTM

pkg/service/backup/worker_upload.go
```diff
@@ -33,88 +33,115 @@ func (w *worker) Upload(ctx context.Context, hosts []hostInfo, limits []DCLimit)
...
+	// NOTE that defers are executed in LIFO order
+	// Abort on cancel.
+	defer func() {
+		if scheduler.IsTaskInterrupted(ctx) {
+			err = parallel.Abort(err)
+		}
+	}()
+	// Add keyspace table info to error mgs.
+	defer func() {
+		err = errors.Wrapf(err, "%s.%s", d.Keyspace, d.Table)
```
VAveryanov8, line 134: nit: I think keyspace and table will be already included into log message as per err handler: 
```go
	w.Logger.Error(ctx, "Failed to upload host",
				"host", d.Host,
				"keyspace", d.Keyspace,
				"table", d.Table,
				"error", err,
			)
```

pkg/service/backup/worker_upload.go
```diff
@@ -33,88 +33,115 @@ func (w *worker) Upload(ctx context.Context, hosts []hostInfo, limits []DCLimit)
...
+	// Skip snapshots that are already uploaded.
+	if d.Progress.IsUploaded() {
+		w.Logger.Info(ctx, "Snapshot already uploaded skipping", "host", h.IP, "keyspace", d.Keyspace, "table", d.Table)
+		return nil
 	}
 
-	return parallel.Run(len(dirs), 1, f, notify)
+	// NOTE that defers are executed in LIFO order
+	// Abort on cancel.
+	defer func() {
```
VAveryanov8, line 127: nit: IMHO one defer would be enough here :) 
```go
defer func() {
  if scheduler.IsTaskInterrupted(ctx) {
			err = parallel.Abort(err)
      return
	 }
   if err != nil {
       return 
   }
   // cleanup
   err = errors.Wrap(w.deleteTableSnapshot(ctx, h, d), "delete table snapshot")
}()
```

VAveryanov8 (approved):
:+1:

Michal-Leszczynski:
@karol-kokoszka please take a look!

docs/source/sctool/partials/sctool_backup_update.yaml
```diff
@@ -88,6 +88,17 @@ options:
...
+      default_value: auto
+      usage: |-
+        Specify API used for uploading files.
+        This flag can be set to either:
+        - 'auto' means that native backup API will be used when possible, otherwise Rclone API will be used.
+        - 'rclone' means that only Rclone API will be used.
+        - 'native' means that only native API will be used.
+        ScyllaDB Manager can use either Rclone API or native Scylla API for uploading snapshot-ed files.
+        In both cases, it requires to configure ScyllaDB Manager Agent 'scylla-manager-agent.yaml' config.
+        In case of native backup API, it also requires to configure 'object_storage_endpoints' in 'scylla.yaml'.
```
karol-kokoszka, line 101: Let's explain the difference between native API and Rclone API (a.k.a. previous comment).

pkg/cmd/agent/root.go
```diff
@@ -65,6 +65,11 @@ var rootCmd = &cobra.Command{
 			return errors.New("configuration file must specify auth-token")
 		}
 
+		// Fill provider config
+		c.S3.AutoFill()
```
karol-kokoszka, line 69: I'm thinking about this part in the context of adding AutoFill to the `/node_info` endpoint.
If the intention is to fill the region value of config, so that AutoFill() doesn't need to read the metadataAPI again, then it's wrong, because AutoFill() just logs on errors. If the provider is not AWS, then the region will never be set and it always need to query the metadata again.

pkg/command/backup/cmd.go
```diff
@@ -163,7 +165,11 @@ func (cmd *command) run(args []string) error {
 		ok = true
 	}
 	if cmd.Flag("skip-schema").Changed {
-		props["skip_schema"] = cmd.purgeOnly
+		props["skip_schema"] = cmd.skipSchema
```
karol-kokoszka, line 168: I think it was not possible to change the skip_schema to ture, as purgeOnly is true only if someone just to explicitly clean expired snapshots.
Nice finding.

pkg/schema/nopmigrate/nopmigrate.go
```diff
@@ -18,7 +22,159 @@ func nopCallback(context.Context, gocqlx.Session, migrate.CallbackEvent, string)
...
+	tmpMetadata := dstMetadata
+	tmpMetadata.Name = tmpName
+	tmp := table.New(tmpMetadata)
+
+	if err := session.ExecStmt(fmt.Sprintf(createDstStmt, tmpName)); err != nil {
+		return errors.Wrap(err, "create tmp_restore_run_progress")
+	}
+
+	err := dbutil.RewriteTable(session, tmp, src, func(m map[string]interface{}) {
+		m["remote_sstable_dir"] = m["manifest_path"]
```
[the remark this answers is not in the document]
karol-kokoszka: Why it it needed for native backup ? It's just a cosmetic change ? I mean, to keep more relevant naming ? Does the new backup keep the current specification and backup dir hierarchy ?

pkg/service/backup/backup.go
```diff
@@ -45,6 +49,16 @@ func makeHostInfo(nodes []scyllaclient.NodeStatusInfo, locations []backupspec.Lo
 			hi[i].RateLimit = dcr[""] // no rate limit is ok, fallback to 0 - no limit
 		}
 		hi[i].Transfers = transfers
+
+		ip, err := netip.ParseAddr(h.Addr)
```
karol-kokoszka, line 53: I saw that you are assuring that the ip is in form of netip.ParseAddr when calling to makeHostInfo from the service. Why it's duplicated here as well ? 
Other method users are not applying this function ?

karol-kokoszka:
`integration-tests-latest` are validating the native backup right ?

karol-kokoszka:
@Michal-Leszczynski Overall, looks very good.
Please resolve the comment about unnecessary calls to the AWS instance metadata endpoint.

pkg/cmd/agent/nodeinfo.go
```diff
@@ -63,6 +66,16 @@ func (h *nodeInfoHandler) versionInfo(ctx context.Context, info *scyllaclient.No
 	return nil
 }
 
+func (h *nodeInfoHandler) rcloneProviderConfig(info *scyllaclient.NodeInfo) {
+	h.config.S3.AutoFill()
```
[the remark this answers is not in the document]
Michal-Leszczynski: > This code is problematic, because it always probes the AWS metadata URL. Regardless if the node is deployed to AWS or not.

This code is not even needed, I just put it there for safety, so that we can assume that the s3 region is always filled if we are deployed in s3. You are right that it should be deleted.

pkg/cmd/agent/root.go
Michal-Leszczynski: Right, so we can delete the `AutoFill` from `/node_info` endpoint.
Note that we are not really concerned with problems like "what if we are deployed in aws but AutoFill failed because of unrelated reasons?", because in such cases rclone server would be incorrectly initialized for the s3 backend and it wouldn't work anyway.

Michal-Leszczynski:
> integration-tests-latest are validating the native backup right ?

Yes, but I will also work on #4393 (in a separate PR) to make it more visible.

Michal-Leszczynski:
Added 3 commits that needs to be squashed before merge:
- [s: update docs](https://github.com/scylladb/scylla-manager/pull/4381/commits/6c81500f38d3cab7ac17632798efe7d65aecf83f)
- [s: remove unnecessary AutoFill calls](https://github.com/scylladb/scylla-manager/pull/4381/commits/6a32b40ba6a8fbb082b2b38106daea43e70d40d3)
- [s: improve handling of empty s3 region](https://github.com/scylladb/scylla-manager/pull/4381/commits/3f5927c2c02b4fc466424f760246d3e6d614a101)

@karol-kokoszka please take a look

karol-kokoszka (approved):
:+1: 
Please squash and merge. It's ok to to have more than 1 commit, but don't leave micro commits, rather rely on features.
Commits like:
```
[feat(restore): adjust model to recreated restore_run_progress]
[refactor(scyllaclient): remove unused ToCanonicalIP]
[feat(sstable): describe sstable specification]
```

Are not informative without the context. Linking the issue is not enough.

Issue #4373, opened 2025-05-05
Provide bandwidth metrics for 1-1-restore (similarly to regular L&S restore)


mikliapko:
Regular L&S restore provides bandwidth characteristics of the process in MiB/s/shard.

Thus, we are able to compare restore process speed across different backup sizes and instance types.

To follow the same goal, it'd be nice to have the same metrics for 1-1-restore.

Michal-Leszczynski:
@mikliapko is this a dup of #4204?

PR #4383, opened 2025-05-13
feat(1-1-restore): add download bandwidth to logs

VAveryanov8:
This introduces restoreStats struct which can be used to collect various statistics during 1-1-restore run.
For now main use case is to provide avg download bandwidth in the same format as regular restore task progress, so SCT (e2e tests) can use it to compare regular and 1-1 restores.
Also this adds bandwidth per table in bytes/ms to logs which can be useful for investigative purposes.

Here is how appropriate log will looks like: 
```text
11:18:09.464    INFO    one2onerestore  Restore stats   {"bandwidth_per_shard": "10.540KiB/s/shard", "_trace_id": "SPYCOXJSSP6bxztF_Rh2IQ"}

```

Refs: #4373

---

Please make sure that:

### DIFF
--- a/pkg/service/one2onerestore/stats.go
+++ b/pkg/service/one2onerestore/stats.go
@@ -1,0 +1,45 @@
+// Copyright (C) 2025 ScyllaDB
+
+package one2onerestore
+
+import (
+	"sync/atomic"
+
+	"github.com/scylladb/scylla-manager/v3/pkg/util/sizesuffix"
+)
+
+// restoreStats is used to collect some run statistics.
+type restoreStats struct {
+	totalShards         int64
+	totalDownloadTimeMS int64
+	totalDownloadBytes  int64
+}
+
+func newRestoreStats(workload []hostWorkload) *restoreStats {
+	var sumShards int
+	for _, w := range workload {
+		sumShards += w.host.ShardCount
+	}
+	return &restoreStats{
+		totalShards: int64(sumShards),
+	}
+}
+
+// incrementDownloadStats atomically increment downloaded bytes and duration in milliseconds.
+func (rs *restoreStats) incrementDownloadStats(downloadedBytes, tookMS int64) {
+	atomic.AddInt64(&rs.totalDownloadBytes, downloadedBytes)
+	atomic.AddInt64(&rs.totalDownloadTimeMS, tookMS)
+}
+
+// averageBandwidthPerShard provides download bandwidth in the same
+// format as `sctool progress` for regular restore task.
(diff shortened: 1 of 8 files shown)


pkg/service/one2onerestore/stats.go
```diff
@@ -0,0 +1,45 @@
...
+
+// incrementDownloadStats atomically increment downloaded bytes and duration in milliseconds.
+func (rs *restoreStats) incrementDownloadStats(downloadedBytes, tookMS int64) {
+	atomic.AddInt64(&rs.totalDownloadBytes, downloadedBytes)
+	atomic.AddInt64(&rs.totalDownloadTimeMS, tookMS)
+}
+
+// averageBandwidthPerShard provides download bandwidth in the same
+// format as `sctool progress` for regular restore task.
+func (rs restoreStats) averageBandwidthPerShard() string {
```
Michal-Leszczynski, line 36: I guess it's cleaner to have all methods as pointer receivers ([link](https://go.dev/doc/faq#methods_on_values_or_pointers)).

pkg/service/one2onerestore/stats.go
```diff
@@ -0,0 +1,45 @@
+// Copyright (C) 2025 ScyllaDB
+
+package one2onerestore
+
+import (
+	"sync/atomic"
+
+	"github.com/scylladb/scylla-manager/v3/pkg/managerclient"
```
Michal-Leszczynski, line 8: There shouldn't be a reason why `one2onerestore` pkg should import `managerclient`. I see that you use it for parsing bandwidth, but since we are only logging it, wouldn't it be enough to just always print it as mb/s and just round it to something like 4 decimal points?

VAveryanov8: @Michal-Leszczynski I can copy `FormatSuffixSize` into a separate package, e.g.  `pkg/utils2/size` and use it here, so we will not be dependent on managerclient. (copy and not move because managerclient is a separate go module). What do you think? 

P.S.
FormatSuffixSize is actually already a copy from rclone :)

VAveryanov8: yeah, indeed, we have v3/pkg/util module which is already used in managerclient modules  :) 
so, i can move it there, but yes, it's gonna take a several steps/PRs.

Michal-Leszczynski: So it's up to you, I'm also fine with copying.

VAveryanov8: done

Michal-Leszczynski (approved):

PR #4390, opened 2025-05-16
feat(1-1-restore): add queue for a refreshNode calls

VAveryanov8:
This adds a queue for refreshNode calls to avoid blocking download operation.

---

Please make sure that:

### DIFF
--- a/pkg/service/one2onerestore/worker_tables.go
+++ b/pkg/service/one2onerestore/worker_tables.go
@@ -11,6 +11,8 @@ import (
 	"github.com/scylladb/scylla-manager/v3/pkg/metrics"
 	"github.com/scylladb/scylla-manager/v3/pkg/scyllaclient"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/parallel"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/timeutc"
+	"golang.org/x/sync/errgroup"
 )
 
 func (w *worker) restoreTables(ctx context.Context, workload []hostWorkload, keyspaces []string) error {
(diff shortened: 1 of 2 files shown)


karol-kokoszka:
All restore integration-tests are failing.
Did you have a chance to check the reason ? Seems the count of rows doesn't match after the restore.

Michal-Leszczynski:
@VAveryanov8 maybe it's worth to rebase feature branch on top of master? Or has it already been done?
Anyhow, it shouldn't create conflicts with the existing PRs, right?

pkg/service/one2onerestore/worker_tables.go
```diff
@@ -20,28 +22,18 @@ func (w *worker) restoreTables(ctx context.Context, workload []hostWorkload, key
...
-			}
-			return nil
-		}); err != nil {
+		// Start refresh node worker to avoid blocking on refreshNode call.
+		refreshGroup.Go(w.refreshNodeWorker(ctx, refreshQueue))
+		if err := w.downloadTablesByHost(ctx, hostTask, keyspaces, refreshQueue); err != nil {
+			w.metrics.SetOne2OneRestoreState(w.runInfo.ClusterID, manifestInfo.Location, manifestInfo.SnapshotTag, host.Addr, metrics.One2OneRestoreStateError)
+			return err
+		}
+		if err := refreshGroup.Wait(); err != nil {
```
Michal-Leszczynski, line 36: Right now the design is that the refresh worker works in a separate goroutine within errgroup and the download worker is running in the scope of the function execution. I would say that a cleaner design would be to run both workers in goroutines within errgroup and use the function scope for waiting for and bookkeeping the results.

One of the reasons is that right now the following scenario is possible:
- download worker downloaded 100 tables
- refresh worker is slower - it refreshed just 25 tables
- download worker encountered an error
- we exit from the function scope without waiting for the refresh worker to finish (detached goroutine)
- refresh worker will keep on refreshing the remaining 75 tables which will take a long time

It would be easier to control with download worker working within the errgroup.

Michal-Leszczynski: About the last dot from the scenario: if the download worker encounters an error, it closes the job chan and ends immediately. The refresh worker will still process whatever is left on the job chan before leaving. I'm just wondering if it's worth to wait for the refresh worker to finish all of its jobs before ending the 1-1-restore with an error. Maybe it would make more sense to notify the refresh worker that it should stop picking up new jobs right after the download worker failed. What do you think?

pkg/service/one2onerestore/worker_tables.go
```diff
@@ -20,28 +22,18 @@ func (w *worker) restoreTables(ctx context.Context, workload []hostWorkload, key
 	return parallel.Run(len(workload), len(workload), func(i int) error {
 		hostTask := workload[i]
 		manifestInfo, host := hostTask.manifestInfo, hostTask.host
-		const (
-			repeatInterval  = 10 * time.Second
-			pollIntervalSec = 10
+		var (
+			refreshQueueSize = max(100, len(hostTask.manifestContent.Index))
```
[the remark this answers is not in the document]
Michal-Leszczynski: On the other hand, if we think that worker compete for the same resources (disk), it might make sense to use `min` here with a lower number. In case the download is way faster than refresh, downloading 100 in advance shouldn't speed up the overall 1-1-restore time. If we are concerned that the download might actually slow down the refresh, then it would make sense to download only a few tables in advance, so that the refresh can work faster.

This is just a general speculation. On the other hand, can using `min` and lower number decrease 1-1-restore performance? If not, then maybe it's worth to try.

pkg/service/one2onerestore/worker_tables.go
```diff
@@ -61,10 +83,43 @@ func (w *worker) createDownloadJob(ctx context.Context, table backupspec.FilesMe
...
+			}
+		}
+	}
+}
+
 func (w *worker) refreshNode(ctx context.Context, table backupspec.FilesMeta, m *backupspec.ManifestInfo, h Host, pr *RunTableProgress) error {
+	start := timeutc.Now()
 	w.metrics.SetOne2OneRestoreState(w.runInfo.ClusterID, m.Location, m.SnapshotTag, h.Addr, metrics.One2OneRestoreStateLoading)
 	err := w.client.AwaitLoadSSTables(ctx, h.Addr, table.Keyspace, table.Table, false, false, true)
 	w.finishDownloadProgress(ctx, pr, err)
```
Michal-Leszczynski, line 115: Why do we run `finishDownloadProgress` after loading the tables to the cluster.
Shouldn't it be called right after the download has finished?

pkg/service/one2onerestore/worker_tables.go
Michal-Leszczynski: It could be done by not only closing, but also draining the job chan - if we want to go into that direction.

pkg/service/one2onerestore/worker_tables.go
```diff
@@ -50,6 +42,36 @@ func (w *worker) restoreTables(ctx context.Context, workload []hostWorkload, key
...
+	return hostTask.manifestContent.ForEachIndexIterWithError(keyspaces, func(table backupspec.FilesMeta) error {
+		w.logger.Info(ctx, "Restoring data", "ks", table.Keyspace, "table", table.Table, "size", table.Size)
+
+		jobID, err := w.createDownloadJob(ctx, table, manifestInfo, host)
+		if err != nil {
+			return errors.Wrapf(err, "create download job: %s.%s", table.Keyspace, table.Table)
+		}
+		pr := w.downloadProgress(ctx, hostTask.host.Addr, table)
+
+		if err := w.waitJob(ctx, jobID, manifestInfo, host, pr, pollIntervalSec); err != nil {
```
Michal-Leszczynski, line 61: `waitJob` implementation is not a part of this PR, but it's always safer to add default case in so that we don't accidentally get stuck in an endless loop:
```go
		switch scyllaclient.RcloneJobStatus(job.Status) {
		case scyllaclient.JobError:
			return errors.Errorf("job error (%d): %s: host %s", jobID, job.Error, h.Addr)
		case scyllaclient.JobSuccess:
			w.logger.Info(ctx, "Job done",
				"job_id", jobID,
				"host", h,
				"took", time.Time(job.CompletedAt).Sub(time.Time(job.StartedAt)),
			)
			return nil
		case scyllaclient.JobRunning:
			continue
		case scyllaclient.JobNotFound:
			return errors.New("job not found")
		}
```

pkg/service/one2onerestore/worker_tables.go
```diff
@@ -50,6 +42,36 @@ func (w *worker) restoreTables(ctx context.Context, workload []hostWorkload, key
...
+			ManifestInfo: manifestInfo,
+			Progress:     pr,
+		}
+		return nil
+	})
+}
+
 func (w *worker) createDownloadJob(ctx context.Context, table backupspec.FilesMeta, m *backupspec.ManifestInfo, h Host) (int64, error) {
 	uploadDir := backupspec.UploadTableDir(table.Keyspace, table.Table, table.Version)
 	remoteDir := m.LocationSSTableVersionDir(table.Keyspace, table.Table, table.Version)
```
Michal-Leszczynski, line 77: Not a part of this PR, but it seems important. One line below we have:
```go
	jobID, err := w.client.RcloneCopyPaths(ctx, h.Addr, scyllaclient.TransfersFromConfig, scyllaclient.NoRateLimit, uploadDir, remoteDir, table.Files)
```
So in the regular restore we have the `--transfers` flag, because it is supposed to be more versatile (e.g. restoring into a running cluster). I would say that here it's all about speed, so we shouldn't use `cyllaclient.TransfersFromConfig` but do what is done for `--transfers=0`:
```
      --transfers int                         Sets the amount of file transfers to run in parallel when downloading files from backup location to Scylla node.
                                              Set to 0 for the fastest download (results in setting transfers to 2*node_shard_count).
                                              Set to -1 for using the transfers value defined in node's 'scylla-manager-agent.yaml' config file.

```

It might explain differences in download speed between regular and 1-1 restore (@mikliapko were the regular restore benchmarks running with `--transfers=0`?).

Michal-Leszczynski (commented):
Some comments exceed the scope of this PR, so they can be packed into issues and addressed in different PRs.

pkg/service/one2onerestore/worker_tables.go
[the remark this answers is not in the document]
Michal-Leszczynski: Another consideration: wouldn't it make sense to restore tables in ascending size order? The problem with no batching is that if we start with a really big table, then the refresh worker will be idle for quite some time.

pkg/service/one2onerestore/worker_tables.go
VAveryanov8: It's because table progress also includes `is_refreshed` flag and from the restore progress point of view we can't say that the table is restored until it's refreshed. 
So I guess the func name is problematic here, I think I need to come up with something more meaningful :)

pkg/service/one2onerestore/worker_tables.go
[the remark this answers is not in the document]
VAveryanov8: > Another consideration: wouldn't it make sense to restore tables in ascending size order? The problem with no batching is that if we start with a really big table, then the refresh worker will be idle for quite some time.

It looks like that in order to read tables in ascending size order we will need to load all tables in memory (e.g [ReadIndex](https://github.com/scylladb/scylla-manager/blob/455acc4d57fdb610485aead71cfba8385a3da520/vendor/github.com/scylladb/scylla-manager/backupspec/manifest.go#L190)) while right now we are reading tables one by one ([ForEachIndexIterWithError](https://github.com/scylladb/scylla-manager/blob/455acc4d57fdb610485aead71cfba8385a3da520/vendor/github.com/scylladb/scylla-manager/backupspec/manifest.go#L234)) from the index file. 

So I propose we leave this idea for later stage optimizations if we start observing that refrerhNodeWorker is spending too much time idle

pkg/service/one2onerestore/worker_tables.go
VAveryanov8: yes, right

pkg/service/one2onerestore/worker_tables.go
Michal-Leszczynski: Issue is [here](https://github.com/scylladb/scylla-manager/issues/4402).

Michal-Leszczynski (approved):

PR #4399, opened 2025-05-29
refactor(managerclient): use util/sizesuffix package

VAveryanov8:
This replaces sizesuffix.go in a managerclient with a package from util module

---

Please make sure that:


v3/pkg/managerclient/go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/lnquy/cron v1.1.1
 	github.com/pkg/errors v0.9.1
 	github.com/scylladb/go-set v1.0.2
-	github.com/scylladb/scylla-manager/v3/pkg/util v0.0.0-20250310104639-18257b71bb48
+	github.com/scylladb/scylla-manager/v3/pkg/util v0.0.0-20250529094758-b6ddc3fd1fee
```
Michal-Leszczynski, line 13: Could we update all the modules to the same commit so that we avoid unexpected issues?
Or do you think that's not a problem?

Michal-Leszczynski (approved):

v3/pkg/managerclient/go.mod
VAveryanov8: I think it's not problem - the main module (scylla-manager) is already updated in feature-branch, so updating it here can cause some conflicts

PR #4403, opened 2025-06-02
restapi: don't expose deleted tasks

### LINKED ISSUE #4400: managerclient's GetTask should return an error for deleted tasks
When trying to get a deleted task, managerclient's `GetTask` method behaves differently than `ListTasks`. See the example below:

```
tasks, err := managerClient.ListTasks(ctx, managerClusterID, managerclient.RepairTask, false, "", managerTaskID.String())
if err != nil {
  panic(err)
}
if len(tasks) > 0 {
  panic("nonempty tasks returned")
}

// GetTask should return an error.
_, err = managerClient.GetTask(ctx, managerClusterID, managerclient.RepairTask, managerTaskID)
if err == nil {
  panic("nil error returned")
}
```

I'd expect GetTask to return an error if ListTasks does not return the task. The existing behaviour is inconsistent.

Michal-Leszczynski:
Fixes #4400


VAveryanov8 (approved):
LGTM 

JFYI This is a breaking change, so in theory we can break someones integration 

(To avoid breaking change, we can introduce some query param, like `?show_deleted=false`)

Michal-Leszczynski:
> JFYI This is a breaking change, so in theory we can break someones integration
(To avoid breaking change, we can introduce some query param, like ?show_deleted=false)

I wouldn't call it a breaking change, because we never documented this behavior (that you can query deleted task over API). It also does not make much sense on its own. In the part of code responsible for listing tasks we never allow for listing deleted tasks (not even with a flag):
```go
	// Never expose deleted tasks over API
	filter.Deleted = false
```
There is also no way for un-deleting a task over API.

I'm not sure why the previous maintainers decided to add `deleted` column to task definition instead of just deleting the task from DB, but maybe it was so that there is some manual recovery procedure (cqlsh into SM DB and change things directly there).

VAveryanov8:
> @VAveryanov8 @karol-kokoszka does this https://github.com/scylladb/scylla-manager/pull/4403#issuecomment-2933932991 convince you or do you think that we should keep the current behavior?

As I already approved this pr, I'm not against this breaking change. We can at least mentioned that the behavior has changed in the release notes

Issue #4410, opened 2025-06-03
Use move_files option for native backup


Michal-Leszczynski:
SM should order Scylla to remove already uploaded files from the snapshot directory, because:
1. Rclone does it and the pause/resume error/retry mechanisms in SM rely on it
2. It decreases disk space amplification caused by keeping the already compacted sstables on the disk

The definition of this API parameter can be found [here](https://github.com/scylladb/scylladb/blob/7fef2c4f619022cd6a08bea6fe79699efdcfbd62/api/api-doc/storage_service.json#L817-L824).

cc: @karol-kokoszka @bhalevy @regevran

bhalevy:
> The option **not** to remove files was never supposed to be in scylla. Somehow it slipped in.

Is there a design doc stating that?
I wonder what went wrong in the process...

regevran:
> I wonder what went wrong in the process...

[Requested](https://github.com/scylladb/scylladb/pull/22285#discussion_r1921515637), [Re-requested](https://github.com/scylladb/scylladb/pull/22285#discussion_r1922360273), [Re-re-requested](https://github.com/scylladb/scylladb/pull/22285#pullrequestreview-2574275355)  and eventually ignored.
🤷‍♂

PR #4411, opened 2025-06-03
swagger/scylla: add move_files param to /storage_service/backup

Michal-Leszczynski:
Needed for #4410.
Swagger schema taken from [here](https://github.com/scylladb/scylladb/blob/7fef2c4f619022cd6a08bea6fe79699efdcfbd62/api/api-doc/storage_service.json#L817-L824).


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 could you take a look? It's just a simple extension of swagger schema.

VAveryanov8 (approved):

PR #4412, opened 2025-06-03
backup: move files with native backup

### LINKED ISSUE #4410: Use move_files option for native backup

Michal-Leszczynski:
Fixes #4410


VAveryanov8 (approved):
:+1:

karol-kokoszka (approved):

Issue #4404, opened 2025-06-02
Do the repair of the base table before proceeding with the index rebuild.


karol-kokoszka:
As long, as there is no possibility of creating the view with pre-selected tombstone-gc-mode, the base table must be repaired before -re-creating the view -> https://github.com/scylladb/scylladb/issues/16454

During 1-1 restore, whenever there is an index available in the schema, the base table must be repaired before rebuilding it.

-----

Add the "repairing base table" to the progress output when the base table is being repaired.

-----

PR #4417, opened 2025-06-05
feat(1-1-restore): repair base table before recreating Index 

VAveryanov8:
This adds repair service to 1-1-restore. Repair will be run only if backup
contains Indexes to ensure data consistency.

This adds --disable-repair feature flag

**Note**: Progress is not adjusted to show repair. This could be done in a separate PR, but on the other hand, if we think that a repair operation is just a temporary fix, then maybe it's not worth the effort of adjusting progress at all.

Refs: https://github.com/scylladb/scylla-manager/issues/4404

---

Please make sure that:

### DIFF
--- a/docs/source/sctool/partials/sctool_restore_1-1-restore.yaml
+++ b/docs/source/sctool/partials/sctool_restore_1-1-restore.yaml
@@ -12,10 +12,6 @@ options:
       usage: |
         Task schedule as a cron `expression`.
         It supports the extended syntax including @monthly, @weekly, @daily, @midnight, @hourly, @every X[h|m|s].
-    - name: disable-repair
-      default_value: "false"
-      usage: |
-        By default, repair is run on tables that have Indexes to ensure data consistency. By disabling repair, you can speed up the restore process, but it may lead to data inconsistency in case of Indexes, so use it with caution.
     - name: dry-run
       default_value: "false"
       usage: |
--- a/docs/source/sctool/partials/sctool_restore_1-1-restore_update.yaml
+++ b/docs/source/sctool/partials/sctool_restore_1-1-restore_update.yaml
@@ -13,10 +13,6 @@ options:
       usage: |
         Task schedule as a cron `expression`.
         It supports the extended syntax including @monthly, @weekly, @daily, @midnight, @hourly, @every X[h|m|s].
-    - name: disable-repair
-      default_value: "false"
-      usage: |
-        By default, repair is run on tables that have Indexes to ensure data consistency. By disabling repair, you can speed up the restore process, but it may lead to data inconsistency in case of Indexes, so use it with caution.
     - name: dry-run
       default_value: "false"
       usage: |


pkg/command/one2onerestore/cmd.go
```diff
@@ -81,6 +82,8 @@ func (cmd *command) init() {
 	w.Unwrap().Var(&cmd.sourceCluster, "source-cluster-id", "")
 	w.Unwrap().StringVarP(&cmd.snapshotTag, "snapshot-tag", "T", "", "")
 	w.Unwrap().Var(&cmd.nodesMapping, "nodes-mapping", "")
+	// Feature flag to disable repair, should be removed once issue https://github.com/scylladb/scylladb/issues/16454 is resolved
+	w.Unwrap().BoolVar(&cmd.disableRepair, "disable-repair", false, "")
```
Michal-Leszczynski, line 86: I still don't get why we have such a flag:
- users should never use it
- repair of index base table is an implementation detail that user should not rely on
- we won't be needing to run repair in a short future, but we can't just remove released flag. Instead, even with `--disable-repair=false` SM would not run the repair? It's confusing.

Do we want to use it for testing whether skipping repair on index base table can cause data resurrection? Why would we want to test that? In terms of general performance testing, the time needed for running the repair should just be subtracted from the total restore time, which should be simple to do.

pkg/service/one2onerestore/service_integration_test.go
```diff
@@ -57,67 +57,137 @@ func TestOne2OneRestoreServiceIntegration(t *testing.T) {
...
+				isView:       true,
+				expectedMode: modeRepair,
+			},
+			// Until the https://github.com/scylladb/scylladb/issues/16454 is fixed,
+			// it is expected that the Secondary Index tombstone_gc mode will be 'timeout'.
+			{
+				ks:           ksName,
+				name:         siTableName,
+				isView:       true,
+				expectedMode: modeTimeout,
```
Michal-Leszczynski, line 110: nit: why is it expected? The current implementation treats indexes in the same way as materialized views (change tgc mode to repair, save schema, drop views, restore base tables, recreate views - with altered tgc mode). This means that this test will start failing for newer Scylla which fixes https://github.com/scylladb/scylladb/issues/16454. IMHO we shouldn't test that index tgc mode is set to timeout exactly, because we don't even care about it with the current workaround. I would not validate index tgc mode at all for now, and validate that it is set to repair starting from some Scylla release.

Michal-Leszczynski (commented):
> Note: Progress is not adjusted to show repair. This could be done in a separate PR, but on the other hand, if we think that a repair operation is just a temporary fix, then maybe it's not worth the effort of adjusting progress at all.

That's right, but I would say that the same logic applies to the `--disable-repair` flag.

pkg/command/one2onerestore/cmd.go
karol-kokoszka: Let's get rid of this flag.

Michal-Leszczynski (approved):

PR #4423, opened 2025-06-10
feat(1-1-restore): add one2onerestore_progress metric

VAveryanov8:
This adds one2onerestore_progress metric that will show percent of downloaded bytes from 0 to 100%.
Considering that loadSSTables calls should be quick and the most amount of time 1-1-restore will spend downloading files from s3 it's fair to say that this metric indicates global 1-1-restore process progress. 

---

Please make sure that:

### DIFF
--- a/pkg/metrics/one2onerestore.go
+++ b/pkg/metrics/one2onerestore.go
@@ -25,7 +25,7 @@ func NewOne2OneRestoreMetrics() One2OneRestoreMetrics {
 			"cluster", "location", "snapshot_tag", "host", "worker"),
 		viewBuildStatus: g("Defines build status of recreated view.", "view_build_status",
 			"cluster", "keyspace", "view"),
-		progress: g("Defines current progress of the 1-1-restore process.", "progress",
+		progress: g("Defines current progress of the 1-1-restore process in percents from 0 to 100%", "progress",
 			"cluster", "snapshot_tag"),
 	}
 }
--- a/pkg/metrics/one2onerestore.go
+++ b/pkg/metrics/one2onerestore.go
@@ -12,6 +12,7 @@ type One2OneRestoreMetrics struct {
 	remainingBytes  *prometheus.GaugeVec
 	state           *prometheus.GaugeVec
 	viewBuildStatus *prometheus.GaugeVec
+	progress        *prometheus.GaugeVec
 }
 
 func NewOne2OneRestoreMetrics() One2OneRestoreMetrics {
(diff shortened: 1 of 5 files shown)


pkg/metrics/one2onerestore.go
```diff
@@ -24,6 +25,8 @@ func NewOne2OneRestoreMetrics() One2OneRestoreMetrics {
 			"cluster", "location", "snapshot_tag", "host", "worker"),
 		viewBuildStatus: g("Defines build status of recreated view.", "view_build_status",
 			"cluster", "keyspace", "view"),
+		progress: g("Defines current progress of the 1-1-restore process.", "progress",
+			"cluster", "snapshot_tag"),
```
Michal-Leszczynski, line 29: Let's mention that it's represented in 0-100%.

pkg/service/one2onerestore/worker_tables.go
```diff
@@ -186,6 +186,7 @@ func (w *worker) waitJob(ctx context.Context, jobID int64, m *backupspec.Manifes
 			)
 			stats.incrementDownloadStats(job.Uploaded, took.Milliseconds())
 			w.metrics.SetOne2OneRestoreState(w.runInfo.ClusterID, m.Location, m.SnapshotTag, h.Addr, downloadWorkerName, metrics.One2OneRestoreStateIdle)
+			w.metrics.SetProgress(w.runInfo.ClusterID, m.SnapshotTag, stats.progress())
 			return nil
 		case scyllaclient.JobRunning:
 			continue
```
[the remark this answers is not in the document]
VAveryanov8: yes, because we printing it only once, when all tables are downloaded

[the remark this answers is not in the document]
Michal-Leszczynski: But stats are used for setting progress, so I guess we also need to update them on running job status if we want to update progress on running job status.

pkg/service/one2onerestore/worker_tables.go
```diff
@@ -189,6 +189,9 @@ func (w *worker) waitJob(ctx context.Context, jobID int64, m *backupspec.Manifes
 			w.metrics.SetProgress(w.runInfo.ClusterID, m.SnapshotTag, stats.progress())
 			return nil
 		case scyllaclient.JobRunning:
+			took := timeutc.Since(time.Time(job.StartedAt))
+			stats.incrementDownloadStats(job.Uploaded, took.Milliseconds())
+			w.metrics.SetProgress(w.runInfo.ClusterID, m.SnapshotTag, stats.progress())
```
Michal-Leszczynski, line 194: It looks like this is adding all uploaded bytes by the job every time we get running job status.
It should be handled in the same way as `SetDownloadRemainingBytes` (incrementing just the delta).
That's for both stats and progress metric.

VAveryanov8: yep, my bad - fixed!

Issue #4367, opened 2025-04-29
Make 1-1-restore task exclusive to all other ones.


karol-kokoszka:
We are expecting the cluster managed by Scylla Manager to be suspended during the 1-1 restore.
The 1-1 restore task goes through the scheduler, what means that the cluster cannot be just suspended for 1-1 restore.

Expectation:
- Scylla Manager suspends all the running tasks 
- Scylla Manager does the 1-1 restore
- Scylla Manager resumes other tasks

Or, 

- Scheduler is updated the way that it allows to perform any restore operation when the cluster is suspended (including creation of 1-1 task).
- 1-1 restore validates if the tasks are suspended
- if yes, moves forward
- if nor, instructs that cluster must be suspended

Michal-Leszczynski:
I'm in favor of using the same mechanism as for the regular restore: https://github.com/scylladb/scylla-manager/pull/4122.

VAveryanov8:
> I'm in favor of using the same mechanism as for the regular restore: https://github.com/scylladb/scylla-manager/pull/4122.

1-1-restore task is already using that mechanism  - https://github.com/scylladb/scylla-manager/blob/d664493ead4ab8986b132f8de724b80c2c82bf80/pkg/cmd/scylla-manager/server.go#L163

The behavior of `exclusive` task is following: 
Exclusive task is executed only if there are no other tasks running in the cluster. Conversely, other tasks can run only if the exclusive task is not running.

So it won't pause any running tasks, but instead will fail to start.

karol-kokoszka:
> So it won't pause any running tasks, but instead will fail to start.

But it is backed off when siren tries to schedule the task and the cluster is suspended. It cannot schedule it then.
So, it should schedule it still, when the cluster is suspended.

Michal-Leszczynski:
> But it is backed off when siren tries to schedule the task and the cluster is suspended. It cannot schedule it then.
So, it should schedule it still, when the cluster is suspended.

So the suggestion is to change the suspended cluster behavior so that it allows for scheduling new tasks, but it does not allow for running them, right? So it's not something connected directly to the 1-1 restore.

PR #4431, opened 2025-06-16
feat(swagger): add `allow_task_type` parameter to `suspended` endpoint

VAveryanov8:
This adds `allow_task_type` query parameter to
`/cluster/{cluster_id}/suspended` endpoint. If this parameter is provided that tasks of this type will be allowed to run in the suspended cluster.

---

Please make sure that:


Michal-Leszczynski:
So what is the design of the functionality solving #4367?
As we discussed, the `GET /cluster/{cluster_id}/suspended` API cannot be extended with the `allow_task_type` because of its returned type. This PR extends the `POST` API so that it allows for suspending all tasks except provided task type, but how will it be possible to check that the cluster is suspended except for provided task type? Or are we just saying that we won't provide such functionality?

Michal-Leszczynski (approved):

PR #4438, opened 2025-06-23
[1-1-restore] Changes in managerclient and swagger modules

VAveryanov8:
This is a first batch of commits for the 1-1-restore feature:
There is no way how we can merge 1-1-restore-feature-branch into master without breaking go module references to `swagger` and `managerclient` modules, so I decided to do it in a batches, firstly we should merge changes inside modules and once they are in the master, I can safely refer to them in next PRs.
Doing it in batches should also simplify the review.
Also I think we should merge 1-1-restore related commits to master without squashing.

---

Please make sure that:


Michal-Leszczynski:
>  Also I think we should merge 1-1-restore related commits to master without squashing.

Agree!

Michal-Leszczynski (approved):
nit: `Relates` -> `Ref` in the first commit message. I don't know if there is any automation relying on it, but we should stick to a single description for consistency. 

nit: `feat(1-1-restore): add swagger schema for 1-1-restore progress` -> `feat(swagger): scylla-manager, add schema for 1-1-restore progress` in the second commit title. The `feat(X)` should describe the scope, which should be swagger in this case. 

nitnit: ` feat(swagger): support for skip_reshape on SSTables load` -> `feat(swagger): scylla_v1, support for skip_reshape on SSTables load`

VAveryanov8:
@Michal-Leszczynski JFYI I've included one more commit into this batch

PR #4443, opened 2025-06-25
feat(sctool): add suspend details table to status command 

VAveryanov8:
This extends `sctool status` output with cluster suspend details table.

Refs: #4367

Example of the new output:
```text
❯ ./sctool.dev status -c my-cluster
╭────────────┬───────────┬──────────────╮
│ Cluster ID │ Suspended │ Allowed Task │
├────────────┼───────────┼──────────────┤
│ my-cluster │ true      │ 1_1_restore  │
╰────────────┴───────────┴──────────────╯
...


```

---

Please make sure that:


pkg/command/status/cmd.go
```diff
@@ -55,6 +55,13 @@ func (cmd *command) run() error {
...
+		suspendDetails, err := cmd.client.SuspendDetails(cmd.Context(), clusterID)
+		if err != nil {
+			return err
+		}
+		if err := suspendDetails.Render(w); err != nil {
+			return err
+		}
 		status, err := cmd.client.ClusterStatus(cmd.Context(), clusterID)
 		if err != nil {
 			return err
```
Michal-Leszczynski, line 67: This seems a little bit clunky for multiple multi-dc clusters.
To be honest, I though that `sctool status` displays status of only a single cluster, but it's actually printing a table per cluster per dc, so the clunkiness comes mostly from this part.

Could we display suspend details only when the cluster is suspended, so that we try not to increase the clunkiness level?

VAveryanov8: agree, I'll render this table only if cluster is suspended

Michal-Leszczynski (approved):

PR #4447, opened 2025-06-27
[1-1-restore] Next batch of commits to master

VAveryanov8:
This is the third batch of commits from 1-1-restore-feature-branch to master that includes following changes: 

* progress rendering in sctool 
* fail validation if tablet-based keyspaces are present
* disable auto-compaction during download&refresh
* 1-1-restore metrics
* download speed optimizations (pin&unpin agent cpu, 2*shard_count transfers)
* alter tombstone_gc mode for views
* queue for refreshNode calls
* swagger: change type of skip_cleanup, skip_reshape from string to boolean to align with the rest of parameters 

---

Please make sure that:


karol-kokoszka (approved):

v3/swagger/scylla_v1.json
```diff
@@ -11887,7 +11887,7 @@
             "name": "skip_cleanup",
             "in": "query",
             "required": false,
-            "type": "string",
+            "type": "boolean",
```
Michal-Leszczynski, line 11890: In general it makes sense, but in scylla swagger definition we have strings: https://github.com/scylladb/scylladb/blob/c6307aafd57745b2f9c27172fb029841918e0cba/api/api-doc/storage_service.json#L2148-L2163.

Michal-Leszczynski (commented):
Is this the final batch in terms of 1-1-restore service implementation?

v3/swagger/scylla_v1.json
VAveryanov8: yes, but also scylla swagger describe `load_and_stream` and `primary_replica_only` params as strings [link](https://github.com/scylladb/scylladb/blob/c6307aafd57745b2f9c27172fb029841918e0cba/api/api-doc/storage_service.json#L2133-L2146)
but in the manager they are described as a booleans [link](https://github.com/scylladb/scylla-manager/blob/64b5a3b94aafabf4b135959c715fe9c07a197e89/v3/swagger/scylla_v1.json#L11872-L11885)

VAveryanov8:
> Is this the final batch in terms of 1-1-restore service implementation?

There will be at least one more with the usage of `skip_reshape` param and also I think it worth to include rendering of recently introduces functionality - suspend with allowed task

Michal-Leszczynski (approved):
Approved as I'm waiting with the review for the final version.

Issue #4459, opened 2025-07-02
1-1-restore nits/leftovers


Michal-Leszczynski:
Here is a list of minor things I spotted when re-reviewing 1-1-restore code merged to master:
- no mentions of the 1-1-restore in the docs apart from the sctool docs. Shouldn't we at least mention it on the restore page?
- if we want 1-1-restore to work with older Scylla versions, then we should add version checks before using newer features (e.g. `skip_cleanup` when loading sstables)
- 1-1-restore requires backup made with proper SM version. I don't think that we explicitly validate that provided backup does not contain empty cpu/shards/rack descriptions
- when re-creating views we just exec the saved CQL stmt, but it might be the case that this CQL stmt is actually a bundle of CQL stmts separated by `;`, so we need to exec them separately, because exec of bundled CQL stmts fails with gocql ExecStmt func
- **no mentions (even in the sctool docs) of the fact that 1-1-restore works with vnodes only**
- **the regular and single host CQL sessions aren't closed**
- **no mentions (even in the sctool docs) of the fact that 1-1-restore changes tgc mode to repair and that if user wants to reset it, they should run repair first**

@VAveryanov8 please verify if those observations are correct.
IMO we should at least address the last 3 points (the ones in **bold**) before releasing SM 3.6. Since they are really small items, it shouldn't delay the release.

@karol-kokoszka please take a look as well.

karol-kokoszka:
@Michal-Leszczynski Thanks for the review !

> no mentions of the 1-1-restore in the docs apart from the sctool docs. Shouldn't we at least mention it on the restore page?

We should, but I don't see it as the release blocker.

> if we want 1-1-restore to work with older Scylla versions, then we should add version checks before using newer features (e.g. skip_cleanup when loading sstables)

If the API call fails due to unknown query parameters, we must fix it, but I'm not sure of that. @VAveryanov8 could you please check ?

> 1-1-restore requires backup made with proper SM version. I don't think that we explicitly validate that provided backup does not contain empty cpu/shards/rack descriptions

1-1-restore requires the infrastructure provided by other orchestration tool as well.
I understand the validation in SM will fail if the backup doesn't contain needed data with some (possibly random) error. @VAveryanov8 ? 

> when re-creating views we just exec the saved CQL stmt, but it might be the case that this CQL stmt is actually a bundle of CQL stmts separated by ;, so we need to exec them separately, because exec of bundled CQL stmts fails with gocql ExecStmt func

[...]

PR #4460, opened 2025-07-02
fix(1-1-restore): fix a few leftovers

VAveryanov8:
This addresses following leftovers described in #4459

- [x] no mentions of the 1-1-restore in the docs apart from the sctool docs. Shouldn't we at least mention it on the restore page?
- [x] when re-creating views we just exec the saved CQL stmt, but it might be the case that this CQL stmt is actually a bundle of CQL stmts separated by ;, so we need to exec them separately, because exec of bundled CQL stmts fails with gocql ExecStmt func
- [x] no mentions (even in the sctool docs) of the fact that 1-1-restore works with vnodes only
- [x] the regular and single host CQL sessions aren't closed
- [x] no mentions (even in the sctool docs) of the fact that 1-1-restore changes tgc mode to repair and that if user wants to reset it, they should run repair first

I suggest to merge this PR without squashing as it contains changes in various parts, e.g. docs and implementation.

Fixes: #4459

---

Please make sure that:


VAveryanov8:
Tests are failing due to same network issues: 

`215.9 E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/n/net-tools/net-tools_1.60%2bgit20181103.0eebece-1ubuntu5.4_amd64.deb  Cannot initiate the connection to security.ubuntu.com:80 (2620:2d:4000:1::102). - connect (101: Network is unreachable) Cannot initiate the connection to security.ubuntu.com:80 (2620:2d:4000:1::101). - connect (101: Network is unreachable) Cannot initiate the connection to security.ubuntu.com:80 (2620:2d:4002:1::103). - connect (101: Network is unreachable) Cannot initiate the connection to security.ubuntu.com:80 (2620:2d:4000:1::103). - connect (101: Network is unreachable) Cannot initiate the connection to security.ubuntu.com:80 (2620:2d:4002:1::102). - connect (101: Network is unreachable) Cannot initiate the connection to security.ubuntu.com:80 (2620:2d:4002:1::101). - connect (101: Network is unreachable) [IP: 91.189.91.82 80]`

Michal-Leszczynski:
@VAveryanov8 IMO we can merge without the checks.

PR #4463, opened 2025-07-03
Merge native restore support

### LINKED ISSUE #4137: Handle SM restore task pause/resume with Scylla API
We should support the same SM task pause/resume functionality with native Scylla restore task.

### LINKED ISSUE #4142: Monitor Scylla restore API progress
When using Rclone API, restoring a batch of SSTables consists of 2 separate actions: download and load&stream.
Each of those actions has different start/end time and progress granularity displayed in `sctool progress`.
They also have separate metrics like `downloaded_bytes` and `streamed_bytes`.

On the other hand, when using Scylla API, restoring a batch of SSTables consists of just a single action: call to the Scylla restore API.

We should decide on how to expose restore progress when the Scylla API was used.
There are two solutions, either populate the old `sctool progress` and metrics so that `restored=downloaded=streamed`, or create new metrics and display for `sctool progress`.

Note that even when using Scylla restore API, the backup could still contain some integer based SSTable IDs, which can't be restored with Scylla API, so they would need to fall back to the Rclone API. The progress representation should also take it into consideration.

Issue about querying progress of restore API call: https://github.com/scylladb/scylladb/issues/21099
Issue about incorrect value returned when querying progress of restore API call (not a blocker for development): https://github.com/scylladb/scylladb/issues/21816

### LINKED ISSUE #4144: Replace batch download + L&S with the call to POST /storage_service/restore
https://docs.google.com/document/d/1d-nJnn5n2gqzRXvP5xpOis0fWxIzYkf-u9X1xrIAG_c/edit?tab=t.0#bookmark=id.ktxjj4naswzn

The restore performed by the Scylla server is expected to work directly on the S3 data, there is no need of copying the SSTables by SM anymore. + There is no need for calling to L&S.

The part above is expected to be replaced with a simple call to POST /storage_service/restore where one of the input parameters is the list of SSTables that are going to be handled by the current node.

### LINKED ISSUE #4158: Set Scylla task TTL

### LINKED ISSUE #4194: Restore integer based SSTables when using Scylla restore API
Taken from https://github.com/scylladb/scylladb/issues/22022#issuecomment-2567605485.

Scylla restore API does not support restoration of SSTables with integer based IDs, so we should keep on using Rclone API in their case.
The fallback to Rclone API should happen only for the integer based SSTables, so all UUID based SSTables should still be restored with Scylla API. This requires to create per SSTable API compatibility check and group SSTables restored with Scylla API and Rclone API into separate batches.

### LINKED ISSUE #4385: Set Scylla task TTL for native restore

Michal-Leszczynski:
This PR if a followup to #4381 which contains remaining changes from `ml/scylla-api-2`.
It introduces a replacement for rclone copypaths API in a form of native restore Scylla API. 

Fixes #4144
Fixes #4137
Fixes #4142
Fixes #4158
Fixes #4385
Fixes #4194

### DIFF
--- a/pkg/metrics/restore.go
+++ b/pkg/metrics/restore.go
@@ -15,6 +15,8 @@ type RestoreMetrics struct {
 	batchSize        *prometheus.GaugeVec
 	remainingBytes   *prometheus.GaugeVec
 	state            *prometheus.GaugeVec
+	restoredBytes    *prometheus.GaugeVec
+	restoreDuration  *prometheus.GaugeVec
 	downloadedBytes  *prometheus.GaugeVec
 	downloadDuration *prometheus.GaugeVec
 	streamedBytes    *prometheus.GaugeVec
@@ -31,6 +33,8 @@ func NewRestoreMetrics() RestoreMetrics {
 		remainingBytes: g("Remaining bytes of backup to be restored yet.", "remaining_bytes",
 			"cluster", "snapshot_tag", "location", "dc", "node", "keyspace", "table"),
 		state:            g("Defines current state of the restore process (idle/download/load/error).", "state", "cluster", "location", "snapshot_tag", "host"),
+		restoredBytes:    g("Restored bytes", "restored_bytes", "cluster", "host"),
+		restoreDuration:  g("Restore duration in ms", "restore_duration", "cluster", "host"),
 		downloadedBytes:  g("Downloaded bytes", "downloaded_bytes", "cluster", "location", "host"),
 		downloadDuration: g("Download duration in ms", "download_duration", "cluster", "location", "host"),
 		streamedBytes:    g("Load&Streamed bytes", "streamed_bytes", "cluster", "host"),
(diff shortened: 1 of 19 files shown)


karol-kokoszka:
I'm fine with merging it.

VAveryanov8 (approved):
:+1:

PR #4464, opened 2025-07-03
fix VERSION_NAME value in makefile

Annamikhlin:
follow the commit: https://github.com/scylladb/scylla-manager/commit/e1b19dacaeb282956826bf92776bf16522183a8b manager-build stopped work due to missing tag version

adding back the VERSION_NAME

Fixes: https://github.com/scylladb/scylla-pkg/issues/5242

Verification passed: https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/manager-build/1040/

PR #4465, opened 2025-07-07
Fix jenkins build for branch-3.6

Michal-Leszczynski:
Backport of https://github.com/scylladb/scylla-manager/pull/4464 into `branch-3.6`.

### DIFF
--- a/dist/Makefile
+++ b/dist/Makefile
@@ -1,16 +1,16 @@
 RELEASE      := 0.$(shell date +%Y%m%d).$(shell git describe --always)
-VERSION=3.6.0
+VERSION      ?=3.6.0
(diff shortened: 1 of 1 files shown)


dist/Makefile
```diff
@@ -1,16 +1,16 @@
 RELEASE      := 0.$(shell date +%Y%m%d).$(shell git describe --always)
-VERSION=3.6.0
+VERSION      ?=3.6.0-dev
```
Annamikhlin, line 2: cause it is a release branch please set it as `VERSION ?=3.6.0` without `-dev`

Annamikhlin: yes, approved :+1:

PR #4466, opened 2025-07-07
Control healthcheck ping cron schedule

### LINKED ISSUE #4445: cqlping health check is cron is hard coded to 15 seconds
https://github.com/scylladb/scylla-manager/blob/5452b87d6aab5c6c488d76c47d6f3350e80894f4/pkg/cmd/scylla-manager/tasks.go#L39

I think it's excessive, especially on larger clusters.

Michal-Leszczynski:
This PR makes it possible to control healthcheck ping cron schedule with `scylla-manager.yaml` config.
Changes to the config are applied to both new and already existing clusters.
This PR also changes default CQL and Alternator ping interval to 1m from 15s.

Fixes #4445


pkg/config/server/testdata/scylla-manager.yaml
```diff
@@ -35,6 +35,9 @@ healthcheck:
   max_timeout: 1s
   probes: 0
   node_info_ttl: 1s
+  cql_ping_cron: '* 5,15 * * * *'
+  rest_ping_cron: '* 5,15 * * * *'
+  alternator_ping_cron: '* 5,15 * * * *'
```
VAveryanov8, line 40: question: 
If I understand  `* 5,15 * * * *` correctly, then it means to run ping, every second of 5 minute and every second of 15 minute. Is this somehow used in the tests? Or maybe I misinterpret it?

pkg/service/healthcheck/config.go
```diff
@@ -20,7 +28,10 @@ type Config struct {
 
 func DefaultConfig() Config {
 	return Config{
-		MaxTimeout:  1 * time.Second,
-		NodeInfoTTL: 5 * time.Minute,
+		MaxTimeout:         1 * time.Second,
+		NodeInfoTTL:        5 * time.Minute,
+		CQLPingCron:        schedules.MustCron("*/15 * * * * *", time.Time{}),
```
VAveryanov8: nit:
"@every 15s"

feel free to just ignore those nits by just clicking "resolve" :)

pkg/config/server/testdata/scylla-manager.yaml
Michal-Leszczynski: This is just a golden file which is used in `TestConfigModification` to validate that changes to `scylla-manager.yaml` are reflected in the config. Those values are not used for anything else, so I just put some random non-default values here (and as you observed, they don't make much sense).

VAveryanov8 (approved):

PR #4475, opened 2025-07-09
Update Scylla versions

Michal-Leszczynski:
This PR updates Scylla versions in:
- github actions test matrix
- SM docs compatibility matrix
- SM docs versions

### DIFF
--- a/.github/cfg/integration-test-cfg.yaml
+++ b/.github/cfg/integration-test-cfg.yaml
@@ -1,38 +1,40 @@
-- scylla-version: scylla-enterprise:2024.1.15
+- scylla-version: scylla-enterprise:2024.1.18
   ip-family: IPV4
   raft-schema: disabled
   tablets: none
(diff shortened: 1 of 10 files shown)


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 I would say that we can merge this PR - the failures are expected and we should test 2025.2.0 in our CI. One other thing is that we could skip the 1-1-restore test if it runs agains 2025.2.0 if we want to have green CI, but since we expect this to be fixed shortly, I would be fine with a little red in our CI (but maybe that's a wrong approach).

karol-kokoszka (approved):
Restarted GH actions for integration-tests. 
Feel free to merge if they are green.

Michal-Leszczynski:
I will merge it when we release SM 3.6 (this PR updates stable docs to 3.6 and adds compatibility matrix for it).

Issue #4474, opened 2025-07-09
Handle describe ring API call with more care


Michal-Leszczynski:
As describe in this [comment](https://github.com/scylladb/scylla-enterprise/issues/5516#issuecomment-3025387703) (and the whole issue thread), it looks like the default timeout for describe ring API call is fragile for bigger clusters. We need to decide if we need to handle it with more care.

Pasted comment for readability:
This is the `/storage_service/describe_ring/{keyspace}` [scylla API](https://github.com/scylladb/scylladb/blob/fd0e04411835a822b91e1344c3a342ce40c0c8cc/api/api-doc/storage_service.json#L395-L429) which returns the token ring for given keyspace. It returns an array of [token ranges](https://github.com/scylladb/scylladb/blob/fd0e04411835a822b91e1344c3a342ce40c0c8cc/api/api-doc/storage_service.json#L3294-L3328) with [additional information](https://github.com/scylladb/scylladb/blob/fd0e04411835a822b91e1344c3a342ce40c0c8cc/api/api-doc/storage_service.json#L3276-L3293). This array should have `256 * 220 = 56320` items (`#vnodes_per_node * #node`).
 
This API is called as a preparation before running actual repairs. This API call is sent to only a single host in a cluster (ignore retries for now) and I'm not sure how is it executed on Scylla side (e.g. does it require communication with other nodes). We use the regular timeout for it (30s by default), but perhaps we need to either:
- just increase the timeout (e.g. `timeout = max(timeout, 5m)`) 
- have a dynamic timeout depending on the cluster size/topology (unclear on the function)
- introduce separate entry in `scylla-manager.yaml` for controlling this timeout value

I think that the first option is the simplest and should cover basically all scenarios (and if not, then it is still possible to change the default timeout in `scylla-manager.yaml`). Second option might be tricky and is just a guess. Third option is less convenient for the user.

Michal-Leszczynski:
So if I understand it correctly, SM 3.6.0 should include two additional changes:
- don't use bugged nodetool refresh params introduced in 2025.2.0 in 1-1-restore (we will start using them in SM 3.6.1 when we test them against 2025.2.X which contains the fixes). The 1-1-restore will still work and it will be faster than the regular restore, but not as fast as it could be. It will also make our CI green.
[...]

PR #4479, opened 2025-07-15
Scyllaclient: set describe ring API call timeout to at least 5min

Michal-Leszczynski:
Refs #4474
Refs https://github.com/scylladb/scylla-enterprise/issues/5516


VAveryanov8 (approved):

VAveryanov8:
The only issue is that after this pr is merged we won't be able to set the describe ring api timeout less than 5 minutes if we want

PR #4480, opened 2025-07-15
fix(1-1-restore): set skip_{cleanup|reshape} only for scylla > 2025.2.0

VAveryanov8:
This sets skip_cleanup and skip_reshape parameters to /storage_service/sstables/{keyspace} call only if scylla version greater than 2025.2.0 as this version has a known bug and will return an error if these parameters are provided.

---

Please make sure that:

### DIFF
--- a/pkg/scyllaclient/client_agent.go
+++ b/pkg/scyllaclient/client_agent.go
@@ -309,6 +309,19 @@ func (ni *NodeInfo) ScyllaObjectStorageEndpoint(provider backupspec.Provider) (s
 		"and in `scylla.yaml` object_storage_endpoints config", provider)
 }
 
+// SupportsSkipCleanupAndSkipReshape returns whether Scylla supports skip_cleanup and skip_reshape parameters
+// in /storage_service/sstables/{keyspace} endpoint as 2025.2.0 has a bug if this parameters are set to true -
+// https://github.com/scylladb/scylladb/issues/24913.
+func (ni *NodeInfo) SupportsSkipCleanupAndSkipReshape() (bool, error) {
+	// Detect master builds
+	if scyllaversion.MasterVersion(ni.ScyllaVersion) {
+		return true, nil
+	}
+
+	// Check ENT
+	return scyllaversion.CheckConstraint(ni.ScyllaVersion, "> 2025.2.0")
+}
+
 // FreeOSMemory calls debug.FreeOSMemory on the agent to return memory to OS.
 func (c *Client) FreeOSMemory(ctx context.Context, host string) error {
 	p := operations.FreeOSMemoryParams{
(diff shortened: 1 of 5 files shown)


VAveryanov8:
@Michal-Leszczynski @karol-kokoszka what are we going to do with this PR? Close it?

pkg/service/one2onerestore/worker_tables.go
```diff
@@ -138,6 +138,15 @@ func (w *worker) refreshNode(ctx context.Context, table backupspec.FilesMeta, m
 	return err
 }
 
+func (w *worker) awaitLoadSSTables(ctx context.Context, host Host, keyspace, table string) error {
+	var skipCleanup, skipReshape bool
+	if host.SkipCleanupAndSkipReshape {
+		skipCleanup, skipReshape = true, true
+	}
+
+	return w.client.AwaitLoadSSTables(ctx, host.Addr, keyspace, table, false, false, skipCleanup, skipReshape)
```
Michal-Leszczynski, line 147: nit: we can just:
 ```go
w.client.AwaitLoadSSTables(ctx, host.Addr, keyspace, table, false, false, host.SkipCleanupAndSkipReshape, host.SkipCleanupAndSkipReshape)
```

File: pkg/service/one2onerestore/worker.go [hunk 2]
```diff
@@ -218,6 +218,11 @@ func (w *worker) prepareHostWorkload(ctx context.Context, manifests []*backupspe
 			return errors.Wrapf(err, "node %s safe describe method", h.Addr)
 		}
 		hw.host.SafeDescribeMethod = method
+		supports, err := nodeInfo.SupportsSkipCleanupAndSkipReshape()
+		if err != nil {
+			return errors.Wrapf(err, "node %s supports skip_cleanup and skip_reshape", h.Addr)
+		}
+		hw.host.SkipCleanupAndSkipReshape = supports
```
Michal-Leszczynski, line 225: nit: I would just store NodeInfo alongside host instead of populating host with simple method calls here. NodeInfo could also be taken from the node config service later on.
On the other hand, since we are already doing it for safe describe schema with internals, it's fine to go along with this pattern.

pkg/scyllaclient/client_agent.go
```diff
@@ -309,6 +309,19 @@ func (ni *NodeInfo) ScyllaObjectStorageEndpoint(provider backupspec.Provider) (s
...
+// https://github.com/scylladb/scylladb/issues/24913.
+func (ni *NodeInfo) SupportsSkipCleanupAndSkipReshape() (bool, error) {
+	// Detect master builds
+	if scyllaversion.MasterVersion(ni.ScyllaVersion) {
+		return true, nil
+	}
+
+	// Check ENT
+	return scyllaversion.CheckConstraint(ni.ScyllaVersion, "> 2025.2.0")
+}
```
Michal-Leszczynski, line 323: As mentioned [here](https://github.com/scylladb/scylla-manager/pull/4480#issuecomment-3072889171), we want to go into the direction of just using those params starting from Scylla version where they were released (2025.2.0). This is a Scylla side bug and we don't want/need to special case it here.

Michal-Leszczynski (approved):

PR #4483, opened 2025-07-15
Backport: fix(1-1-restore): set skip_{cleanup|reshape} only for scylla >= 2025.2.0

VAveryanov8:
This sets skip_cleanup and skip_reshape parameters to /storage_service/sstables/{keyspace} call only if scylla version is greater than or equal to 2025.2.0.

---

Please make sure that:

### DIFF
--- a/pkg/scyllaclient/client_agent.go
+++ b/pkg/scyllaclient/client_agent.go
@@ -299,6 +299,20 @@ func (ni *NodeInfo) ScyllaObjectStorageEndpoint(provider backupspec.Provider) (s
 		"and in `scylla.yaml` object_storage_endpoints config", provider)
 }
 
+// SupportsSkipCleanupAndSkipReshape returns whether Scylla supports skip_cleanup and skip_reshape parameters
+// in /storage_service/sstables/{keyspace} endpoint.
+// Note that scylla 2025.2.0 has a bug if this parameters are set to true -
+// https://github.com/scylladb/scylladb/issues/24913.
+func (ni *NodeInfo) SupportsSkipCleanupAndSkipReshape() (bool, error) {
+	// Detect master builds
+	if scyllaversion.MasterVersion(ni.ScyllaVersion) {
+		return true, nil
+	}
+
+	// Check ENT
+	return scyllaversion.CheckConstraint(ni.ScyllaVersion, ">= 2025.2.0")
+}
+
 // FreeOSMemory calls debug.FreeOSMemory on the agent to return memory to OS.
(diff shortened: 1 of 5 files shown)


Michal-Leszczynski:
I don't think that we want to backport it anymore.
It's not necessary, and we have already started running tests against `branch-3.6`.

karol-kokoszka:
@mikliapko did we test the Scylla Manager 1:1 against older versions of Scylla ? I mean the current build (or some recent older builds) against Scylla 2024.1 for example ? Or something older than 2025.2.0 ? If so, we can skip backporting this PR for sure.

karol-kokoszka:
@Michal-Leszczynski If you have so strong opinion against backporting this PR. And this is against the rules. I'm not pushing for doing it. The fact is that 1:1 restore is supported >= 2025.2. If someone triggers it outside of the cloud, he may end up with an error. That's fine.

mikliapko:
> @mikliapko did we test the Scylla Manager 1:1 against older versions of Scylla ? I mean the current build (or some recent older builds) against Scylla 2024.1 for example ? Or something older than 2025.2.0 ? If so, we can skip backporting this PR for sure.

I'll do the test now.

> Backporting it right now means that we change the tested SM build so we need to rerun the tests.

Totally agree. We will have to go through the whole release testing cycle again.

VAveryanov8:
@Michal-Leszczynski As far as I understand from our planning call, we want to make this check stricter, e.g use skip_cleanup&&skip_reshape only for scylla > 2025.2.0, right?

pkg/scyllaclient/client_agent.go
```diff
@@ -299,6 +299,20 @@ func (ni *NodeInfo) ScyllaObjectStorageEndpoint(provider backupspec.Provider) (s
...
 
+// SupportsSkipCleanupAndSkipReshape returns whether Scylla supports skip_cleanup and skip_reshape parameters
+// in /storage_service/sstables/{keyspace} endpoint.
+// Note that scylla 2025.2.0 has a bug if this parameters are set to true -
+// https://github.com/scylladb/scylladb/issues/24913.
+func (ni *NodeInfo) SupportsSkipCleanupAndSkipReshape() (bool, error) {
+	// Detect master builds
+	if scyllaversion.MasterVersion(ni.ScyllaVersion) {
+		return true, nil
+	}
```
Michal-Leszczynski, line 310: nit: this doesn't do anything (master check is bugged), but we can delete it later on.

Michal-Leszczynski (approved):

Issue #4482, opened 2025-07-15
Increase the timeout on describe ring API call


Michal-Leszczynski:
This is a first, temporary band-aid step towards #4474 and general fixing of backoff/retry mechanism in SM.
For now we will increase the timeout to `max(default_timeout, 5min)`, in order to alleviate https://github.com/scylladb/scylla-enterprise/issues/5516 before other fixes land. This change should be removed after the proper fixes have landed.

Issue #4469, opened 2025-07-08
Improve gocql port availability


Michal-Leszczynski:
Late issue for #4456.

Michal-Leszczynski:
Fixed in #4456.

Issue #4468, opened 2025-07-08
Don't mix --skip-schema and --purge-only flags


Michal-Leszczynski:
By mistake, `--skip-schema` flag is set to `--purge-only` flag value when using `sctool backup` command:
```go
	if cmd.Flag("skip-schema").Changed {
		props["skip_schema"] = cmd.purgeOnly
		ok = true
	}
```

Issue #4448, opened 2025-06-27
Inconsistent values shown by `sctool progress` for a repair job and `manager:repair_progress` and/or `scylla_manager_repair_progress` metrics values


vladzcloudius:
SM Version: 3.4.2-0.20250313.f47915cac

## Description
While the repair is running and `sctool` showing 10%:

```
$ sctool progress repair/all-weekly -c gumgum-tally-prod
Run:		946fa22c-5378-11f0-a5bc-02840ee7de1b
Status:		RUNNING
Start time:	27 Jun 25 17:03:00 UTC
Duration:	4h30m56s
Progress:	10%
Intensity:	2/5
Parallel:	1/1 (max)
Datacenters:	
  - AWS_AP_SOUTHEAST_1_4
  - AWS_EU_WEST_1_3
  - AWS_US_EAST_1_6
  - AWS_US_WEST_2_3

╭───────────────────────────────┬──────────────────────────────────┬──────────┬──────────╮
│ Keyspace                      │                            Table │ Progress │ Duration │
├───────────────────────────────┼──────────────────────────────────┼──────────┼──────────┤
│ ad_performance                │ daily_performance_by_tracking_id │ 100%     │ 2s       │
│ ad_performance                │                data_by_ad_buy_id │ 100%     │ 38s      │
│ ad_performance                │               data_by_package_id │ 0%       │ 1m18s    │
│ ad_performance                │                 data_by_phase_id │ 100%     │ 1h35m22s │
│ ad_performance                │              data_by_tracking_id │ 0%       │ 0s       │
│ ad_performance                │         package_performance_data │ 100%     │ 20s      │
│ ad_performance                │           phase_performance_data │ 100%     │ 25s      │
│ ad_performance                │                     test_counter │ 100%     │ 3s       │
├───────────────────────────────┼──────────────────────────────────┼──────────┼──────────┤
│ audit                         │                        audit_log │ 100%     │ 4s       │
├───────────────────────────────┼──────────────────────────────────┼──────────┼──────────┤
│ dev                           │                data_by_ad_buy_id │ 100%     │ 2s       │
│ dev                           │              data_by_tracking_id │ 100%     │ 2s       │
│ dev                           │         package_performance_data │ 100%     │ 2s       │
│ dev                           │           phase_performance_data │ 100%     │ 2s       │
├───────────────────────────────┼──────────────────────────────────┼──────────┼──────────┤
[...]

PR #4485, opened 2025-07-16
SM 3.5.1 patch release

Michal-Leszczynski:
This PR contains all patch fixes that were supposed to land in SM 3.6.0 release:
- Make local query instead of scan in healthcheck CQL ping ([#4329](https://github.com/scylladb/scylla-manager/issues/4329))
- Respect ‘backup --skip-schema’ flag ([#4468](https://github.com/scylladb/scylla-manager/issues/4468))
- Fix snaphost tag collision for a single ScyllaDB Manager instance ([#3873](https://github.com/scylladb/scylla-manager/issues/3873))
- Fix restore batch mechanism which caused partial restore on retry ([#4344](https://github.com/scylladb/scylla-manager/issues/4344))
- Increase the timeout on describe ring API call ([#4482](https://github.com/scylladb/scylla-manager/issues/4482))
- Don’t timeout on repair status and restore load&stream calls ([#4419](https://github.com/scylladb/scylla-manager/issues/4419))
- Improve gocql port availability which caused errors on creating CQL session ([#4469](https://github.com/scylladb/scylla-manager/issues/4469))
- Don’t long poll tablet repair status as it causes sporadic http error 400 ([#4434](https://github.com/scylladb/scylla-manager/issues/4434))
- Fix race condition when client call finished when context was cancelled ([#3848](https://github.com/scylladb/scylla-manager/issues/3848))
- Align restore bandwidth reported in ‘sctool progress’ and metrics ([#4358](https://github.com/scylladb/scylla-manager/issues/4358))
- Populate ‘scylla_manager_repair_progress’ metric ([#4448](https://github.com/scylladb/scylla-manager/issues/4448))
- Don’t return deleted task in manager client ([#4400](https://github.com/scylladb/scylla-manager/issues/4400))

It **does not** contain support for native backup and 1-1-restore.

I cherry-picked commits from the main module only - so no commits changing util/managerclient/swagger submodules. Those commits are not needed, but we actually still reference them via go.mod pointing to the master branch. So from the practical POV, those commits are here. As we are not properly releasing our submodules, I'm not sure about the right direction here. We could either reference those submodules from master (no real harm, just perhaps confusing), or we could drop our go.mod commits which target submodules updates and keep only the other ones (require simple conflict resolution).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/scylladb/go-reflectx v1.0.1
 	github.com/scylladb/go-set v1.0.2
 	github.com/scylladb/gocqlx/v2 v2.8.0
-	github.com/scylladb/scylla-manager/backupspec v1.0.1
+	github.com/scylladb/scylla-manager/backupspec v1.0.2
 	github.com/scylladb/scylla-manager/v3/pkg/managerclient v0.0.0-20250310104639-18257b71bb48
 	github.com/scylladb/scylla-manager/v3/pkg/util v0.0.0-20250310104639-18257b71bb48
 	github.com/scylladb/scylla-manager/v3/swagger v0.0.0-20250310104639-18257b71bb48
--- a/go.sum
+++ b/go.sum
@@ -1061,8 +1061,8 @@ github.com/scylladb/google-api-go-client v0.34.1-patched h1:DW+T0HA+74o6FDr3TFzV
 github.com/scylladb/google-api-go-client v0.34.1-patched/go.mod h1:RriRmS2wJXH+2yd9PRTEcR380U9AXmurWwznqVhzsSc=
 github.com/scylladb/rclone v1.54.1-0.20240312172628-afe1fd2aa65e h1:lJRphCtu+nKd+mfo8whOTeFkgjMWvk8iCSlqgibKSa8=
 github.com/scylladb/rclone v1.54.1-0.20240312172628-afe1fd2aa65e/go.mod h1:JGZp4EvCUK+6AM1Fe1dye5xvihTc/Bk0WnHHSCJOePM=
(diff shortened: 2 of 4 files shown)


Michal-Leszczynski:
> I cherry-picked commits from the main module only - so no commits changing util/managerclient/swagger submodules. Those commits are not needed, but we actually still reference them via go.mod pointing to the master branch. So from the practical POV, those commits are here. As we are not properly releasing our submodules, I'm not sure about the right direction here. We could either reference those submodules from master (no real harm, just perhaps confusing), or we could drop our go.mod commits which target submodules updates and keep only the other ones (require simple conflict resolution).

@karol-kokoszka @VAveryanov8 FYI: I'm trying to make go.mod from branch-3.5 point to branch-3.5 submodules for now

VAveryanov8 (approved):
I haven't reviewed the code, but according to the PR description and commit list everything looks good to me!

karol-kokoszka:
@Michal-Leszczynski please update backupspec to v1.0.2

Otherwise:
```
➜  scylla-manager git:(ml/sm-3.5.1-release) ✗ go mod vendor
go: downloading github.com/scylladb/scylla-manager/backupspec v1.0.1
verifying github.com/scylladb/scylla-manager/backupspec@v1.0.1: checksum mismatch
        downloaded: h1:PcAi9uGGoiG8cjHa4+oUXjps5Lw14PvcZptzoEfudSY=
        go.sum:     h1:WaFVHEWwVA3WUeCf2WakmhQ59o+wTYdzo4+qQf8maXk=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.

```

Other than that, +1

PR #4488, opened 2025-07-17
Revert "chore(go.mod): bump gocql to v1.14.5"

Michal-Leszczynski:
This reverts commit 8e14aa45758cd6409a4c8c9c6c87ccba6988788e.


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 after the revert, we haven't observed the panic in gocql ([argus](https://argus.scylladb.com/tests/generic/e8fc916d-68b5-44f2-bbb6-9a7d2c988bf6/dtest)).
This is just a single DTest suite run, but it gives some hope. If we merge anything to branch-3.5 we still will need to rerun our whole test suite before release, so I propose to merge this PR, rerun all tests and release if we don't observe the panic again.

In the meantime, I will also keep on triggering DTests to see if they observe the panic again, but I don't want to delay the the whole test suite.

Michal-Leszczynski:
cc: @mikliapko

PR #4489, opened 2025-07-17
Fix scylla master checks

### LINKED ISSUE #4484: scylla master version check does not work
Since given SM version supports multiple Scylla versions and because of https://github.com/scylladb/scylladb/issues/16424, when we want to use some new scylla API, we need to check if it is exposed/supported. The version checks for released scylla versions work fine, but it looks like `version.MasterVersion` does not work anymore.

It is implemented like this: https://github.com/scylladb/scylla-manager/blob/3bb55636b9994a3b2f12521e2d88ff6842f16219/v3/pkg/util/version/version.go#L31-L48

But the current Scylla version obtained from `scylladb/scylla-nightly:latest` looks like `2025.4.0~dev-0.20250714.145a38bc2e37`.

The check for native restore API support actually passes for `scylla-nightly:latest`, but that's not because of the master check, but because of the optimistic `2025.3` check: https://github.com/scylladb/scylla-manager/blob/3bb55636b9994a3b2f12521e2d88ff6842f16219/pkg/scyllaclient/client_agent.go#L283-L291

To be honest, we shouldn't have such checks in the code - we can't be sure that the native restore API will actually be a part of the 2025.3.0 release (but this is still only on master, so we are fine for now). It should be replaced by just the master version check, and then it would also fail for the `scylla-nightly:latest` image.

Another problem is that both SM implementation and the tests use the same check function for checking if given scylla API is exposed. This means that even if we have a problem with the check function, it will be ignored (both test and implementation will think that scylla API is not exposed, when in reality it is).

In general the check for scylla master version is fragile - that's why it broke down in the first place.
We could either fix it in several ways:
- check for `~dev` suffix or just the `dev` infix (quick fix)
- add test env variable dictating whether we run against master or not. The test code will use (`against_master_test_var || check_function()`) instead of just the `check_function()`. This way we would spot the tests which are running against master and are supposed to assume that all scylla API is exposed, but they don't (longer fix)

cc: @karol-kokoszka @VAveryanov8

Michal-Leszczynski:
This PR tackles #4484 by getting rid of using direct NodeInfo methods for checking scylla features availability.
The were problematic, because they used `version.MasterVersion` that doesn't work and it's difficult to make it work robustly for a production code.
Instead, this PR introduces feature gates:
- `ScyllaFeatureGate` - used in production - does not try to detect master version at all
- `ScyllaMasterFeatureGate` - used in tests - tries to detect master version by looking for `dev` substring (this method might not work in production for custom builds) 

This PR also removes the NodeInfo methods from the code, so that they are not used by mistake.

Finally, this PR allows for removing assumption that native restore will be ready to use in scylla 2025.3 without impacting the ability to test it against scylla master builds. This assumption is just guessing and we can't release any SM version with such a check.

Fixes #4484

### DIFF
--- a/pkg/cmd/scylla-manager/server.go
+++ b/pkg/cmd/scylla-manager/server.go
@@ -117,6 +117,7 @@ func (s *server) makeServices(ctx context.Context) error {
 		s.session,
 		s.config.Backup,
 		metrics.NewBackupMetrics().MustRegister(),
+		featuregate.ScyllaFeatureGate{},
 		s.clusterSvc.GetClusterName,
 		s.clusterSvc.Client,
 		s.clusterSvc.GetSession,
--- a/pkg/cmd/scylla-manager/server.go
+++ b/pkg/cmd/scylla-manager/server.go
@@ -14,6 +14,7 @@ import (
 	"github.com/scylladb/go-log"
 	"github.com/scylladb/gocqlx/v2"
 	config "github.com/scylladb/scylla-manager/v3/pkg/config/server"
+	"github.com/scylladb/scylla-manager/v3/pkg/featuregate"
 	"github.com/scylladb/scylla-manager/v3/pkg/metrics"
 	"github.com/scylladb/scylla-manager/v3/pkg/restapi"
 	"github.com/scylladb/scylla-manager/v3/pkg/schema/table"
@@ -102,6 +103,7 @@ func (s *server) makeServices(ctx context.Context) error {
 		s.session,
 		s.config.Repair,
 		metrics.NewRepairMetrics().MustRegister(),
+		featuregate.ScyllaFeatureGate{},
 		s.clusterSvc.Client,
 		s.clusterSvc.GetSession,
 		s.configCacheSvc,
--- a/pkg/cmd/scylla-manager/server.go
+++ b/pkg/cmd/scylla-manager/server.go
@@ -133,6 +133,7 @@ func (s *server) makeServices(ctx context.Context) error {
 		s.session,
 		s.config.Restore,
 		metrics.NewRestoreMetrics().MustRegister(),
+		featuregate.ScyllaFeatureGate{},
 		s.clusterSvc.Client,
 		s.clusterSvc.GetSession,
 		s.configCacheSvc,
--- a/pkg/cmd/scylla-manager/server.go
+++ b/pkg/cmd/scylla-manager/server.go
@@ -151,6 +151,7 @@ func (s *server) makeServices(ctx context.Context) error {
 		s.configCacheSvc,
 		s.logger.Named("one2onerestore"),
 		metrics.NewOne2OneRestoreMetrics().MustRegister(),
+		featuregate.ScyllaFeatureGate{},
 	)
 	if err != nil {
 		return errors.Wrapf(err, "one2onerestore service")
--- a/pkg/featuregate/featuregate.go
+++ b/pkg/featuregate/featuregate.go
@@ -36,8 +36,8 @@ func (fg ScyllaFeatureGate) NativeBackup(version string) (bool, error) {
 }
 
 // NativeRestore - scylla exposes /storage_service/restore API.
-func (fg ScyllaFeatureGate) NativeRestore(version string) (bool, error) {
-	return checkConstraints(version, ">= 2025.3")
+func (fg ScyllaFeatureGate) NativeRestore(_ string) (bool, error) {
+	return false, nil
 }
 
 // SkipCleanupAndSkipReshape - scylla supports skip_cleanup and skip_reshape params in /storage_service/sstables/{keyspace} API.
--- a/pkg/featuregate/featuregate.go
+++ b/pkg/featuregate/featuregate.go
@@ -1,0 +1,60 @@
+// Copyright (C) 2025 ScyllaDB
+
+package featuregate
+
+import (
+	"github.com/pkg/errors"
+	scyllaversion "github.com/scylladb/scylla-manager/v3/pkg/util/version"
+)
+
+// ScyllaFeatureGate is a helper for checking scylla feature availability based on scylla version.
+type ScyllaFeatureGate struct{}
+
+// RepairSmallTableOptimization - scylla supports small_table_optimization param in /storage_service/repair_async/{keyspace} API.
+func (fg ScyllaFeatureGate) RepairSmallTableOptimization(version string) (bool, error) {
+	return checkConstraints(version, ">= 6.0, < 2000", ">= 2024.1.5")
+}
+
+// TabletRepair - scylla exposes /storage_service/tablets/repair API.
+func (fg ScyllaFeatureGate) TabletRepair(version string) (bool, error) {
+	return checkConstraints(version, ">= 2025.1")
+}
+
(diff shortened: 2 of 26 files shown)


Michal-Leszczynski:
Not that I think about it, this solution does not allow to test SM with native restore outside of SM repo tests...
That's not good.

What about extending `scylla-manager.yaml` config with feature gates? Something like:
```yaml
# scylla_feature_gates allows to either:
# -1 - forbid SM from using the feature
# 0  - let SM decide on using the feature based on scylla version
# 1  - force SM to use the feature
# The default value for all features is 0 and it shouldn't be changed unless for testing
# or when trying to make an ad-hoc fix to an issue caused by the feature use.
scylla_feature_gates:
  native_restore: 0
```

Then we could replace `ScyllaMasterFeatureGate` by appropriate config setup.

Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 what about https://github.com/scylladb/scylla-manager/pull/4489#issuecomment-3084400550?
In general, this PR aims at a more general solution, but maybe it's an overkill.
The version checks are annoying and we could actually benefit from the `--method` flags in those cases like:
- `--method=native` means that SM **will not** perform scylla version check (it will verify the configuration)
- `--method=auto` means that SM **will** perform scylla version check
- `--method=rclone` is not affected

The same could be done for the backup.

karol-kokoszka:
> @karol-kokoszka @VAveryanov8 what about https://github.com/scylladb/scylla-manager/pull/4489#issuecomment-3084400550?
In general, this PR aims at a more general solution, but maybe it's an overkill.
The version checks are annoying and we could actually benefit from the --method flags in those cases like:

> --method=native means that SM will not perform scylla version check (it will verify the configuration)
--method=auto means that SM will perform scylla version check
--method=rclone is not affected
The same could be done for the backup.

So, on production you keep checking current Scylla version against what's hardcoded in SM.
But, as you use ScyllaMasterFeatureGate, you can keep testing features that are being implemented in SM against some "expected" Scylla versions.
IIUC, before you hardcode the exact Scylla version into check, you just validate it with the gh CI that it works and let the Scylla team to actually test their features against Scylla Manager.

That's fine.

And I agree they are annoying :)

The doubt you have about "--method" flag is what exactly ? You want to add this method to restore task ? If so, I'm completely fine with it.

Michal-Leszczynski:
> So, on production you keep checking current Scylla version against what's hardcoded in SM.
But, as you use ScyllaMasterFeatureGate, you can keep testing features that are being implemented in SM against some "expected" Scylla versions.

Right - no flaky master version checks are performed on production, only the proper version checks.
The flaky checks are performed in our repo tests with the mentioned `ScyllaMasterFeatureGate`. 

> IIUC, before you hardcode the exact Scylla version into check, you just validate it with the gh CI that it works and let the Scylla team to actually test their features against Scylla Manager.

To be more precise - before I hardcode the exact Scylla version into check, I want to validate with the gh CI that it works with some proper scylla release - not just with scylla master. I don't want to end up in a situation where:
- new feature is being developed in scylla and sm is supposed to use it
- we validate that it works with `ScyllaMasterFeatureGate`
- we added the next minor scylla relase to the check for this feature
- we make sm release with those changes
- new minor scylla relase does not contain this feature
- the feature check is broken in already released sm is broken which results in this sm relase not working with the newest scylla

Getting back to the original concern - it would be good to have something like `ScyllaMasterFeatureGate` (limited to sm repo tests) for testing sm with sct and dtests (performed by QA and core devs).

> The doubt you have about "--method" flag is what exactly ? You want to add this method to restore task ? If so, I'm completely fine with it.

In general we need to add it for the purpose of testing. I was just wondering if it could also solve the problem with being able to test native restore outside of sm repo tests when we don't have any properly released scylla version supporting the native restore.

karol-kokoszka:
> > So, on production you keep checking current Scylla version against what's hardcoded in SM.
> > But, as you use ScyllaMasterFeatureGate, you can keep testing features that are being implemented in SM against some "expected" Scylla versions.
> 
> Right - no flaky master version checks are performed on production, only the proper version checks. The flaky checks are performed in our repo tests with the mentioned `ScyllaMasterFeatureGate`.
> 
> > IIUC, before you hardcode the exact Scylla version into check, you just validate it with the gh CI that it works and let the Scylla team to actually test their features against Scylla Manager.
> 
> To be more precise - before I hardcode the exact Scylla version into check, I want to validate with the gh CI that it works with some proper scylla release - not just with scylla master. I don't want to end up in a situation where:
> 
> * new feature is being developed in scylla and sm is supposed to use it
> * we validate that it works with `ScyllaMasterFeatureGate`
> * we added the next minor scylla relase to the check for this feature
> * we make sm release with those changes
> * new minor scylla relase does not contain this feature
> * the feature check is broken in already released sm is broken which results in this sm relase not working with the newest scylla
> 
> Getting back to the original concern - it would be good to have something like `ScyllaMasterFeatureGate` (limited to sm repo tests) for testing sm with sct and dtests (performed by QA and core devs).

Got it. But QA are rather not testing against master builds, but against RCs ? 
IIUC, with the current implementation, SM will behave as expected (testing the expected functionality) against master, but not RC ?

> 
> > The doubt you have about "--method" flag is what exactly ? You want to add this method to restore task ? If so, I'm completely fine with it.
> 
> In general we need to add it for the purpose of testing. I was just wondering if it could also solve the problem with being able to test native restore outside of sm repo tests when we don't have any properly released scylla version supporting the native restore.

I think I got it now. Method makes much more sense now. It makes this split between ScyllaMasterFeatureGate and ScyllaFeatureGate bit unnecessary ? 
Just leave default to old way (not native one).

Michal-Leszczynski:
> Got it. But QA are rather not testing against master builds, but against RCs ?

SM related QAs are usually testing against proper releases (perhaps sometimes RCs as well).
Scylla core related QAs are usually testing against scylla master builds. 

> IIUC, with the current implementation, SM will behave as expected (testing the expected functionality) against master, but not RC ?

By the current you mean current SM master or the SM master with the current changes from this PR?

1. Current SM master
As described in the issue, SM checks against scylla master version don't work.
In terms of the native restore - this check checks for master (always fails) or for version >= 2025.3. Right now master is 2025.3/4, so the second check passes, but we don't have any guarantee that the native restore will actually end up in 2025.3. That's why even though it's currently working*, it's dangerous to have checks like >= 2025.3.

2. SM master with the current changes from this PR
The master checks and the >= 2025.3 check will be removed from the native restore check. The native restore check will pass only in SM repo tests (because of `ScyllaMasterFeatureGate`), but it will fail in SCT and DTests even when they are executed against scylla master build.

Michal-Leszczynski:
> I think I got it now. Method makes much more sense now. It makes this split between ScyllaMasterFeatureGate and ScyllaFeatureGate bit unnecessary ?
Just leave default to old way (not native one).

That's right, but:
- it is not a general solution - feature checks which don't have something like `--method` flag won't be able to utilize that (but right now we don't have such features)
- it changes the semantic of `--method=native` (it won't perform the version check like it used to, so it gives less upfront validation)

@karol-kokoszka so the main question is: are we fine with those 2 concerns.
Since `ScyllaFeatureGate` / `ScyllaMasterFeatureGate` don't solve the whole problem, I'm fine with not merging them (we can always bring them back if a need for more elaborate feature check is needed.)

PR #4503, opened 2025-07-30
tmp: test sm 3.6.0 against scylla 2025.2.0

Michal-Leszczynski:


Michal-Leszczynski:
@VAveryanov8 @mikliapko FYI: branch-3.6 works with scylla 2025.2.0.

PR #4518, opened 2025-08-06
Always fill options on provider registration

Michal-Leszczynski:
This reverts part of the a64bf059 commit which moved provider options filling to the top of agent root command in order to utilize filled options in node info handler.
The problem is that it also removed provider options filling from functions registering providers. It worked fine for the agent root command, but it resulted in not filling provider options in other sm-agent commands (e.g. 'scylla-manager-agent check-location').

### DIFF
--- a/pkg/rclone/providers.go
+++ b/pkg/rclone/providers.go
@@ -49,6 +49,7 @@ func RegisterS3Provider(opts S3Options) error {
 		backend = "s3"
 	)
 
+	opts.AutoFill()
 	if err := opts.Validate(); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


karol-kokoszka:
How did you test that it works fine after this revert ?

Michal-Leszczynski:
> How did you test that it works fine after this revert ?

I plan to do it by running the same test that discovered the issue (in siren, as SM repo lacks infra for such test).

Michal-Leszczynski:
@karol-kokoszka checks have passed (not important, because this change does not affect our test setup). Can we merge so I can build SM 3.6 and run the test which discovered this issue against it?

karol-kokoszka (approved):

PR #4519, opened 2025-08-06
Always fill options on provider registration

Michal-Leszczynski:
This reverts part of the a64bf059 commit which moved provider options filling to the top of agent root command in order to utilize filled options in node info handler.
The problem is that it also removed provider options filling from functions registering providers. It worked fine for the agent root command, but it resulted in not filling provider options in other sm-agent commands (e.g. 'scylla-manager-agent check-location').


karol-kokoszka:
How does it differ from http://github.com/scylladb/scylla-manager/pull/4518 ?

Issue #4512, opened 2025-08-04
Extend backup with alternator schema in alternator friendly format


Michal-Leszczynski:
Requires #4511.

Right now we back up schema in a form of the output of `DESCRIBE SCHEMA WITH INTERNALS`.
This schema is not suitable for alternator keyspace restoration (see https://github.com/scylladb/scylladb/issues/19112).
SM should use backup and restore alternator schema (mostly) from the output of alternator API calls.

For backing up alternator schema, SM needs to use:
- `ListTables` - to get the list of all tables. In theory, if you have a large number of tables, you may need to call ListTables multiple times to retrieve several pages.
- ` DescribeTable` - to describe each table listed by ListTables separately
- `ListTagsOfResource` - need to call this for each table as well: Although a table can be created with tags (or they can be added later), these tags aren't listed in DescribeTable and you need this separate ListTagsOfResource call to retrieve those tags. Note that users can use these tags for their own uses, but some tags have special meanings for the system (e.g., this is how you set write isolation mode or tablets for a table).

Here is a simple example code prepared by @nyh:
```python
#!/usr/bin/python
import boto3

alternator_url = 'http://localhost:8000/'
dynamodb = boto3.resource('dynamodb', endpoint_url=alternator_url, region_name='nonexistent')
client = dynamodb.meta.client
   
# DynamoDB's ListTables request returns up to a single page of table names (e.g., up to 100) and it is up to the caller to call
# it again and again to get the next page. This is a utility function which calls it repeatedly as much as necessary to get the entire list.
def list_tables(limit=100):
    ret = []
    pos = None
    while True:
        if pos:
            page = client.list_tables(Limit=limit, ExclusiveStartTableName=pos);
        else:
            page = client.list_tables(Limit=limit);
        results = page.get('TableNames', None)
        ret = ret + results
        newpos = page.get('LastEvaluatedTableName', None)
        if not newpos:
            break;
        pos = newpos
    return ret

for table_name in list_tables():
    print(f'"{table_name}": ')
    desc = client.describe_table(TableName=table_name)['Table']
    # Although CreateTable can create a table with tags, DescribeTable cannot
    # list them, and a separate ListTagsOfResource call is necessary to list
    # them. Let's add them to the printed object as a "Tags" field
    desc['Tags'] = client.list_tags_of_resource(ResourceArn=desc['TableArn'])['Tags']
    print(desc)
    print()
```

[...]

PR #4525, opened 2025-08-07
Backupspec: add alternator schema format and path

Michal-Leszczynski:
This PR extends `backupspec` with alternator schema format and paths needed for backing up alternator schema.

Refs #4528
Refs #4512


karol-kokoszka:
yes, referencing with commit hash is fine

karol-kokoszka (approved):

PR #4541, opened 2025-08-19
Force update config cache before reading it

### LINKED ISSUE #4524: Force update config cache before reading it
Since config cache has been introduced for the health check reasons, its usage started spreading to other places as well.
Using config cache instead of aggregating node info from scratch gives better performance, but creates problems when config cache lags behind topology changes. Because of that, we need to always force update config cache when executing tasks which require up to date node info (backup/repair/restore).

This was discovered by running [operator tests](https://prow.scylla-operator.scylladb.com/view/gs/scylla-operator-prow/pr-logs/pull/scylladb_scylla-operator/2868/pull-scylla-operator-master-e2e-gke-multi-datacenter-parallel/1953386425575542784):
```
{Timed out after 900.001s.
The function passed to Eventually failed at github.com/scylladb/scylla-operator/test/e2e/utils/verification/scylladbmanager.go:34 with:
Backup task "cdcade07-09c9-4e07-8af6-6190ac62d481" is not done, status: "ERROR", cause: "no node config for 10.125.32.114; no node config for 10.125.32.116; no node config for 10.125.32.118; no node config for 10.110.192.28; no node config for 10.110.192.34; no node config for 10.110.192.36".
Expected
    <string>: ERROR
to equal
    <string>: DONE failed [FAILED] Timed out after 900.001s.
The function passed to Eventually failed at github.com/scylladb/scylla-operator/test/e2e/utils/verification/scylladbmanager.go:34 with:
Backup task "cdcade07-09c9-4e07-8af6-6190ac62d481" is not done, status: "ERROR", cause: "no node config for 10.125.32.114; no node config for 10.125.32.116; no node config for 10.125.32.118; no node config for 10.110.192.28; no node config for 10.110.192.34; no node config for 10.110.192.36".
Expected
    <string>: ERROR
to equal
    <string>: DONE
In [It] at: github.com/scylladb/scylla-operator/test/e2e/set/scylladbmanagertask/multidatacenter/scylladbmanagertask_scylladbcluster_globalmanager_object_storage.go:190 @ 08/07/25 10:14:48.91
}
``` 

cc: @rzetelskik @karol-kokoszka @mikliapko

Michal-Leszczynski:
Even though SM does not handle topology changes during backup/restore/repair tasks, we still should respect those which finish right before task is executed, but before config cache is updated.

Fixes #4524

### DIFF
--- a/pkg/service/backup/service.go
+++ b/pkg/service/backup/service.go
@@ -165,7 +165,7 @@ func (s *Service) targetFromProperties(ctx context.Context, clusterID uuid.UUID,
 	}
 
 	if p.Method == MethodNative {
-		if err := s.validateHostNativeBackupSupport(clusterID, liveNodes, p); err != nil {
+		if err := s.validateHostNativeBackupSupport(ctx, clusterID, liveNodes, p); err != nil {
 			return Target{}, err
 		}
 	}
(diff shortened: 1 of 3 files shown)


Michal-Leszczynski:
Failures against `scylla-nightly:latest` are caused by #4542.

VAveryanov8 (approved):
Repair and 1-1-restore are not actually using configCache, right?

Michal-Leszczynski:
> Repair and 1-1-restore are not actually using configCache, right?

They are not - I tracked all usages of `configCache.ReadAll` and added `configCache.ForceUpdateCluster` to them.

Issue #4510, opened 2025-08-04
Allow for setting alternator credentials for a cluster


Michal-Leszczynski:
Communicating over alternator API requires different credentials than communicating over CQL.
It is described in [alternator docs](https://docs.scylladb.com/manual/stable/alternator/compatibility.html#authentication-and-authorization), but essentially CQL credentials consist of CQL username and password, where alternator credentials consist of CQL username and hashed password which needs to be retrieved from `system.roles`.
Because of that, we need to add a way for registering cluster with `--alternator-access-key-id` and `--alternator-secret-access-key` (we abstract from the fact that it's CQL username and hashed password in order to make it more independent). Those credentials should be kept and processed similar to the regular CQL credentials:
- should be stored in secrets store
- shouldn't ever be returned to the client
- should be possible to delete them via `sctool cluster update --delete-alternator-credentials`
- etc.

Side note: implementing this issue will also allow for unblocking alternator query ping in health check (#4036).

PR #4543, opened 2025-08-19
Handle alternator creds

Michal-Leszczynski:
Alternator credentials need to be handled similarly to CQL credentials. User should be able to add them with `POST /clusters/`. User should be able to delete them with `DELETE /cluster/<cluster ID>/`. User should be able to check if the credentials are set with `GET /clusters/`.

Refs #4510


VAveryanov8 (approved):
LGTM!

PR #4544, opened 2025-08-19
Migrate to aws sdk go v2

Michal-Leszczynski:
As we want to use aws-sdk-go-v2 in alternator backup/restore, we should also use it in tests and health check.


Michal-Leszczynski:
Resolved conflicts in `go.mod`.

PR #4550, opened 2025-08-20
.github: bump tested scylla versions

Michal-Leszczynski:
This PR updates tested scylla versions.
It also makes necessary adjustments to tests.


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 please take a look

PR #4553, opened 2025-08-20
Add alternator client

### LINKED ISSUE #4511: Configure and create alternator client in cluster service
Requires #4510.

Cluster service is responsible for creating scylla client and CQL session and should be extended with creating alternator client as well.

Note that it's not possible to configure alternator server to authenticate client TLS certs (no way of providing custom CA in alternator [configuration](https://docs.scylladb.com/manual/stable/reference/configuration-parameters.html#confval-alternator_encryption_options)).

On the other hand, it's still possible to validate alternator server TLS cert by SM client. The thing is that we never validate scylla server TLS cert when communicating with it (we set `InsecureSkipVerify: true` for both scylla client and CQL session). I guess that's because:
- we don't care that much, it's server that contains important user data, not SM
- it would make SM and cluster setup more complicated

So in the scope of this issue, I guess we are fine with not validating alternator server TLS cert - perhaps we will want to change it in the future for scylla client and CQL session as well.

Michal-Leszczynski:
This PR adds alternator client configuration and creation to cluster service.

It is rebased on #4543 and #4544.

Fixes #4511


pkg/service/cluster/service_integration_test.go
```diff
@@ -228,6 +230,54 @@ func ipsNotInSlice(a []string, b []string) []string {
...
+		t.Fatal(err)
+	}
+
+	client, err := s.GetAlternatorClient(context.Background(), c.ID, ManagedClusterHost())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	const tableName = ".scylla.alternator.system_schema.tables"
+	_, err = client.Scan(context.Background(), &dynamodb.ScanInput{
```
karol-kokoszka, line 272: nit: could you please harden the assertion a bit and validate if the scanresult is not nil ?

karol-kokoszka (approved):

pkg/service/cluster/service_integration_test.go
Michal-Leszczynski: sure

PR #4555, opened 2025-08-22
Explicitly disable tablets for SimpleStrategy replication

paszkow:
Currently, tablets enabled by default are not enforced. Thus, creating keyspaces that do not support tablets, silently disable them and use vnodes.

This patch adds `AND tablets = {'enabled': false}` to explicitly disable tablets for keyspaces using SimpleStrategy replication.

This is a preparation step for https://github.com/scylladb/scylladb/pull/25342 that changes the current behaviour and forces users to explicitly disable tablets when enabled by default.

Refs: https://github.com/scylladb/scylladb/issues/25340

### DIFF
--- a/pkg/cmd/scylla-manager/db.go
+++ b/pkg/cmd/scylla-manager/db.go
@@ -61,7 +61,8 @@ func createKeyspace(ctx context.Context, c config.Config, logger log.Logger) err
 	return session.Query(mustEvaluateCreateKeyspaceStmt(c)).Exec()
 }
 
-const createKeyspaceStmt = "CREATE KEYSPACE {{.Keyspace}} WITH replication = {'class': 'SimpleStrategy', 'replication_factor': {{.ReplicationFactor}}}"
+const createKeyspaceStmt = "CREATE KEYSPACE {{.Keyspace}} WITH replication = {'class': 'SimpleStrategy', 'replication_factor': {{.ReplicationFactor}}}" +
+	" AND tablets = {'enabled': false}"
 
 func mustEvaluateCreateKeyspaceStmt(c config.Config) string {
 	t := template.New("")
--- a/pkg/scyllaclient/client_scylla_integration_test.go
+++ b/pkg/scyllaclient/client_scylla_integration_test.go
@@ -130,7 +130,11 @@ func TestClientDescribeRingIntegration(t *testing.T) {
 	for i := range testCases {
 		tc := testCases[i]
 		t.Run(tc.name, func(t *testing.T) {
-			if err := clusterSession.ExecStmt(fmt.Sprintf("CREATE KEYSPACE %s WITH replication = %s", tc.name, tc.replicationStmt)); err != nil {
+			ksStmt := fmt.Sprintf("CREATE KEYSPACE %s WITH replication = %s", tc.name, tc.replicationStmt)
+			if tc.replication == scyllaclient.SimpleStrategy {
+				ksStmt += " AND tablets = {'enabled': false}"
+			}
+			if err := clusterSession.ExecStmt(ksStmt); err != nil {
 				t.Fatal(err)
 			}
 			defer func() {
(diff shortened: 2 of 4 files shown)


pkg/cmd/scylla-manager/db.go
```diff
@@ -58,7 +58,8 @@ func createKeyspace(ctx context.Context, c config.Config, logger log.Logger) err
 	return session.Query(mustEvaluateCreateKeyspaceStmt(c)).Exec()
 }
 
-const createKeyspaceStmt = "CREATE KEYSPACE {{.Keyspace}} WITH replication = {'class': 'SimpleStrategy', 'replication_factor': {{.ReplicationFactor}}}"
+const createKeyspaceStmt = "CREATE KEYSPACE {{.Keyspace}} WITH replication = {'class': 'SimpleStrategy', 'replication_factor': {{.ReplicationFactor}}}" +
+	" AND tablets = {'enabled': false}"
```
Michal-Leszczynski: This unfortunately won't work with 2024.1, which is currently still supported by SM.
For this scylla version, we can't use the `tablets = {'enabled': false}` syntax.
Tests are passing because we don't have real e2e tests in SM repo, and those tests use different functions for creating SM keyspace. If we were to execute SCT/DTests with this change and scylla 2024.1, we would fail on SM startup.

.github/workflows/integration-tests-2024.1.21-IPV4.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
                 ssl-enabled: ${{ env.ssl-enabled }}
                 tablets: ${{ env.tablets }}
             - name: Run tests
-              run: make pkg-integration-test IP_FAMILY=${{ env.ip-family }} SSL_ENABLED=${{ env.ssl-enabled}} BACKUP_METHOD=${{ env.backup-method }} RESTORE_METHOD=${{ env.restore-method }} PKG=./pkg/service/backup
+              run: make pkg-integration-test TABLETS=${{ env.tablets }} IP_FAMILY=${{ env.ip-family }} SSL_ENABLED=${{ env.ssl-enabled}} BACKUP_METHOD=${{ env.backup-method }} RESTORE_METHOD=${{ env.restore-method }} PKG=./pkg/service/backup
```
Michal-Leszczynski: In general, we have a simple script for generating workflows with tests against different scylla versions. In order to modify them, `.github/cfg/integration-test-core.yaml` should be adjusted and `go run .github/cfg/main.go` should be executed.

pkg/scyllaclient/client_scylla_integration_test.go
```diff
@@ -129,7 +130,12 @@ func TestClientDescribeRingIntegration(t *testing.T) {
 	for i := range testCases {
 		tc := testCases[i]
 		t.Run(tc.name, func(t *testing.T) {
-			if err := clusterSession.ExecStmt(fmt.Sprintf("CREATE KEYSPACE %s WITH replication = %s", tc.name, tc.replicationStmt)); err != nil {
+			ksStmt := fmt.Sprintf("CREATE KEYSPACE %s WITH replication = %s", tc.name, tc.replicationStmt)
+			if tablets := os.Getenv("TABLETS"); tablets == "enabled" && tc.replication == scyllaclient.SimpleStrategy {
+				ksStmt += " AND tablets = {'enabled': false}"
+			}
```
Michal-Leszczynski: In general, I don't really like relying on checking env variables in tests.
We do it in many places for convenience and out of laziness, because we don't really have vnode or tablet specific tests - we just run all the tests twice, once with tablets as default replication type, once with vnodes, but this is dirty and results in rerunning many tests for no reason at all (as they don't depend on tablet vs vnode).
If possible, it should be replaced with checking scylla version / default replication type.

Michal-Leszczynski:
@paszkow I left 3 comments, 2 are more or less nits, but https://github.com/scylladb/scylla-manager/pull/4555#discussion_r2652547324 is a real one.

After spending some time on understanding the issue on the scylla side and the proposed implementation, I came to the conclusion that it would be better to move away from SimpleStrategy in favor of NetworkTopologyStrategy. I created [PR](https://github.com/scylladb/scylla-manager/pull/4727) with this approach and described my reasoning with more details there. Please take a look, and again, sorry for not reviewing/working on this item for a long time.

PR #4575, opened 2025-09-24
Add test for changing backup method

Michal-Leszczynski:
This commit adds TestRestoreFullChangingMethodIntegration, which verifies that changing backup method from rclone to native and so forth results in a correct restore.
It also removes new task ID generation on every backup/restore run, as it shouldn't be needed and prevents from verifying retention policy behavior.

Refs https://github.com/scylladb/siren/issues/14212


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 this PR is ready for review!

PR #4578, opened 2025-09-25
feat(sstable): add `ms` to the set of known SSTable format versions

michoecho:
Prepare for the addition of a new SSTable format version, `ms`, in an upcoming Scylla version.

Refs https://github.com/scylladb/scylladb/pull/26215

### DIFF
--- a/pkg/sstable/sstable_naming.go
+++ b/pkg/sstable/sstable_naming.go
@@ -16,8 +16,8 @@ import (
 )
 
 var (
-	regexNewLaMx = regexp.MustCompile(`(la|m[cde])-([^-]+)-(\w+)-(.*)`)
-	regexLaMx    = regexp.MustCompile(`(la|m[cde])-(\d+)-(\w+)-(.*)`)
+	regexNewLaMx = regexp.MustCompile(`(la|m[cdes])-([^-]+)-(\w+)-(.*)`)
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
Just a general question related to testing - will `ms` sstables start appearing on their own as regular table's sstables, or do we need to enable some feature in `scylla.yaml` or use some table/view create option in order to observe them?

michoecho:
> Just a general question related to testing - will `ms` sstables start appearing on their own as regular table's sstables, or do we need to enable some feature in `scylla.yaml` or use some table/view create option in order to observe them?

Currently the plan is that the new format will become the default, which means that they will start appearing on their own.

But if we discover some problems with the new format during testing, we might eventually make it opt-in or disable it altogether.

michoecho:
> > How do I get this merged?
> 
> This PR contains changes needed by new scylla version

Yes.

> and was created because of a failed dtest, right?

Yes.

> In general the changes in this PR are small and easy to verify, but I guess that it would be nice to see that the dtest which previously failed works fine with SM version from this branch.
> Did we execute it against SM build from this branch? 

I didn't.

> If you provide me with a link to the failed dtest, then I can build SM and execute it.

The dtest is `manager_restore_tests.py::TestRestoreWithEaR::test_basic_restore_with_kms_rotate_key`

Michal-Leszczynski (approved):

PR #4579, opened 2025-09-25
Fix alternator backup/restore for scylla 2024.1

Michal-Leszczynski:
In scylla 2024.1 creating GSIs on existing table is not possible.
Because of that, we need to implement the same workaround as for
restoring LSIs, so just don't interact with them during data restoration.

The test for alternator backup/restore was mistakenly skipped because of
the assumption that if we can't back up alternator schema in a form of schema file,
then we can't restore it. That's not true for scylla 2024.1 where we can safely back up and restore
alternator schema from sstables, but we should still properly use alternator
api for handling other things (e.g. dropping and recreating views).


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 this PR is ready for review!

Issue #4589, opened 2025-10-15
TestRestoreTablesRestartAgentsIntegration is flaky


Michal-Leszczynski:
From time to time, we see some failures in `TestRestoreTablesRestartAgentsIntegration`.
This test basically restarts agents in the middle of restore - sometimes agents get up quickly enough for it to be handled by regular timeout/retry mechanisms, sometimes they don't.

The test was originally copied from backup `TestBackupResumeIntegration` test. Backup procedure implementation has some additional retries in case sstable upload has failed (it is retried up to 10 hardcoded times):
```go
	// Upload sstables
	var (
		sstablesPath = w.remoteSSTableDir(h, d)
		dataDst      = h.Location.RemotePath(sstablesPath)
		dataSrc      = d.Path
		retries      = 10
	)
	for range retries {
		if err := w.uploadDataDir(ctx, h, dataDst, dataSrc, d); err != nil {
			if errors.Is(err, errJobNotFound) {
				continue
			}
			return errors.Wrapf(err, "copy %q to %q", dataSrc, dataDst)
		}
		break
	}
```
We don't have such advanced retry policies in restore.

If we wanted to add such thing to restore, we would need to take care of a few additional concerns.
Backup upload and restore download with rclone are idempotent (it's more about syncing dirs). I'm not sure if the same thing can be said about l&s - are we sure that when it fails we can safely retry it without any data loss (l&s works on sstables in snapshot dir and deletes them on the fly). Even if that's not the case, we could retry the whole batch with the download phase. Same could be performed for the native restore, where download and l&s are combined into the same api call.

Having said that, I'm not sure if we need to support agent restart during restore. Also, what has been described above helps only when the restart happened during download and l&s phase, not any other phase (but this is probably the longest one).

So we could either add some more retries, or just delete this test.
It has a low priority, so this issue is more for visibility and potential discussion.

PR #4602, opened 2025-10-21
Don't access unset Run.EndTime

### LINKED ISSUE #4601: SM panics when accessing unset task end time
Here is a fragment of SM code responsible for running and logging task execution:

```go
r := newRunFromTaskInfo(ti)
	s.runs[ti.TaskID] = *r
	s.mu.Unlock()

	defer func() {
		s.mu.Lock()
		delete(s.runs, ti.TaskID)
		s.mu.Unlock()
	}()

	runCtx := log.WithTraceID(ctx)
	logger := s.logger.Named(ti.ClusterID.String()[0:8])

	if ti.TaskType != HealthCheckTask {
		logger.Info(runCtx, "Run started",
			"task", ti,
			"retry", ctx.Retry,
		)
		defer func() {
			if r.Status == StatusError {
				logger.Info(runCtx, "Run ended with ERROR",
					"task", ti,
					"status", r.Status,
					"cause", r.Cause,
					"duration", r.EndTime.Sub(r.StartTime), 
				)
			} else {
				logger.Info(runCtx, "Run ended",
					"task", ti,
					"status", r.Status,
					"duration", r.EndTime.Sub(r.StartTime),
				)
			}
		}()
	}

	if !ok {
		return util.ErrNotFound
	}
	if err := s.putRunAndUpdateTask(r); err != nil {
		return errors.Wrap(err, "put run")
	}
        ...
```

`Run.EndTime` is a pointer, and we can clearly see that we access it in the deferred function which logs run statistics.
Unfortunately, this deferred function might be executed even when no run occurred (e.g. when encountering an error on SM DB insert), so with unset `Run.EndTime` which results in panic. 
This issue has been there for a really long time, so it doesn't seem like a regression.

cc: @pogzie

Michal-Leszczynski:
Accessing `Run.EndTime` in deferred function caused panic in cases where the run couldn't be executed because of other errors.

The issue is obvious, but we don't have tools for easily reproducing it with existing codebase.
I tested the fix manually by always returning an error from `putRunAndUpdateTask`.
Without the fix, it panics when `putRunAndUpdateTask` returns an error, with the fix it works fine and just logs `unknown` run duration.

Fixes #4601

### DIFF
--- a/pkg/service/scheduler/model.go
+++ b/pkg/service/scheduler/model.go
@@ -366,6 +366,15 @@ type Run struct {
 	EndTime   *time.Time `json:"end_time,omitempty"`
 }
 
+// Duration checks if both EndTime and StartTime are set correctly and returns
+// their difference formatted as a string. Otherwise, it returns "unknown" duration.
+func (r Run) Duration() string {
+	if r.EndTime != nil && !r.EndTime.IsZero() && !r.StartTime.IsZero() {
+		return r.EndTime.Sub(r.StartTime).String()
+	}
+	return "unknown"
+}
+
 func newRunFromTaskInfo(ti taskInfo) *Run {
 	var id uuid.UUID
 	if ti.TaskType == HealthCheckTask {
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
[Failure](https://github.com/scylladb/scylla-manager/actions/runs/18680734473/job/53261075265?pr=4602) is a known [flake](https://github.com/scylladb/scylla-manager/issues/4589).

PR #4616, opened 2025-10-24
docs: update supported OS versions of Upgrade guide

mikliapko:
Removed some versions which are EOL and added some relevant versions we test with.


Michal-Leszczynski:
@mikliapko this should be backported to SM 3.7, right?

PR #4629, opened 2025-11-10
Be more resilient to spaces in cpuset definition

### LINKED ISSUE #4627: sm-agent fails to parse scylla cpu config in cloud
At startup, sm-agent tries to parse scylla cpu config, so that it knows free cpus to which it can be pinned.
It reads `/etc/scylla.d/cpuset.conf` and uses a regex for extracting interesting information.
The problem is that in the cloud (not sure if everywhere) this file looks like:
```
# DO NO EDIT
# This file should be automatically configure by scylla_cpuset_setup
#
# CPUSET="--cpuset 0 --smp 1"
CPUSET="--cpuset 0-1 "
```
Agent parses it line by line, and unfortunatelly `CPUSET="--cpuset 0-1 "` does not match its regex.
It would be matched if not for the last space before the double quotes (`CPUSET="--cpuset 0-1"`)...

I'm not sure what is the contract on this file specification,perhaps we had one and it was broken on either scylla or siren part, but we should just update mentioned regex to be more resilient to random white spaces in the cpuset definition...

cc: @mykaul @karol-kokoszka

Michal-Leszczynski:
CI will fail until https://github.com/scylladb/scylla-manager/pull/4630 is merged.

Fixes #4627

### DIFF
--- a/v3/pkg/util/cpuset/cpuset.go
+++ b/v3/pkg/util/cpuset/cpuset.go
@@ -21,7 +21,7 @@ const (
 	scyllaConfigFile = "/etc/scylla.d/cpuset.conf"
 	// cpuSetPattern matches the configuration line in the scyllaConfigFile ex.
 	// CPUSET="--cpuset 0 --smp 1".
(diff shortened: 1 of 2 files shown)


tarzanek (approved):
lgtm

Issue #4631, opened 2025-11-10
docs: Issue on page Restore tables


annastuchlik:
I would like to report an issue on page https://manager.docs.scylladb.com/branch-3.7/restore/restore-tables

### Problem

The warning mentions ScyllaDB 6.0:

> Restoring data related to authentication (system_auth) and service levels (system_distributed.service_levels) is not supported in ScyllaDB 6.0.

We no longer support OSS.
And we definitely don't support version 6.0.
The table name system_distributed.service_levels is changing its name: https://github.com/scylladb/scylladb/pull/26556

###  Suggest a fix

I think this warning should be removed, unless it applies to versions other than Open Source.
If it does apply, remove the version name "6.0".

This update should also be backported to Manager previous versions, as we don't want to imply that we still support Open Source versions.

annastuchlik:
@Michal-Leszczynski I think you're The One who can help me with that.

PR #4639, opened 2025-11-17
Backport #4631 for branch-3.6

Michal-Leszczynski:
Refs #4631


Michal-Leszczynski:
Setting up testing deps fails because we use `@latest` version in `install_dependencies.sh`:
```
go install github.com/go-enry/go-license-detector/v4/cmd/license-detector@latest
```
which does not work with the go version from this branch.

PR #4727, opened 2025-12-29
Move away from simple strat for SM keyspace

Michal-Leszczynski:
Alternative approach to #4555.

While the other PR focused on keeping SimpleStrategy and working around its limitations, this PR goes into direction of using NetworkTopologyStrategy instead. 

The reasoning for that is:
- this change targets just the default SM keyspace - user can create the keyspace manually if needed
- for default single node SM DB those strategies are equivalent
- for single DC SM DB they are also equivalent
- for multi DC SM DB we already recommend user to create SM keyspace manually with desired replication
- SimpleStrategy is not recommended for production in general
- if we keep SimpleStrategy, we might end up with a similar issue in some time    

More details in the commit messages.

<details>
<Summary> Merged commit messages </Summary>

SimpleStrategy is not recommended for the production environments.
It also does not support tablets, which are the way towards scylla
is moving (in some distant time, vnodes might even be deprecated).
Because of that, it also runs into unexpected problems like
https://github.com/scylladb/scylla-manager/pull/4555.

Because of those reasons, we should switch from keeping SM data in
SimpleStrategy to NetworkTopologyStrategy keyspace by default.

For the default, single local node SM DB cluster, both strategies
result in the single node containing all of SM data. For single DC
SM DB cluster they are also equivalent. On the other hand, for multi
DC SM DB cluster (really not recommended...), NetworkTopologyStrategy
results in keeping replication_factor replicas per DC and not per cluster.
This is theoretically problematic, but such setup is also problematic for
SimpleStrategy, which would still replicate the data in all DCs, so it's
not a new problem. Moreover, this change only defines the default SM
keyspace creation. In case user has some non-default (perhaps not recommended)
SM DB setup, they can always create SM keyspace manually before starting
SM server to configure it to their liking.

Because of https://github.com/scylladb/scylla-manager/pull/4555, but also because
it's better to test recommended keyspace replication strategy, we need to replace
SimpleStrategy with NetworkTopologyStrategy in tests wherever possible.
In case test is aiming to run against SimpleStrategy specifically, it needs to be
adjusted so that it ensures that keyspace with SimpleStrategy does not use tablets.

</details>

### DIFF
--- a/pkg/scyllaclient/client_scylla_integration_test.go
+++ b/pkg/scyllaclient/client_scylla_integration_test.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2017 ScyllaDB
+// Copyright (C) 2025 ScyllaDB
 
 //go:build all || integration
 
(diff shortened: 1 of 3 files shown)


VAveryanov8 (approved):
Seems reasonable to me!

I'm curios, how how will it behave with replace manager procedure in siren, e.g. when the scylla-manager schema is restored from the self-backup?

Michal-Leszczynski:
@paszkow does the approach presented in this PR make sense to you?

PR #4756, opened 2026-02-09
Get rid of LWT query updating STOPPING run status

### LINKED ISSUE #4754: Get rid of LWT query when stopping the task
In general, SM does not use advanced scylla features like views, cdc, etc.
Because of that, it is safe in terms of using either tablet or vnode SM keyspace, as both of those should support all of its use cases.
There is one small missed detail here - SM performs LWT query when stopping the task:
```go
func (s *Service) updateRunStatus(r *Run) error {
	// Only update if running as there is a race between manually stopping
	// a run and the run returning normally.
	return table.SchedulerTaskRun.
		UpdateBuilder("status").
		If(qb.EqNamed("status", "from_status")).
		Query(s.session).
		BindStructMap(r, qb.M{"from_status": StatusRunning}).
		ExecRelease()
}
```
This results in errors when stopping task with SM with SM keyspace repliacted with tablets and SM DB with scylla version not supporting LWT for tablets.

In general, this use case is not critical and there should be some way of handling `STOPPING` status without LWT query.

UPDATE:

It looks like the current implementation is not working as expected.
When stopping the task, mentioned LWT query is used to update `scheduler_task_run` table only, but not `scheduler_task` table. This means that it's not possible to see `STOPPING` task status when running `sctool tasks` (it's only visible when inspecting individual runs with `sctool info`). When task finishes running, it updates both of those tables.

cc: @mikliapko

Michal-Leszczynski:
The original cause of this change was to get rid of LWT queries in SM codebase, so that we don't rely on specific SM ks replication and SM DB version (e.g. LWT is not supported for tablets in 2025.3). To achieve that, lightweight internal synchronization mechanism is added to scheduler svc.

This commit also extends StopTask test coverage, which resulted in discovering that task status STOPPING was never applied on task stop (it was applies only to task run, so it wasn't visible with `sctool tasks` command). This was an overlook that was fixed replacing `updateRunStatus` with `putRunAndUpdateTask` in StopTask.

Fixes #4754

### DIFF
--- a/pkg/service/scheduler/service_integration_test.go
+++ b/pkg/service/scheduler/service_integration_test.go
@@ -7,7 +7,9 @@ package scheduler_test
 import (
 	"context"
 	"encoding/json"
+	"math/rand"
 	"regexp"
+	"slices"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/service/scheduler/service_integration_test.go
```diff
@@ -462,22 +502,124 @@ func TestServiceScheduleIntegration(t *testing.T) {
...
+					// small sleeps in between to verify that there are no races between task
+					// finishing naturally and task being stopped.
+					switch rand.Intn(2) {
+					case 0:
+						if err := h.service.StopTask(ctx, task); err != nil {
+							t.Error(err)
+						}
+						randSleep(1)
+						runner.Cancel(task.ID)
+					default:
```
VAveryanov8: question: Do you think it worth adding a test case with trying to stop and cancel task from different goroutines to simulate some concurrency?

PR #4796, opened 2026-03-18
Remove scylla 2024.1 and 2025.3 support

Michal-Leszczynski:
Even though scylla 2024.1 and 2025.3 are no longer supported, we still test against them in our CI.
The goal of this PR is to remove them from our CI runs, making it faster and less flaky, and to remove any codebase related to the no longer supported scylla 2024.1.

2024.1 is a problematic release, as it is the last one that might be operating on non raft schema / auth / service levels. Because of that, restore procedures for those entities were different for those versions.
Apart from that, there are also other scylla features which are now supported on all supported scylla versions, so this PR also removes their feature checks and assumes that they are always there. 

This PR also adjusts the docs containing a lot of information related to workaround needed for scylla 2024.1.

When removing 2024.1 and 2025.3 from test matrix, this PR also added 2026.2.0-rc0 in their place.


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 this is a needed, but lengthy PR (mostly code removal and version bumps). I tried to divide the changes into sensible commits to make it easier to review, but I can also replace this one PR with several smaller ones if that helps.

karol-kokoszka (approved):
Nice !
Please monitor https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/ after the merge.

Michal-Leszczynski:
Small changes:
- addressed copilot comment about outdated comment about tablet repair api
- bumped tested 2026.2.0-rc0 to 2026.2.0-rc1

Michal-Leszczynski:
Encountered new failures against scylla nightly related to new changes on scylla master.
Added commits addressing them.

PR #4808, opened 2026-04-07
fix: add --amend flag to docker manifest create for idempotent re-runs

yaronkaikov:
## Summary

- Adds `--amend` flag to all `docker_manifests` `create_flags` in `dist/.goreleaser-docker.yaml`
- Fixes docker manifest creation failure when the promote job is re-run for the same version

## Problem

When the promote job runs more than once for the same version (e.g. `3.9.1`), the arch-specific tags like `scylladb/scylla-manager:3.9.1-x86_64` already point to manifest lists from the previous run. `docker manifest create` fails with:

```
docker.io/scylladb/scylla-manager:3.9.1-x86_64@sha256:... is a manifest list
```

because it expects plain single-arch image manifests, not manifest lists.

## Fix

Adding `--amend` to `create_flags` makes `docker manifest create` idempotent — it amends existing manifest lists instead of failing when the source tags are already manifest lists.

Ref: Jenkins build https://jenkins.scylladb.com/view/scylla-manager/job/manager-3.9/job/manager-promote/6/console

### DIFF
--- a/dist/.goreleaser-docker.yaml
+++ b/dist/.goreleaser-docker.yaml
@@ -101,6 +101,7 @@ docker_manifests:
     - "scylladb/scylla-manager:{{ .Version }}-aarch64"
     create_flags:
     - --insecure
+    - --amend
     push_flags:
     - --insecure
     skip_push: false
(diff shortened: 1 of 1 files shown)


Michal-Leszczynski (approved):

Michal-Leszczynski:
@yaronkaikov merged and backported

PR #4814, opened 2026-04-10
Don't list entire bucket when resolving hosts

Michal-Leszczynski:
When establishing which host has access to which location, SM performs simple non-recursive listing of bucket root dir. In the majority of the cases, this is a low-cost way of checking listing permissions. The problem is that the bucket root dir can theoretically contain many other user dirs/objects, as SM controls just the /backup directory. To avoid timeouts when listing bucket root dir with thousands of dirs/objects, we can list the SM controlled /backup dir containing just the 3 sub-dirs (/meta, /schema, /sst). Even if the /backup dir does not exist, we ignore the not found error and treat is as empty listing, as it already proves that we have access to this bucket (otherwise we would get the permission denied error).

### DIFF
--- a/pkg/service/backup/service.go
+++ b/pkg/service/backup/service.go
@@ -610,7 +610,7 @@ func (s *Service) resolveHosts(ctx context.Context, client *scyllaclient.Client,
 		}
 
 		for _, h := range checklist {
-			_, err := client.RcloneListDir(ctx, h, l.RemotePath(""), nil)
+			_, err := client.RcloneListDir(ctx, h, l.RemotePath("backup"), nil)
 			if err != nil {
 				s.logger.Debug(ctx, "Location check FAILED", "host", h, "location", l, "error", err)
 			} else {


Michal-Leszczynski:
@karol-kokoszka this is a simple fix (or rather improvement) that aims to remove SCT failures encountered when testing rc for SM 3.10. In this sense, it's the last piece needed for SM 3.10 release (or rather its testing).

As always during release times, we are lucky, and we encountered new issue (most likely SM side) when testing against scylla-nightly:latest image. It's related to changed index names in DESC SCHEMA output and was caused by https://github.com/scylladb/scylladb/pull/29083 merged to scylla master yesterday.

Mentioned PR is tagged for 2026.2 only (no backports), so it's shouldn't affect SM 3.10 which does not yet officially support scylla 2026.2. Are we fine with merging this PR as is, creating `branch-3.10` from master and proceeding with the release and fixing this for the next releases, or do you want it to be already fixed in SM 3.10 release?

karol-kokoszka (approved):

PR #4819, opened 2026-04-17
Fix CheckPermissions for NFS

Michal-Leszczynski:
4f1adf6f introduced explicit cleanup of test dirs in CheckPermissions. The problem was that the Purge happened while the file descriptor from the "cat" verification was still held. On a regular filesystem this is not an issue, as Purge would still just remove the dir with its contents and make the file descriptor broken. The situation is different with NFS, as when another process tries to remove file with open descriptor, NFS will perform a "silly rename" and create an ephemeral .nfsXXX file so that the descriptor is not broken. So when rclone runs Purge, it first removes all its files, this results in creation of .nfsXXX files and when it proceeds with dir removal, it fails with "directory not empty".

This commit fixes the bug related to holding the file descriptor for too long. It also explicitly checks the permission to remove a file and performs the cleanup without failing the CheckPermissions if just the cleanup failed, as it is not critical.

PR #4821, opened 2026-04-17
Backport: Fix CheckPermissions for NFS (#4819)

Michal-Leszczynski:
4f1adf6f introduced explicit cleanup of test dirs in CheckPermissions. The problem was that the Purge happened while the file descriptor from the "cat" verification was still held. On a regular filesystem this is not an issue, as Purge would still just remove the dir with its contents and make the file descriptor broken. The situation is different with NFS, as when another process tries to remove file with open descriptor, NFS will perform a "silly rename" and create an ephemeral .nfsXXX file so that the descriptor is not broken. So when rclone runs Purge, it first removes all its files, this results in creation of .nfsXXX files and when it proceeds with dir removal, it fails with "directory not empty".

This commit fixes the bug related to holding the file descriptor for too long. It also explicitly checks the permission to remove a file and performs the cleanup without failing the CheckPermissions if just the cleanup failed, as it is not critical.

This is a backport of #4819.

### DIFF
--- a/pkg/rclone/operations/operations.go
+++ b/pkg/rclone/operations/operations.go
@@ -4,6 +4,7 @@ package operations
 
 import (
 	"context"
+	stderr "errors"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


karol-kokoszka (approved):
Is there a way to reproduce the original problem ?

Michal-Leszczynski:
> Is there a way to reproduce the original problem ?

This is reproducible only with an actual NFS setup, so it's not a part of our CI, but it was tested manually ([comment](https://github.com/scylladb/scylla-manager/pull/4819#issuecomment-4267512267)).

PR #4827, opened 2026-05-13
Add fake gcs server in test env

Michal-Leszczynski:
Description from commit messages:

This commit adds fake-gcs-server to test env.
It's a separate docker container that's always started
on test env startup. It is exposed on HTTP, but can be
configured over HTTPS in the future if needed (like minio).
It is added to SM agent config and scylla.yaml config so
that it can be used by rclone and native backups.

This commit sets default backup provider to GCS for two CI
workflows - one with 2025.4 and rclone backup/restore method,
one with latest and native backup/rclone method.

Fixes https://scylladb.atlassian.net/browse/CLOUD-1910


testing/Makefile
```diff
@@ -69,6 +74,7 @@ up:
 	@cp scylla/config/scylla.yaml scylla/scylla.yaml
 	@cp scylla/config/scylla-sm.yaml scylla/scylla-sm.yaml
 	@$(YQ) write -i scylla/scylla.yaml 'object_storage_endpoints[0].(name)' $(MINIO_HOST)
+	@$(YQ) write -i scylla/scylla.yaml 'object_storage_endpoints[1].(name)' $(SCYLLA_GCS_ENDPOINT)
```
karol-kokoszka, line 77: Is it fine that MINIO_HOST is just an IP, but SCYLLA_GCP_ENDPOINT is the full URL with schema and port ? 

```
MINIO_HOST := 192.168.200.99
SCYLLA_GCS_ENDPOINT := http://192.168.200.97:4443
```

karol-kokoszka (approved):

testing/Makefile
Michal-Leszczynski: yeah, initially, s3 endpoints had a different object storage endpoints syntax, where name was just the IP, while http/s and port was configured with different yaml fields. It was changed in 2026.1 so that name contained the http/s and port as well. So we configure gs endpoints with new syntax, as that's the only one supported for gs, while we use the old syntax for s3, as we also test native backup/restore for versions older than 2026.1.

PR #4829, opened 2026-05-14
Match OSE against given provider

Michal-Leszczynski:
ScyllaObjectStorageEndpoint didn't work correctly when there were multiple object storage endpoints with different types (s3/gcs) configured in scylla.yaml. That's because it was looking for any matching endpoint instead of a matching endpoint for given provider.


karol-kokoszka (approved):
Please merge if the CI is green.

PR #4834, opened 2026-05-21
Handle retention locks on SM side

Michal-Leszczynski:
This PR adds new `StageRetentionLock` to SM backup procedure.
This stage is configured according to `RetentionLockMode`, `OverrideRetentionLock`, `RetentionDays` backup Target fields (not yet exposed via `sctool` flags).

Most things are straight forward and described in commit messages, but here are a few considerations that could be discussed:
- on SM agent rclone side, we do best effort approach to object locking, on SM server side we do fail fast approach - is it a problem?
- no progress resume for `StageRetentionLock` - it always starts from scratch
- no `sctool progress` display for `StageRetentionLock` - progress can be tracked only via metrics
-  `StageRetentionLock` operates for up to 12 nodes in parallel to limit the amount of manifests stored in memory - can be fixed in the future, but is this a problem?
- no clean handling of permission and lock configuration errors - UX is going to be improved in next PRs
- google SDK for GCS (big addition to vendor) added just because of test code - is this a problem?

Fixes https://scylladb.atlassian.net/browse/CLOUD-1915


Michal-Leszczynski:
@copilot+claude-opus-4.6 review this PR

karol-kokoszka (approved):

PR #4836, opened 2026-05-24
Retention lock permission check

Michal-Leszczynski:
This PR extends rclone permission check with the ability
to check retention lock related permissions.
It does so by creating additional file, locking it for
a minute in unlocked mode, and then overriding it to
the locked mode. This approach results in not being
able to remove this file at the end of permission check.
To help with accumulating leftover files, permission check
was refactored to work under per cluster prefix. Before
checking permissions during backup procedure, this prefix
is cleared to remove all leftover files with passed test
retention period.

This is obviously not the best approach, as we should try
to create the file in unlocked mode and remove this file
retention lock to test override. It's not done like that
because of the current rclone fork limitation that only
allows to set retention lock, but not remove it
(locked boolean param). When this is fixed on rclone side,
we can improve the permission check procedure here.
Note that the refactor of the prefix of permission check
test files will still be useful then, as we won't always
have the permission to override the lock.

Fixes https://scylladb.atlassian.net/browse/CLOUD-1968


pkg/scyllaclient/client_ping.go
```diff
@@ -29,7 +29,7 @@ func (c *Client) GetNodesWithLocationAccess(ctx context.Context, nodes NodeStatu
 	err := parallel.Run(len(nodes), parallel.NoLimit, func(i int) error {
 		n := nodes[i]
 
-		err := c.RcloneCheckPermissions(ctx, n.Addr, remotePath)
+		err := c.RcloneCheckPermissions(ctx, n.Addr, remotePath, false, false)
```
karol-kokoszka: **Suggestion:** These bare `false, false` args are hard to parse without jumping to the signature. Consider introducing a `PermissionCheckOpts` struct so call sites read as `PermissionCheckOpts{}` (zero value = no lock check). This also future-proofs against a third boolean (e.g. `checkVersioning`) and eliminates the risk of accidentally swapping the two bools — the compiler can't catch `locked, overrideLock` vs `overrideLock, locked` since both are `bool`.

pkg/service/backup/service.go
```diff
@@ -261,13 +266,47 @@ func (s *Service) checkLocationsAvailableFromNodes(ctx context.Context, client *
...
+	// When checking permissions related to retention lock, it's expected
+	// that some files might be left locked for a short retention period.
+	// We need to clean them up on the next permission check.
+	// Since the same bucket is used by multiple SM agents, the cleanup
+	// needs to be performed separately, so that it does not interfere
+	// with some agents still ongoing permission checks.
+	// Since multiple clusters can be backed up to the same bucket,
+	// only the files related to current cluster permission checks
+	// should be cleaned up.
+	permissionCheckPath := path.Join(".permission-check", "cluster", clusterID.String())
```
karol-kokoszka, line 280: **Nit:** Consider expanding this comment to explain *why* cleanup runs unconditionally — not just what it does. A future reader will instinctively think "why not guard this with `if locked`?" and the answer is the lock→unlock transition scenario.

Something like:

```go
// Cleanup runs unconditionally (even when locked=false) because a cluster
// may have previously had retention lock enabled. Files created during prior
// lock checks persist after lock is disabled, and this is the only path
// that removes them. The cost of a 404'd RcloneDeleteDir is negligible.
```

pkg/rclone/operations/operations.go
```diff
@@ -57,9 +59,13 @@ func (e OperationError) StatusCode() int {
...
-// CheckPermissions checks if file system is available for listing, getting,
-// creating, and deleting objects.
-func CheckPermissions(ctx context.Context, l fs.Fs) error {
+// CheckPermissions checks if file system is available for
+// listing, getting, creating, and deleting objects.
+// The remote parameter specifies a subdirectory within l
+// where the test directory is created.
+// Locked and overrideLock controls whether retention lock
+// permissions should be verified.
+func CheckPermissions(ctx context.Context, l fs.Fs, remote string, locked, overrideLock bool) error {
```
karol-kokoszka, line 68: **Suggestion:** The parameter `locked` is ambiguous here. In this function it means "should we check retention lock permissions?" but a few lines below, `RetentionLock(ctx, remote, false, ...)` uses a boolean to mean "governance vs compliance mode" — also colloquially called "unlocked vs locked."

Consider renaming to `checkRetentionLock` to make it clear this is a feature gate, not a mode selector:

```go
func CheckPermissions(ctx context.Context, l fs.Fs, remote string, checkRetentionLock, overrideLock bool) error {
```

This avoids the reader having to disambiguate three meanings of "locked" in one call chain: (1) feature enabled, (2) governance vs compliance mode, (3) the method name `RetentionLock` itself.

(Pairs well with the options struct suggestion on `client_ping.go` — if you go that route, this becomes `opts.CheckRetentionLock` and the ambiguity disappears.)

pkg/service/backup/service.go
```diff
@@ -261,13 +266,47 @@ func (s *Service) checkLocationsAvailableFromNodes(ctx context.Context, client *
...
+		cleanedUpDCs[n.Datacenter] = struct{}{}
+
+		l, ok := dcl[n.Datacenter]
+		if !ok {
+			l = dcl[""]
+		}
+
+		err := client.RcloneDeleteDir(ctx, n.Addr, l.RemotePath(permissionCheckPath))
+		if err != nil && scyllaclient.StatusCodeOf(err) != http.StatusNotFound {
+			s.logger.Info(ctx, "Failed to cleanup permission check dir",
```
karol-kokoszka, line 297: **Suggestion:** A failed cleanup indicates an infrastructure problem (permission denied, network issue, bucket misconfiguration). Logging at Info level makes this invisible to operators who filter on Error/Warning.

Consider upgrading to Error — this is not a normal operational event, it's a failure that causes silent accumulation of orphaned files:

```go
s.logger.Error(ctx, "Failed to cleanup permission check dir",
    "host", n.Addr,
    "location", l,
    "error", err,
)
```

If cleanup keeps failing (e.g., IAM change revoked delete permissions), operators should know before they get a surprise storage bill from thousands of orphaned lock-test files.

karol-kokoszka (approved):
Few comment, but overall +1.

pkg/rclone/operations/operations.go
Michal-Leszczynski: yeah - in general, I will do a refactor of permission check in the near future, when I update rclone fork so that it allows to:
- check current retention lock configuration
- remove retention lock configuration
But I will rename it as suggested for now.

pkg/service/backup/service.go
Michal-Leszczynski: I didn't log it on error level, as it might fail because of locked files.
When backup task is added and scheduled to run right away, SM first generates the backup target (and checks permissions) to validate task correctness, and then generates target as a part of backup procedure. It might happen that the second permission check would still be unable to remove files locked by the first permission check. Moreover, the permission check itself validates all required permissions, so if the cleanup failed because of lack of permissions, the permission check will fail as well.
This could be a good candidate for warning level log, but we don't have/use such log level in SM.

PR #4838, opened 2026-05-27
Expose retention lock feature

Michal-Leszczynski:
This PR contains the last missing pieces needed for the MS1 of WORM/Immutable backup.
It populates retention lock fields used when displaying `sctool backup --dry-run` and `sctool progress backup`.
It adds the `--retention-lock-mode` and `--override-retention-lock` flags.
It documents this feature on both `sctool --help` and backup docs level.

Fixes https://scylladb.atlassian.net/browse/CLOUD-1916


karol-kokoszka (changes requested):
**Design doc needs updating to reflect implementation constraints.**

The [design doc](https://scylladb.atlassian.net/wiki/spaces/RND/pages/325386801/Immutable+WORM+backup) states "we won't allow to decrease `--retention-days`" but doesn't describe:

- How this is enforced (server-side validation on task update)
- The interaction between disabled/deleted tasks and the `retentionMap` used by other tasks' purge stages
- The `defaultRetentionForDeletedTask()` (30-day) fallback and how it interacts with longer GCS locks
- The operational workflow for users who want to stop a retention-locked backup

These are not edge cases — they're the natural "day 2" operations that users will attempt. Please update the design doc alongside the code fixes.

pkg/service/backup/model.go
```diff
@@ -129,14 +129,17 @@ type Run struct {
...
-	Stage       Stage
+	PrevID                uuid.UUID
+	SnapshotTag           string
+	Units                 []Unit
+	DC                    []string
+	Nodes                 []string
+	Location              []backupspec.Location
+	StartTime             time.Time
+	Stage                 Stage
+	RetentionLockMode     RetentionLockMode
```
karol-kokoszka, line 140: **Missing validation: decreasing `retention-days` must be rejected when retention lock is enabled.**

The design doc explicitly states: "we won't allow to decrease `--retention-days`". This is critical because GCS object locks are set at `snapshot_time + retention_days` at the time of backup. If the user later decreases `retention-days`, `staleTags()` will mark snapshots as eligible for purge while GCS locks are still active, causing `StagePurge` to fail with hard errors from the cloud provider.

There is no recovery path for `locked` mode — you cannot shorten or remove GCS locks even as the bucket owner. The validation in `taskProperties` (around line 262) should compare the new `retention-days` against the current task's value and reject any decrease when `retention-lock-mode` is set.

The docs at `retention-lock.rst:113-116` describe these rules (`--retention-days` can only be increased in locked mode, decrease in unlocked requires `--override-retention-lock`) but the server-side enforcement is missing.

pkg/service/backup/service.go
```diff
@@ -695,14 +695,17 @@ func (s *Service) Backup(ctx context.Context, clusterID, taskID, runID uuid.UUID
...
-		Stage:     StageInit,
+		ClusterID:             clusterID,
+		TaskID:                taskID,
+		ID:                    runID,
+		Units:                 target.Units,
+		DC:                    target.DC,
+		Location:              target.Location,
+		StartTime:             timeutc.Now().UTC(),
+		Stage:                 StageInit,
+		RetentionLockMode:     target.RetentionLockMode,
```
karol-kokoszka, line 722: **`TaskDecorator` excludes disabled tasks from the `retentionMap`, which poisons other tasks' purge.**

At line 105 of this file, `TaskDecorator` calls `ListTasks` with the default `ListFilter`, which has `Disabled: false` — meaning it only returns enabled tasks (see `service_list.go:46-48`). When a retention-locked task is disabled, its manifests are still in GCS but no longer represented in the `retentionMap`. Any other active backup task sharing the same location will then:

1. See those manifests during its purge (`listManifestsInAllLocations` lists all manifests for the cluster)
2. Fall back to `defaultRetentionForDeletedTask()` = 30 days
3. Try to delete snapshots older than 30 days
4. Fail because GCS locks (set at the original, longer retention) are still active

This silently breaks unrelated backup tasks. The fix should either pass `ListFilter{Disabled: true}` in `TaskDecorator` (line 105), or the fallback in `GetRetention` should use the lock metadata from the manifest itself rather than a hardcoded 30-day default.

docs/source/backup/retention-lock.rst
```diff
@@ -0,0 +1,126 @@
...
+
+ScyllaDB Manager deduplicates SSTables across backups — multiple snapshots (possibly from different backup tasks)
+can reference the same underlying files. When retention lock is applied to a snapshot, it is also applied
+to all files referenced by that snapshot, including files shared with other snapshots.
+
+.. warning::
+
+   Applying conflicting retention lock configurations (e.g., different modes or retention periods)
+   from different backup tasks to overlapping files may result in errors. Ensure that backup tasks
+   targeting the same backup location use consistent retention lock settings.
```
karol-kokoszka: **Missing operational guidance for retention-locked backups.**

The docs should additionally cover:

- That `retention-days` cannot be decreased once retention lock is enabled in `locked` mode (the rules are listed above at line 113-116 but the *consequences* of violation are not explained — i.e., purge failures that break the backup task)
- What happens if you disable or delete a backup task that has active locks: files remain in GCS, and other tasks sharing the same location may fail during purge
- The safe procedure for stopping a locked backup: disable the task → wait for `last_snapshot_time + retention_days` to pass → only then delete the task (and only if no other active backup tasks share the location)
- Warning that multiple backup tasks sharing a GCS location interact through the shared `retentionMap` and purge stages — disabling/deleting one task can cause purge failures in another

docs/source/backup/retention-lock.rst
```diff
@@ -0,0 +1,126 @@
...
+
+* ``disabled`` (default): No retention lock is applied to snapshot files.
+* ``unlocked``: Retention lock is applied but can be shortened or removed with special permissions (see `Prerequisites`_).
+* ``locked``: Retention lock is applied and cannot be overridden. Once set, the lock cannot be removed or shortened, even by the bucket owner.
+
+Override Lock
+=============
+
+The :ref:`sctool backup --override-retention-lock <sctool-backup>` flag allows overriding previously
+applied retention locks in ``unlocked`` mode. This flag is required in the following cases:
```
karol-kokoszka: **"Required" is not enforced server-side.**

The docs state `--override-retention-lock` is "required" for these transitions, but SM never validates this. The `validate()` function in `model.go:276` only sees the new task properties — it has no access to the previous task state (previous mode, previous retention-days), so it cannot determine whether the override flag should be mandatory.

As a result, if a user forgets `--override-retention-lock` when e.g. changing from `unlocked` → `locked`, the operation proceeds until GCS rejects the lock override attempt, returning a cloud-provider-level error instead of a clear SM validation message.

To enforce this properly, the check would need to live in the update path (e.g., `scheduler.PutTask()` or `service.GetTarget()`) where both old and new properties are accessible. Either:
1. Add validation in the update path that compares old vs. new and rejects if override flag is missing when required, or
2. Reword the docs from "required" to "recommended" and document that GCS will reject the operation if omitted (making the behavior explicit rather than surprising).

vendor/github.com/scylladb/scylla-manager/v3/pkg/managerclient/utils.go
```diff
@@ -261,30 +261,49 @@ func FormatIntensity(v float64) string {
...
 	}
+	res += FormatRetentionLock(retentionLockMode, overrideRetention, retentionDays)
+
 	return res
 }
 
+// FormatRetentionLock returns text representation of retention lock configuration.
+func FormatRetentionLock(mode string, override bool, retentionDays int64) string {
+	if (mode == "" || mode == "disabled") && !override {
+		return fmt.Sprintf("  - Retention lock is not applied to snapshot files\n")
```
karol-kokoszka, line 288: nit: `fmt.Sprintf` with no format verbs here and on line 295. These are plain strings — no `%` placeholders — so `fmt.Sprintf` is unnecessary. Will trigger staticcheck SA1006.

```go
// Before
return fmt.Sprintf("  - Retention lock is not applied to snapshot files\n")
res += fmt.Sprintf("  - Override retention lock configuration on existing snapshot files\n")

// After
return "  - Retention lock is not applied to snapshot files\n"
res += "  - Override retention lock configuration on existing snapshot files\n"
```

pkg/service/backup/service_retention_lock_integration_test.go
```diff
@@ -124,33 +124,64 @@ func TestBackupRetentionLockIntegration(t *testing.T) {
...
 	WriteData(t, clusterSession, testKeyspace, 1)
 
-	Print("When: first backup with retention lock (unlocked, 1 day) is executed")
-	props := defaultTestProperties(location, testKeyspace)
-	props["retention_days"] = 1
-	props["retention_lock_mode"] = "unlocked"
-	if CheckConstraint(t, ni.ScyllaVersion, "< 2026.1") {
-		props["method"] = "rclone"
-	}
+	getTargetAndValidate := func(lockMode string, overrideLock bool, retentionDays int) backup.Target {
```
karol-kokoszka, line 127: **Missing negative/error-path integration tests for the new flags.**

The tests here cover only the happy path (valid inputs → correct behavior). There are no tests verifying that invalid inputs are rejected with clear errors. The validation in `model.go:276-293` has three guards that are untested:

1. `--retention-lock-mode=locked` without `--retention-days` → should error: "retention days must be set"
2. `--retention-lock-mode=locked` with count-based `--retention` → should error: "count-based retention mustn't be set"
3. `--retention-lock-mode=locked` with a non-GCS provider → should error: "retention lock is not supported for ... provider"

Without error-path tests, someone could accidentally break validation logic and no test would catch it until production.

Per `AGENTS.md`: "Integration tests should follow table-driven approach whenever possible, have separate setup and test execution stages and focus on comprehensiveness." A table-driven test exercising each invalid combination would satisfy this.

pkg/service/backup/model.go
Michal-Leszczynski: In general, I should update the design doc.
Initially I thought that we can validate updated retention lock params, but:

1. It's not that easy - let's say that user executed backup task with --retention-days=10 and --retention-lock-mode=locked. Then, even without updating the backup task, they can create another one with --retention-days=5 --retention-lock=unlocked and execute this task. They would end up with an error, as sstable deduplication works across different backup tasks, so the sstables previously locked for 10 days would cause conflicts. To have a full safety nets, we would need to analyze all backup tasks and tables on which they were taken. We would also need to reject creation of backup task with conflicting params. What if the task was deleted? Or a new SM instance was used with already existing backup bucket with already locked files? There is no end to a proper validation.

2. In the next releases, we can improve how SM agent applies the locks:
- override=false - SM agent tries to apply specified lock configuration. If it fails, it check the current lock config and if it has retention days greater than the ones currently specified, we proceed without an error. Basically, when override=false, we ensure minimal retention, not specific retention. This would allow users to safely decrease retention days and keep sane and safe behavior.
- override=true - SM agent sets given retention configuration, if it fails, it returns an error. This would also be used for cases like removing retention locks (in unlocked mode) by running backup with override=true and mode=disabled.

So because of those reasons, for MS1, I decided to just document how the locks are applied, what changes to the configuration are safe and explain the reasons why deduplicated files create those problems.

pkg/service/backup/service.go
Michal-Leszczynski: In MS2, we will update purge to read snapshot retention (by reading manifest retention) and decide which tags are stale based on that. For now, we don't have such a feature. I would say that not including disabled tasks in retention map is a separate bug that should be addressed regardless of the WORM backup feature. I can add this in this or separate PR.

vendor/github.com/scylladb/scylla-manager/v3/pkg/managerclient/utils.go
Michal-Leszczynski: So this is in vendored file - it can be fixed, but requires separate PR for managerclient module. I would fix it later on.

pkg/service/backup/service_retention_lock_integration_test.go
Michal-Leszczynski: I will extend `TestGetTargetErrorIntegration` with those cases.

docs/source/backup/retention-lock.rst
Michal-Leszczynski: Same as for https://github.com/scylladb/scylla-manager/pull/4838#discussion_r3311550834, it's never enough to verify just the transitions - there are multiple cases where things can fail when different conflicting configurations are used.

I can reword it from required to recommended, but it won't change the fact that the task will fail, so required seemed more strict and appropriate here.

docs/source/backup/retention-lock.rst
Michal-Leszczynski: I will add this info to the docs.

pkg/service/backup/service_retention_lock_integration_test.go
Michal-Leszczynski: done

docs/source/backup/retention-lock.rst
Michal-Leszczynski: Reworded from required to recommended (and must to should) and added a line about the lack of server side validation.

docs/source/backup/retention-lock.rst
Michal-Leszczynski: Extended the docs with purge stage considerations.

pkg/service/backup/service.go
Michal-Leszczynski: Added new commit fixing this logic.

karol-kokoszka (approved):
Thanks ! Approved.

PR #4842, opened 2026-05-29
Rename v3.11.0.cql migration file

Michal-Leszczynski:
DB migrations are applied in lexicographical order, so by default "v3.11.0.cql" would go before "v.3.2.0.cql" and cause errors. This is a tmp workaround to ensure that the migration associated with SM 3.11.0 changes goes after.

cc: @mikliapko


karol-kokoszka (approved):

mikliapko:
@Michal-Leszczynski I haven't yet run upgrade scenarios from 3.9 or lower -> to 3.11, but with the latest build it's expected to fail due to the same reason?

PR #4845, opened 2026-05-30
feat(sstable): add `mt` to the set of known SSTable format versions

michoecho:
Prepare for the addition of a new SSTable format version, `mt`, in an upcoming Scylla version.

Refs https://github.com/scylladb/scylladb/pull/30139
Similar to https://github.com/scylladb/scylla-manager/pull/4578

### DIFF
--- a/pkg/sstable/sstable_naming.go
+++ b/pkg/sstable/sstable_naming.go
@@ -16,8 +16,8 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski (approved):
Thanks for coming here to adjust SM to this change!

Single failure is a flake is not related to this PR and is described in https://scylladb.atlassian.net/browse/CLOUD-2162.

PR #4846, opened 2026-06-01
Bump minio version

Michal-Leszczynski:
Permission checks against s3 started behaving flaky recently.
This PR bumps minio version to check if the root cause is on the outdated minio version side.
When rerunning CI for this PR multiple times, this flake was no longer observed, so it looks like it was indeed minio issue.

Bumoing minio version requires a change in how s3 bucket are recreated in test code.
Previously, test code recreated dir associated with test bucket.
Newer minio versions don't play well with such filesystem manipulation, so this PR
changes the approach to use s3 client for recreating the bucket with s3 API.

Refs https://scylladb.atlassian.net/browse/CLOUD-2162


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 please take a look, this PR tackles SM CI flakiness.

VAveryanov8 (approved):
LGTM!

PR #4847, opened 2026-06-01
Run full repair after restore

Michal-Leszczynski:
After both rclone and native restore, repair is used to propagate data from primary replica to secondary replicas, so we always expect it to be a full repair. On the other hand, we want to set incremental mode to full for explicitness and so that we don't run into any edge cases related to restoring data into a table with some previous repair metadata.

Fixes https://scylladb.atlassian.net/browse/CLOUD-2181

### DIFF
--- a/pkg/service/restore/service_restore_integration_test.go
+++ b/pkg/service/restore/service_restore_integration_test.go
@@ -728,6 +728,26 @@ func smokeRestore(t *testing.T, target Target, keyspace string, loadCnt, loadSiz
 		target.SnapshotTag = srcH.simpleBackup(target.Location[0])
 	}
 
+	ni, err := dstH.Client.AnyNodeInfo(t.Context())
+	if err != nil {
+		t.Fatal(err)
+	}
+	supportsIncrementalRepair, err := ni.SupportsIncrementalRepair()
+	if err != nil {
+		t.Fatal(err)
+	}
+	encounteredTabletRepair := atomic.NewInt64(0)
+	dstH.Hrt.SetInterceptor(httpx.RoundTripperFunc(func(req *http.Request) (*http.Response, error) {
+		if strings.HasPrefix(req.URL.Path, "/storage_service/tablets/repair") && req.Method == http.MethodPost {
+			encounteredTabletRepair.Inc()
+			incrementalMode := req.URL.Query().Get("incremental_mode")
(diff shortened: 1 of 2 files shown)


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 please take a look in a spare time!

karol-kokoszka:
Nice, @Michal-Leszczynski we may want to run full-repair after adding new DC as well.
Not sure if it should be handled by SM or Siren.

Michal-Leszczynski:
> Nice, @Michal-Leszczynski we may want to run full-repair after adding new DC as well.
Not sure if it should be handled by SM or Siren.

Things like that should be handled by siren, but I'm not sure if it uses SM for this purpose or perhaps nodetool.
With SM it can be achieved by running `sctool repair --incremental-mode=full`.

karol-kokoszka:
> > Nice, @Michal-Leszczynski we may want to run full-repair after adding new DC as well.
> > Not sure if it should be handled by SM or Siren.
> 
> Things like that should be handled by siren, but I'm not sure if it uses SM for this purpose or perhaps nodetool. With SM it can be achieved by running `sctool repair --incremental-mode=full`.

Would it interfere with the scheduled incremental repair ?

Michal-Leszczynski:
> Would it interfere with the scheduled incremental repair ?

SM scheduler does not allow to SM repair tasks to run at the same time, so one would be retired later on.

karol-kokoszka:
> > Would it interfere with the scheduled incremental repair ?
> 
> SM scheduler does not allow to SM repair tasks to run at the same time, so one would be retired later on.

What can be done to make the full-repair exclusive ? Siren would need to stop any ongoing repairs - suspend the task and run the full-repair, correct ?

Michal-Leszczynski:
But in general, do we have an issue to run repair after adding DC in siren?
This is a part of adding a new DC in scylla [docs](https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/add-dc-to-existing-dc.html), but I'm not sure how adding DCs to existing cluster is handled in siren.
If needed, I can work on such issue and research the topic of how it's managed in cloud.
Repairing after adding DC is not really connected to this PR, as this would need to be managed outside of SM.

karol-kokoszka (approved):

PR #4852, opened 2026-06-08
Make backup GetTarget safe to run in parallel

Michal-Leszczynski:
This PR makes backup permission check procedure
safe for concurrent calls. Previously, I assumed that
only a single backup task can run in parallel for
a given cluster, but the permission check is also
invoked during dry run, task creation, task update.
Because of that, it was possible that permission check
from task update deleted permission check files from
backup task execution or the other way around.
This PR adds and uses the permission_check_files_ttl config
option to clean up only the stale leftover permission
check files and allow GetTarget to be safely executed
in parallel.

Fixes https://scylladb.atlassian.net/browse/CLOUD-2276

### DIFF
--- a/pkg/service/backup/config.go
+++ b/pkg/service/backup/config.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2017 ScyllaDB
+// Copyright (C) 2026 ScyllaDB
 
 package backup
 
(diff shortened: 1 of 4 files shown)


VAveryanov8 (approved):
👍

karol-kokoszka (commented):
Minor nit: otherwise LGTM after rebase onto master.

pkg/service/backup/config.go
```diff
@@ -37,6 +42,9 @@ func (c *Config) Validate() error {
 	if c.AgeMax < 0 {
 		err = multierr.Append(err, errors.New("invalid age_max, must be >= 0"))
 	}
+	if c.PermissionCheckFilesTTL < 0 {
+		err = multierr.Append(err, errors.New("invalid permission_check_ttl, must be >= 0"))
```
karol-kokoszka: Nit: error message says `permission_check_ttl` but the YAML field is `permission_check_files_ttl`. Should match for consistency.

karol-kokoszka (approved):

Issue #4667, opened 2025-12-02
increase integrated Scylla memory to be in sync with recommended ratio of 1 cpu to at least 2 G of memory


tarzanek:
Currently SM integrated Scylla is limited to 500M as per
https://github.com/scylladb/scylla-manager/blob/master/dist/scripts/scyllamgr_setup#L123
(and 1 cpu as per smp few lines above)

Scylla best practice is to have at least 2G of memory per shard.

Above script should be adjusted to try to get at least 2G (or at least respective fraction of available memory) and definitely not hardcode it.

Motivation is simple - for bigger cluster with > 15 nodes which contain more than 400 tables (and MVs) , then their respective
repair state/progress tables might have partitions even bigger than 100M ( so thinking about remodeling data model of SM repair tables is another topic), which will then delay operations such as repair.
We've seen a delay caused by above to 120h for a repair, that normally (with local Scylla not bottlenecked by lack of memory) takes around 14h.

Fix for this issue should also include upgrade path to readjust existing SMs so they will be improved.

@karol-kokoszka @Michal-Leszczynski

PR #4853, opened 2026-06-09
Update SM DB setup

Michal-Leszczynski:
This PR changes 2 thins in the initial SM DB configuration:
- memory increased from 500M to 1G
- auto_snapshot is disabled

The auto_snapshot property is just not needed, as we don't store important user data and SM is the only automated user of of this DB.
In terms of increasing the memory, the original [issue](https://github.com/scylladb/scylla-manager/issues/4667) recommends to increase it to the minimal recommended production value of 2G, but I decided to make a smaller step first and bump it to 1G. That's because SM DB is not really supposed to store huge amounts of data.
Issue mentions the case of repairing a huge cluster, which indeed is a problem, but that should be fixed by optimizing SM DB usage during vnode repair, as the problem lays in the poorly written code rather than memory limitation on SM DB side. Increasing memory to 2G could not even be enough to fix it, although it would probably help.
On the other hand, it could hurt other, smaller setups, where this additional 1G of memory would be taken away from SM.
In general, the SM DB setup script is just a quick minimal way to configure SM DB, but if a different SM DB config is needed, it can be done either manually or by changing the setup script and running it again.
So for now the memory will be increased to 1G, we will observe how that impacts cloud deployments and increase it further if needed.

Tested it by downloading custom SM build with this changes, deploying it in a docker container, and querying scylla config/logs:
```
[root@17761274a3c6 /]# curl http:/localhost:10000/v2/config/auto_snapshot
false
...
[root@17761274a3c6 /]# journalctl -u scylla-server -b --no-pager | grep "parsed command line options"
Jun 09 08:11:22 17761274a3c6 scylla[506]: parsed command line options: [memory, (positional) 1G, log-to-syslog, (positional) 0, log-to-stdout, (positional) 1, default-log-level, (positional) info, network-stack, (positional) posix, auto-snapshot: false, developer-mode: 1, smp, (positional) 1]
Jun 09 08:18:57 17761274a3c6 scylla[931]: parsed command line options: [memory, (positional) 1G, log-to-syslog, (positional) 0, log-to-stdout, (positional) 1, default-log-level, (positional) info, network-stack, (positional) posix, auto-snapshot: false, developer-mode: 1, smp, (positional) 1]
```

Keeping the initial issue open https://github.com/scylladb/scylla-manager/issues/4667
Fixes https://scylladb.atlassian.net/browse/CLOUD-2227
Fixes https://scylladb.atlassian.net/browse/CLOUD-2228

### DIFF
--- a/dist/scripts/scyllamgr_setup
+++ b/dist/scripts/scyllamgr_setup
@@ -120,7 +120,7 @@ if command -v scylla_setup > /dev/null && [[ ${SETUP_SCYLLA} == 1 ]]; then
         f=/etc/sysconfig/scylla-server
     fi
     if [[ "$f" != "" ]]; then
-      sed -i -e 's/^SCYLLA_ARGS=.*/SCYLLA_ARGS="--memory 1G --log-to-syslog 0 --log-to-stdout 1 --default-log-level info --network-stack posix"/g' ${f}
+      sed -i -e 's/^SCYLLA_ARGS=.*/SCYLLA_ARGS="--memory 1G --log-to-syslog 0 --log-to-stdout 1 --default-log-level info --network-stack posix --auto-snapshot=false"/g' ${f}
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


Michal-Leszczynski:
We can also think about setting the memory limit as something like `max(1G, X% of available memory)`, but not sure about X value (5/10/20?).

karol-kokoszka (approved):

PR #4857, opened 2026-06-10
ci: pin GitHub Actions to commit SHAs

roydahan:
## Summary
- Pin all external GitHub Actions to full commit SHAs to reduce supply chain attack surface
- Upgrade outdated actions to their latest versions

**This PR was generated automatically. Please verify that GitHub Actions work as expected with these changes before merging.**

Reference: https://github.com/scylladb/scylladb/pull/29421

### DIFF
--- a/.github/workflows/docs-pages.yaml
+++ b/.github/workflows/docs-pages.yaml
@@ -18,17 +18,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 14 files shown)


Michal-Leszczynski:
Thanks for making this PR! We generate our CI workflows from config files, so they need to be updated as well. I took care of it in https://github.com/scylladb/scylla-manager/pull/4858 which supersedes this PR.

PR #4858, opened 2026-06-11
Pin gh actions CI to commit SHAs

Michal-Leszczynski:
This PR supersedes #4857 - it makes the same changes, but aligned with how SM CI workflows are generated.


VAveryanov8 (approved):
👍

PR #4862, opened 2026-06-12
Fix repair worker pool cleanup

Michal-Leszczynski:
This PR targets repair test flakiness related to leaking repair jobs across different subtests.
The two main changes are:
- fixing repair worker pool cleanup
- when killing all repairs, abort cluster wide tablet repair tasks first

I rerun CI for this PR multiple times and no more flakes were observed.

This PR additionally performs slight refactor of code related to listing and killing active repairs.
Since all currently supported scylla versions support task manager (scylla svc), we no longer
need to list and kill active repairs with the legacy endpoints. Instead, we can simplify the 
implementation and use just the task manager API, which supports both regular node level
repairs (both vnode and tablet) and the cluster level tablet repairs. 

Fixes https://scylladb.atlassian.net/browse/CLOUD-1969

### DIFF
--- a/pkg/scyllaclient/client_scylla.go
+++ b/pkg/scyllaclient/client_scylla.go
@@ -784,6 +784,11 @@ func (c *Client) ActiveRepairs(ctx context.Context, hosts []string) ([]*models.T
 	return c.activeScyllaTasks(ctx, ScyllaTaskModuleRepair, ScyllaTaskTypeRepair, hosts)
 }
 
+// ActiveTabletRepairs returns scheduled or running ScyllaTaskTypeUserRepair tasks.
+func (c *Client) ActiveTabletRepairs(ctx context.Context) ([]*models.TaskStats, error) {
+	return c.activeScyllaTasks(ctx, ScyllaTaskModuleTablets, ScyllaTaskTypeUserRepair, nil)
+}
+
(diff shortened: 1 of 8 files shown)


Michal-Leszczynski:
@karol-kokoszka @VAveryanov8 please take a look, this PR tackles SM CI flakiness.

pkg/scyllaclient/client_scylla.go
```diff
@@ -777,88 +779,95 @@ func (c *Client) longPollingTimeout(waitSeconds int) time.Duration {
...
 
-	notify := func(i int, err error) {
-		host := hosts[i]
-		c.logger.Error(ctx, "Failed to terminate repair",
-			"host", host,
-			"error", err,
-		)
+			for _, task := range tasks {
+				if err := c.ScyllaAbortTask(egCtx, host, task.TaskID); err != nil {
+					c.client.logger.Error(egCtx, "%s: failed to abort task %s", host, task.TaskID)
```
karol-kokoszka: You miss fmt.Sprintf

```
// Error logs a message with some additional context.
func (l Logger) Error(ctx context.Context, msg string, keyvals ...interface{}) {
	l.log(ctx, zapcore.ErrorLevel, msg, keyvals)
}
```

or use 
```
if err := c.ScyllaAbortTask(egCtx, host, task.TaskID); err != nil {
    c.logger.Error(egCtx, "Failed to abort task", "host", host, "task_id", task.TaskID, "error", err)
}
```

karol-kokoszka (approved):
👍 , but please fix the logging

PR #4864, opened 2026-06-19
docs: update dependencies

annastuchlik:
Update docs Python dependencies (uv.lock) by running \make update\.

### DIFF
--- a/docs/uv.lock
+++ b/docs/uv.lock
@@ -22,15 +22,15 @@ wheels = [
 
 [[package]]
 name = "anyio"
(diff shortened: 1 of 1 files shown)


annastuchlik:
@Michal-Leszczynski Please have a look and merge. This PR addresses the alerts reported by the dependabot: https://github.com/scylladb/scylla-manager/security/dependabot

Michal-Leszczynski (approved):
Thanks! I will remember that we have `make update` that should be executed from time to time.

PR #4865, opened 2026-06-19
fix(repair): fix per-table duration inflation after run resume

VAveryanov8:
[refactor(repair): repair run progress duration](https://github.com/scylladb/scylla-manager/pull/4865/changes/6bbe2d87440a6b023fda97a8f8f920b2ec84201d)
This refactors repair run progress duration. Previously, duration was
calculated differently on different aggregation levels and some of them
were including duration inherited from previous runs and some were not.

This PR introduces following rules:
- Progress duration is per task run (not per task execution, so no need
  to inherit duration from previous task runs)
- Duration should be a wall-clock time (not a job execution duration)

Relates: CLOUD-2473

---

[fix(repair): skip rows with no work from progress merging](https://github.com/scylladb/scylla-manager/pull/4865/changes/e83688ad71551ebca18abd9cf508fce1bf4b5158)
Relates: CLOUD-2473

---

[refactor(managerclient): repair duration rendering](https://github.com/scylladb/scylla-manager/pull/4865/changes/bc2f35fefe4142a84043078c2d90a79aa2035b73)
This changes how repair run progress duration is rendered:
As repair run progress is per task run now (previously it was per task
execution in some cases), it's possible that some progress entries will
have progress 100%, but zero duration, meaning that they were fully
repaired in the previous task run and current task run spent 0s on them.
To make it more clear, in such cases duration will rendered as "done".

Relates: CLOUD-2473

Examples: 
```bash 

# Repair was stopped and resumed:

❯ ./sctool.dev info -c my-cluster repair/82701d59-da9f-4792-8394-76d73af3b59a
Name:   repair/82701d59-da9f-4792-8394-76d73af3b59a
Cron:   {"spec":"","start_date":"0001-01-01T00:00:00Z"} (no activations scheduled)
Tz:     Europe/Warsaw
Retry:  3 (initial backoff 10m)
╭──────────────────────────────────────┬─────────────────────────┬──────────┬─────────╮
│ ID                                   │ Start time              │ Duration │ Status  │
├──────────────────────────────────────┼─────────────────────────┼──────────┼─────────┤
│ 319aaab3-7ed2-11f1-ab6e-124ac6f7831c │ 13 Jul 26 17:47:46 CEST │ 7s       │ DONE    │
│ 156d8636-7ed2-11f1-ab6d-124ac6f7831c │ 13 Jul 26 17:46:59 CEST │ 7s       │ STOPPED │
╰──────────────────────────────────────┴─────────────────────────┴──────────┴─────────╯

# Then final repair run progress has entries inherited from initial run, but they doesn't affect duration:

❯ ./sctool.dev progress -c my-cluster repair/82701d59-da9f-4792-8394-76d73af3b59a
Run:            319aaab3-7ed2-11f1-ab6e-124ac6f7831c
Status:         DONE
Start time:     13 Jul 26 17:47:46 CEST
End time:       13 Jul 26 17:47:54 CEST
Duration:       7s
Progress:       100%
Intensity:      1
Parallel:       0
Datacenters:
  - dc1
  - dc2

╭───────────────────────────────┬────────────────────────────────┬──────────┬──────────╮
│ Keyspace                      │                          Table │ Progress │ Duration │
├───────────────────────────────┼────────────────────────────────┼──────────┼──────────┤
│ audit                         │                      audit_log │ 100%     │ done     │
├───────────────────────────────┼────────────────────────────────┼──────────┼──────────┤
│ system_distributed_everywhere │ cdc_generation_descriptions_v2 │ 100%     │ done     │
├───────────────────────────────┼────────────────────────────────┼──────────┼──────────┤
│ system_distributed            │      cdc_generation_timestamps │ 100%     │ done     │
│ system_distributed            │    cdc_streams_descriptions_v2 │ 100%     │ done     │
│ system_distributed            │                 service_levels │ 100%     │ done     │
│ system_distributed            │              view_build_status │ 100%     │ done     │
├───────────────────────────────┼────────────────────────────────┼──────────┼──────────┤
│ system_replicated_keys        │                 encrypted_keys │ 100%     │ done     │
├───────────────────────────────┼────────────────────────────────┼──────────┼──────────┤
│ test_tablet_ks                │                           tab1 │ 100%     │ 7s       │
│ test_tablet_ks                │                           tab2 │ 100%     │ done     │
├───────────────────────────────┼────────────────────────────────┼──────────┼──────────┤
│ test_vnode_ks                 │                           tab1 │ 100%     │ done     │
│ test_vnode_ks                 │                           tab2 │ 100%     │ done     │
╰───────────────────────────────┴────────────────────────────────┴──────────┴──────────╯

# Initial repair run progress:
❯ ./sctool.dev progress -c my-cluster repair/82701d59-da9f-4792-8394-76d73af3b59a --run 156d8636-7ed2-11f1-ab6d-124ac6f7831c
Run:            156d8636-7ed2-11f1-ab6d-124ac6f7831c
Status:         STOPPED
Cause:          scheduler and all tasks were stopped
Start time:     13 Jul 26 17:46:59 CEST
End time:       13 Jul 26 17:47:07 CEST
Duration:       7s
Progress:       51%
Intensity:      1
Parallel:       0
Datacenters:
  - dc1
  - dc2

╭───────────────────────────────┬────────────────────────────────┬──────────┬───────────╮
│ Keyspace                      │                          Table │ Progress │ Duration  │
├───────────────────────────────┼────────────────────────────────┼──────────┼───────────┤
│ audit                         │                      audit_log │ 100%     │ 17h17m57s │
├───────────────────────────────┼────────────────────────────────┼──────────┼───────────┤
│ system_distributed_everywhere │ cdc_generation_descriptions_v2 │ 100%     │ 1s        │
├───────────────────────────────┼────────────────────────────────┼──────────┼───────────┤
│ system_distributed            │      cdc_generation_timestamps │ 100%     │ 1s        │
│ system_distributed            │    cdc_streams_descriptions_v2 │ 100%     │ 1s        │
│ system_distributed            │                 service_levels │ 100%     │ 1s        │
│ system_distributed            │              view_build_status │ 100%     │ 1s        │
├───────────────────────────────┼────────────────────────────────┼──────────┼───────────┤
│ system_replicated_keys        │                 encrypted_keys │ 100%     │ 1s        │
├───────────────────────────────┼────────────────────────────────┼──────────┼───────────┤
│ test_tablet_ks                │                           tab1 │ 0%       │ 0s        │
│ test_tablet_ks                │                           tab2 │ 100%     │ 7s        │
├───────────────────────────────┼────────────────────────────────┼──────────┼───────────┤
│ test_vnode_ks                 │                           tab1 │ 100%     │ 1s        │
│ test_vnode_ks                 │                           tab2 │ 100%     │ 1s        │
╰───────────────────────────────┴────────────────────────────────┴──────────┴───────────╯

```

**Note**:  `audit_log` duration was growing non stop, so I guess I found another bug, fix is there - [fix(repair): skip rows with no work from progress merging](https://github.com/scylladb/scylla-manager/pull/4865/changes/e83688ad71551ebca18abd9cf508fce1bf4b5158)

---

[chore(swagger/sm): update repair progress duration_ms field description](https://github.com/scylladb/scylla-manager/pull/4865/changes/effb877480d4673cf1ed8fc648dbc37c76525279)
Relates: CLOUD-2473

---

Please make sure that:

### DIFF
--- a/pkg/service/repair/progress.go
+++ b/pkg/service/repair/progress.go
@@ -181,7 +181,10 @@ func (pm *dbProgressManager) initProgress(plan *plan) error {
 			}
 			cp := *rp
 			cp.RunID = pm.run.ID
+			cp.StartedAt = nil         // Reset so duration reflects current run only
+			cp.CompletedAt = nil       // Will be set when completed in this run
 			cp.DurationStartedAt = nil // Reset duration counter from previous run
+			cp.Duration = 0            // Don't inherit duration from previous runs
 			cp.Error = 0               // Failed ranges will be retried
 			pm.progress[pk] = &cp
 		})
@@ -279,12 +282,8 @@ func (pm *dbProgressManager) OnJobStart(ctx context.Context, j job) {
 		pm.mu.Lock()
 		rp := pm.progress[pk]
 
-		rp.runningJobCount++
 		if rp.StartedAt == nil {
 			rp.StartedAt = &start
-		}
-		if rp.DurationStartedAt == nil {
-			rp.DurationStartedAt = &start
 		}
 
 		q.BindStruct(rp)
@@ -319,11 +318,6 @@ func (pm *dbProgressManager) onJobEndProgress(ctx context.Context, result jobRes
 			rp.Success += int64(len(result.ranges))
 		} else {
 			rp.Error += int64(len(result.ranges))
-		}
-
-		rp.runningJobCount--
-		if rp.runningJobCount == 0 {
-			rp.AddDuration(end)
 		}
 
 		if rp.Completed() {
@@ -438,7 +432,7 @@ func (pm *dbProgressManager) AggregateProgress() (Progress, error) {
 			}
 		}
 		host.Tables = append(host.Tables, tp)
-		host.progress = mergeProgress(host.progress, prog)
+		host.progress = mergeProgress(host.progress, prog, now)
 		perHost[rp.Host] = host
 
 		tk := tableKey{keyspace: rp.Keyspace, table: rp.Table}
@@ -452,7 +446,7 @@ func (pm *dbProgressManager) AggregateProgress() (Progress, error) {
 				Table:    rp.Table,
 			}
 		}
-		tab.progress = mergeProgress(tab.progress, prog)
+		tab.progress = mergeProgress(tab.progress, prog, now)
 		perTable[tk] = tab
 	})
 	if err != nil {
@@ -473,7 +467,7 @@ func (pm *dbProgressManager) AggregateProgress() (Progress, error) {
 			return t1.Keyspace+t1.Table < t2.Keyspace+t2.Table
 		})
 
-		p.progress = mergeProgress(p.progress, hp.progress)
+		p.progress = mergeProgress(p.progress, hp.progress, now)
 		p.Hosts = append(p.Hosts, hp)
 	}
 	sort.Slice(p.Hosts, func(i, j int) bool {
@@ -481,7 +475,6 @@ func (pm *dbProgressManager) AggregateProgress() (Progress, error) {
 	})
 
 	for _, tp := range perTable {
-		tp.Duration = recalculateDuration(tp.progress, now)
 		p.Tables = append(p.Tables, tp)
 	}
 	sort.Slice(p.Tables, func(i, j int) bool {
@@ -493,7 +486,6 @@ func (pm *dbProgressManager) AggregateProgress() (Progress, error) {
 	if p.CompletedAt == &ancient {
 		p.CompletedAt = nil
 	}
-	p.Duration = recalculateDuration(p.progress, now)
 
 	if p.TokenRanges == 0 {
 		p.SuccessPercentage = -1
@@ -550,25 +542,27 @@ func recalculateDuration(pr progress, now time.Time) int64 {
 }
 
 func extractProgress(rp *RunProgress, now time.Time) progress {
-	return progress{
+	pr := progress{
 		Success:     rp.Success,
 		Error:       rp.Error,
 		TokenRanges: rp.TokenRanges,
 		StartedAt:   rp.StartedAt,
 		CompletedAt: rp.CompletedAt,
-		Duration:    rp.CurrentDuration(now).Milliseconds(),
 	}
+	pr.Duration = recalculateDuration(pr, now)
+	return pr
 }
 
-func mergeProgress(pr1, pr2 progress) progress {
-	return progress{
+func mergeProgress(pr1, pr2 progress, now time.Time) progress {
+	pr := progress{
 		TokenRanges: pr1.TokenRanges + pr2.TokenRanges,
 		Success:     pr1.Success + pr2.Success,
 		Error:       pr1.Error + pr2.Error,
 		StartedAt:   chooseStartedAt(pr1.StartedAt, pr2.StartedAt),
 		CompletedAt: chooseCompletedAt(pr1.CompletedAt, pr2.CompletedAt),
-		Duration:    pr1.Duration + pr2.Duration,
 	}
+	pr.Duration = recalculateDuration(pr, now)
+	return pr
 }
 
 // chooseStartedAt returns earlier time or nil if each is unset.
--- a/pkg/service/repair/progress.go
+++ b/pkg/service/repair/progress.go
@@ -157,6 +157,10 @@ func (pm *dbProgressManager) initProgress(plan *plan) error {
 		for _, tp := range kp.Tables {
 			for _, h := range plan.Hosts {
 				pk := newHostKsTable(h, kp.Keyspace, tp.Table)
+				// Skip progress entries for tables with no ranges to repair.
+				if _, ok := plan.Stats[pk]; !ok {
+					continue
+				}
 				pm.progress[newHostKsTable(h, kp.Keyspace, tp.Table)] = &RunProgress{
 					ClusterID:   pm.run.ClusterID,
 					TaskID:      pm.run.TaskID,
@@ -177,6 +181,10 @@ func (pm *dbProgressManager) initProgress(plan *plan) error {
 		err := pm.ForEachPrevRunProgress(func(rp *RunProgress) {
 			pk := newHostKsTable(rp.Host, rp.Keyspace, rp.Table)
 			if _, ok := pm.progress[pk]; !ok {
+				return
+			}
+			// Skip progress entries for tables with no ranges to repair.
+			if rp.TokenRanges == 0 {
 				return
 			}
 			cp := *rp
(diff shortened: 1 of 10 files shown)


karol-kokoszka:
@VAveryanov8 as per slack.
Please remove that part which caps max progress to 100%. We don't need it. 
The intention is to have separate repair task for vNodes and for tablets.

Current discrepancy comes from the fact that the customer was using single repair task to cover both vnodes and tablets.

VAveryanov8:
@Michal-Leszczynski @karol-kokoszka  Could you please take a look at this PR - it's quite small. Maybe it's worth including into 3.11.2?

pkg/service/repair/progress.go
```diff
@@ -181,8 +181,16 @@ func (pm *dbProgressManager) initProgress(plan *plan) error {
...
 			cp.Error = 0               // Failed ranges will be retried
+			cp.DurationStartedAt = nil // Reset duration counter from previous run
+			// AggregateProgress calculates duration from min StartedAt and max CompletedAt,
+			// so don't carry previous run timestamps into resumed run progress.
+			cp.StartedAt = &pm.run.StartTime
+			if cp.Completed() {
+				cp.CompletedAt = &pm.run.StartTime
+			} else {
+				cp.CompletedAt = nil
+			}
```
Michal-Leszczynski: I don't really understand this change, but I also know that the repair progress is a mess in general (I wrote it a long time ago myself:)).

This change sets run progress (per table/host) start time to run (per task execution) start time. So even if it wasn't set in the previous run, as the table hasn't yet been repaired back then, it will be set on resume to a value that does not reflect table actual start time, but the task execution start time. Does it make sense?

Michal-Leszczynski: I would say that some part of the mess is a result of mixing statistics for the entire task execution chain (multiple executions reusing progress from one another), and stats for a single task execution. When we look at this part of repair progress:
```
Start time:	19 May 26 14:45:35 UTC
End time:	29 May 26 18:12:11 UTC
Duration:	243h26m35s
``` 
We see start/end time for given task execution/run, but have some aggregated duration across multiple runs and it looks strange. 

Another part of the mess is that we store repair job stats per table/host. This works well for printing per host progress, but complicates stuff for per table progress, as then we can't simply add duration, as it would be multiplied roughly by the RF (replica set size).

Finally, we the SM DB data modelling is also problematic. For the purposes of resuming repair progress, we store all token ranges for given table in a single cell. Each token range is two int64. Each node brings 256 token ranges to the cluster by default. So this single cell has a size of node_cnt * 2 * 8B * 256 = node_cnt * 4KB. For a huge clusters, like the one from this issue, where we have ~100 nodes, this gives us ~400KB cells. By default, large cell warning is emitted on 1MB cells ([refs](https://opensource.docs.scylladb.com/stable/troubleshooting/large-rows-large-cells-tables#find-large-cells)), but SM DB is really small and resource depleted, so the ~400KB being around half of the cell size emitting warning might be really problematic for repair speed. We had seen in the past that the SM DB speed can be a bottleneck in overall SM repair task speed.

This comment is a general description of the poor repair progress state - I'm wondering if it would be easier to rewrite it with more common sense than trying to patch it up, but a rewrite would take more time, could run into different problems, and we might not want to do that as the tablet repair does not have such problems and this is the way to go in the future.

VAveryanov8: > This change sets run progress (per table/host) start time to run (per task execution) start time. So even if it wasn't set in the previous run, as the table hasn't yet been repaired back then, it will be set on resume to a value that does not reflect table actual start time, but the task execution start time. Does it make sense?

Ok, I've just missed part that repair_run_progress contains tables for which repair hasn't yet started :) 

> We see start/end time for given task execution/run, but have some aggregated duration across multiple runs and it looks strange.

In general my idea is to just show progress per current run (ignoring previous runs that were stopped). Do you think it possible/reasonable?

> Another part of the mess is that we store repair job stats per table/host. This works well for printing per host progress, but complicates stuff for per table progress, as then we can't simply add duration, as it would be multiplied roughly by the RF (replica set size).

I think I saw something for that in code were we just taking min(StartedAt) and max(CompletedAt) for the table duration. 

> This comment is a general description of the poor repair progress state - I'm wondering if it would be easier to rewrite it with more common sense than trying to patch it up, but a rewrite would take more time, could run into different problems, and we might not want to do that as the tablet repair does not have such problems and this is the way to go in the future.

Well, I think we can do both - patch what we can without making it more complex and prepare some plan for refactoring (with simplicity in mind).

VAveryanov8:
@Michal-Leszczynski I've updated PR as we agreed: repair progress duration is per run only and doesn't inherit previous runs duration (after pause/resume). Please review it once you have time! Thanks! 

P.S. There are some integration test failures, but I believe they are not related to this PR, but  most likely to some recent changes in the latest scylla.

Michal-Leszczynski:
> P.S. There are some integration test failures, but I believe they are not related to this PR, but most likely to some recent changes in the latest scylla.

Created an issue for that: https://scylladb.atlassian.net/browse/CLOUD-2993

pkg/service/repair/progress.go
```diff
@@ -426,6 +420,13 @@ func (pm *dbProgressManager) AggregateProgress() (Progress, error) {
 			Table:    rp.Table,
 		}
 
+		// Skip rows with no work from progress merging.
+		// Hosts with TokenRanges == 0 have nil StartedAt/CompletedAt,
+		// which would incorrectly make the merged CompletedAt nil
+		// (interpreted as "still in progress"), causing ever-growing
+		// duration for stopped/completed runs.
+		noWork := rp.TokenRanges == 0
```
Michal-Leszczynski: Good catch! But it seems to me that it doesn't fix the root cause.
IIUC, DB record with zero TokenRanges shouldn't be there in the first place, as we would always want to exclude them from any progress display and they could result in other, yet unknown progress display errors.

Such records can be either initialized in the current run, or inherited from the previous one.
Could we:
- not inherit 0 token ranges records from the previous run
- in `initProgress`, init only the records from `plan.Stats`

What do you think about that?

In case of this place in the code, we can keep this check, but perhaps do it only a single time:
```go
if rp.TokenRanges == 0 {
  return
}
```
at the beginning of the callback.

pkg/service/repair/progress.go
```diff
@@ -567,7 +574,6 @@ func mergeProgress(pr1, pr2 progress) progress {
 		Error:       pr1.Error + pr2.Error,
 		StartedAt:   chooseStartedAt(pr1.StartedAt, pr2.StartedAt),
 		CompletedAt: chooseCompletedAt(pr1.CompletedAt, pr2.CompletedAt),
-		Duration:    pr1.Duration + pr2.Duration,
```
Michal-Leszczynski, line 570: nit: I know it's a spaghetti code, but currently:
- `extractProgress` includes duration
- `mergeProgress` does not include duration
- we still need to call `recalculateDuration` on everything

Could we add `recalculateDuration` as a part of `mergeProgress` and not call it directly in `AggregateProgress`, so that it's easy to see that it's always evaluated and not skipped?

Michal-Leszczynski (commented):
Looks good! Just 2 comments.

pkg/service/repair/progress.go
VAveryanov8: Yep, make sense, updated in 434b38a7cdd2e7dcaa49f538df01b8c826d32168

pkg/service/repair/progress.go
VAveryanov8: Fixed in 2fa52c7699ab0e27be4f68b00df0767aec71765b (I'll squash it with initial fix commit before the merge)

Michal-Leszczynski (approved):

PR #4879, opened 2026-07-15
chore: managerclient and swagger deps bump

VAveryanov8:
Follow up to https://github.com/scylladb/scylla-manager/pull/4865 with v3/ modules update

---

Please make sure that:


Michal-Leszczynski (commented):
For completeness, when swagger related to managerclient is bumped, the swagger dep in managerclient module go.mod should also be bumped. It's not that important here, as the change is only about param description, but it would be nice to keep those in sync.

Michal-Leszczynski (approved):
Thanks!
One last nit would be scope in commit messages.
First commit: `chore(v3/managerclient): update to the latest version` sounds like making changes in the `v3/managerclient` module, while it actually does changes in the main module by bumping the managerclient version used there. I would use something like `chore(go.mod): update managerclient dep` for this commit instead.
Last commit: `chore(v3/managerclient): bump swagger dependency` has the same scope, but it indeed makes changes in the managerclient module, so here it sounds right.
Does this make sense?

Michal-Leszczynski:
@VAveryanov8 could we merge this PR? I'm going to be bumping swagger dep in my next PR, so it would be nice to get this in before this happens.

PR #4880, opened 2026-07-16
fix(restore): remove flaky partial repair progress assertion

karol-kokoszka:
The assertion 'Expected partial repair progress' is inherently racy with tablet keyspaces because tablet repair is atomic per table — if interrupted, it restarts from scratch on resume. Since the test repairs only one table (big_table), there is no reliable intermediate state to observe. The test still validates that pause/resume during repair works correctly by completing Run 3 successfully.

For tablet keyspaces, a single table's repair is atomic — if interrupted, it restarts from scratch on resume (see generator.go: "Always repair unfinished tablet table from scratch as tablet load balancing is enabled when repair is interrupted"). Since this test repairs only one table (big_table), there is no meaningful partial progress to observe between "no tables done" and "all tables done".

Fixes: https://scylladb.atlassian.net/browse/CLOUD-1970

### DIFF
--- a/pkg/service/restore/service_restore_integration_test.go
+++ b/pkg/service/restore/service_restore_integration_test.go
@@ -970,14 +970,6 @@ func restoreWithResume(t *testing.T, target Target, keyspace string, loadCnt, lo
 		t.Fatalf("Expected context error but got: %+v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


Michal-Leszczynski (approved):
So the more conservative change would be to add new table and expect that restore paused in between those table repairs (with some interceptor logic) would have some non empty repair progress. But I'm not sure if this test case is that important - it's kind of testing repair functionality in restore environment. It's also not that interesting from tablet repair POV.

PR #4882, opened 2026-07-17
Bump rclone fork to v1.54.0-patched-2

Michal-Leszczynski:
New rclone fork version introduces new interfaces needed for handling retention locks and event based holds for CLOUD-2890. It also changes retention lock interfaces in a breaking way. Because of that, their usage needs to be adjusted in the same commit which introduces new rclone version.

Here are the new commits included in the rclone fork bump: https://github.com/scylladb/rclone/compare/073b5e53...ccf542aa.

Refs https://scylladb.atlassian.net/browse/CLOUD-2890

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -167,6 +167,6 @@ require (
 
 replace (
 	github.com/gocql/gocql => github.com/scylladb/gocql v1.18.3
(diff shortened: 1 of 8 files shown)


VAveryanov8 (approved):
As far as I can tell changes looks good!

PR #4884, opened 2026-07-20
Try harder when parsing sstable UUID

Michal-Leszczynski:
SM relies on knowing sstable format to extract sstable ID from
sstable file name. In general that's correct, but it adds friction
when scylla adds new, similar sstable format. It requires immediate
adjustments on SM side and can block scylla side development.
Most new sstable formats have just different 2 letter prefix, but
they keep sstable UUID in the same place.

We can leverage that, so that when SM encounters sstable with unknown format,
it would still try to get its UUID in the same way as for the known formats.

Since we are dealing with unknown format, we need to add validation
to ensure that extracted substring matches scylla UUID specs.
To do that, we promote a part of the test code responsible for
parsing sstable UUID and retrieving its timestamp. This code was written
by scylla dev and works fine with currently used sstable file names.

As a final refactor, this commit also removes a few no longer used functions
related to sstable renaming (they are no longer used since we don't support
scylla 2024.1 where we needed to rename sstables on schema restore).

Fixes https://scylladb.atlassian.net/browse/CLOUD-2884

### DIFF
--- a/pkg/sstable/sstable_naming.go
+++ b/pkg/sstable/sstable_naming.go
@@ -1,23 +1,19 @@
-// Copyright (C) 2023 ScyllaDB
+// Copyright (C) 2026 ScyllaDB
 
 package sstable
 
 import (
-	"encoding/binary"
-	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
-	"time"
 
-	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


VAveryanov8 (approved):
Looks good to me - no other comments!

PR #4888, opened 2026-07-23
fix(repair): wait for repair goroutine to finish in repair_restart subtest

karol-kokoszka:
Closes https://scylladb.atlassian.net/browse/CLOUD-3157

Fix flaky `TestServiceRepairIntegration/repair_restart_respect_repair_control`.

**Root cause:** `repair_restart` returns without waiting for its repair goroutine to finish. The next subtest's `ensureNoActiveRepairs()` finds the still-active Scylla repair task left by the previous goroutine's incomplete cleanup.

**Fix:** Explicitly cancel and wait for `Repair()` to return before the subtest exits.

Failing CI run: https://github.com/scylladb/scylla-manager/actions/runs/30001492503/job/89187476313

### DIFF
--- a/pkg/service/repair/service_repair_integration_test.go
+++ b/pkg/service/repair/service_repair_integration_test.go
@@ -1238,6 +1238,9 @@ func TestServiceRepairIntegration(t *testing.T) {
 
 		Print("And: repair of all nodes continues")
 		h.assertProgress(100, longWait)
+
+		Print("And: wait for repair to fully stop")
+		h.assertDone(shortWait)
 	})
 
 	t.Run("repair restart respect repair control", func(t *testing.T) {


karol-kokoszka (commented):
Copilot's concern about async aborts is not applicable here: `ScyllaAbortTask` is synchronous — when it returns, the task is in a terminal state. Additionally, between the kill calls and `ensureNoActiveRepairs`, there are multiple network round-trips (cluster creation, host connectivity checks, session creation, etc.) providing ample time for any hypothetical async cleanup. No polling loop needed.

Michal-Leszczynski (commented):
Repair jobs cleanup is a part of the SM repair task functionalities. If it behaves incorrectly and leaks repair jobs after SM repair task has finished, that's a bug on the SM repair task implementation side. The other possibility is that the previous test doesn't wait for the SM repair task to finish, but that's the previous test bug. Adding this additional repair jobs killing on initializing repair test helper results in hiding such issues, so it doesn't seem like the right approach here.

karol-kokoszka:
@Michal-Leszczynski thanks for the review. 
The fix is still applied to the test, it waits until the repair is "finished" - either cancelled or completed. There is new assertion helper function added `assertFinished` - it checks if runRepair() completed the work by asserting the helper done property.

pkg/service/repair/service_repair_integration_test.go
```diff
@@ -1238,6 +1247,10 @@ func TestServiceRepairIntegration(t *testing.T) {
 
 		Print("And: repair of all nodes continues")
 		h.assertProgress(100, longWait)
+
+		Print("And: wait for repair to fully stop")
+		cancel()
+		h.assertFinished(shortWait)
```
Michal-Leszczynski: Could we use already existing `assertDone`, as we also want to validate that repair finished without any errors?
Actually, other places in the code using `assertProgress` also call it, so it was probably just missed here by mistake.

karol-kokoszka: I was not sure if assertDone is a good fit here.
assertDone checks if the repair completed without any errors including context cancelled.
AssertFinished only checks if the function returned - without asserting errors.

Michal-Leszczynski: The previous repair task execution had ctx canceled and asserted status stopped.
This repair task execution (after restart) doesn't have ctx canceled and is expected to reach 100% progress.
The problem here is that 100% progress does not mean that the repair task finished, there might still be some cleanup going on, so we also want to wait for that.
Your solution works fine in this regard, but in a test scenario, we should prefer to validate exact repair status, whenever we expect one, like here, not that it just finished without validating that it didn't failed.

karol-kokoszka: makes sense, let me change it to assertDone

Michal-Leszczynski (approved):

PR #4893, opened 2026-07-28
Don't return an error on all busy CPUs

Michal-Leszczynski:
In some (small) setups, we might not have any entirely free
CPUs to pin SM agent to. The codebase already handles it in 
findAndPinCPUs by not pinning to any specific CPU, but
setting runtime.GOMAXPROCS(1). So the codebase expects
that there might be no free CPUs to pin to and handles
this correctly. Not to mention, that the pinning error
is ignored in SM agent startup.

This commit removes the error returned by AvailableCPUs
when there is no free CPU. This was not visible on SM
agent startup, as the pinning error is simply ignored
there. It was visible in SM error logs when trying
to re-pin SM agent to CPUs after 1-1 or regular restore
tasks, but it didn't cause any failures.

This commit also adjusts AvailableCPUs comment, so that
it doesn't provide outdated information on AvailableCPUs
behavior.

Note that this is a change in vendored SM submodule,
so the actual fix will land on SM master only after
the module version is bumped.

Fixes https://scylladb.atlassian.net/browse/CLOUD-2380

### DIFF
--- a/v3/pkg/util/cpuset/cpuset_linux.go
+++ b/v3/pkg/util/cpuset/cpuset_linux.go
@@ -1,4 +1,4 @@
-// Copyright (C) 2017 ScyllaDB
+// Copyright (C) 2026 ScyllaDB
 //go:build linux
 // +build linux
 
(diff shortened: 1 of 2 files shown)


VAveryanov8 (approved):
👍

PR #4894, opened 2026-07-29
Handle protected manifests

Michal-Leszczynski:
This PR adds the notion of protected manifest - a manifest which has either event based hold set or is under not expired retention period.
Such manifests shouldn't be skipped during purge and explicit removal (sctool backup delete) of such manifests should return an error.
This is especially needed for event based hold backup, where we expect backup retention policy (--retention, --retention-days) to drift from the actual manifest protection, as:
- holds are released only during the next backup task execution
- we can't set exact retention date after hold is released - it simply starts the timer with the default bucket retention policy as soon as the hold is removed

Having said that, that's also useful for other WORM backup approaches where backup task retention policy has changed in between backup task executions.  

Fixes https://scylladb.atlassian.net/browse/CLOUD-3212

### DIFF
--- a/pkg/service/backup/list.go
+++ b/pkg/service/backup/list.go
@@ -73,15 +73,6 @@ func listRemoteManifestsInAllLocations(ctx context.Context, client *scyllaclient
 	return manifests, nil
 }
 
-// listManifests is a type wrapper for listRemoteManifests.
-func listManifests(ctx context.Context, client *scyllaclient.Client, host string, location backupspec.Location, clusterID uuid.UUID) ([]*backupspec.ManifestInfo, error) {
-	manifests, err := listRemoteManifests(ctx, client, host, location, clusterID)
-	if err != nil {
-		return nil, err
-	}
-	return manifestInfos(manifests), nil
-}
-
 // listRemoteManifests returns remote manifests for all nodes of a given cluster in the location.
 // Manifests are sorted deterministically by their ClusterID, TaskID, SnapshotTag and NodeID.
 // If cluster is uuid.Nil then it returns manifests for all clusters it can find.
@@ -251,6 +242,14 @@ func groupManifestsByTask(manifests []*backupspec.ManifestInfo) map[uuid.UUID][]
 	v := map[uuid.UUID][]*backupspec.ManifestInfo{}
 	for _, m := range manifests {
 		v[m.TaskID] = append(v[m.TaskID], m)
+	}
+	return v
+}
+
+func groupManifestsByLocation(manifests []*backupspec.ManifestInfo) map[backupspec.Location][]*backupspec.ManifestInfo {
+	v := map[backupspec.Location][]*backupspec.ManifestInfo{}
+	for _, m := range manifests {
+		v[m.Location] = append(v[m.Location], m)
 	}
 	return v
 }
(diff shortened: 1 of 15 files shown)


pkg/service/backup/service_retention_lock_integration_test.go
```diff
@@ -372,6 +376,184 @@ func TestBackupResumeOnRetentionLockStageIntegration(t *testing.T) {
...
+			assertSnapshotFiles(tagB, tagBFiles)
+
+			Print("When: last snapshot is deleted")
+			if err := h.service.DeleteSnapshot(ctx, h.ClusterID, []backupspec.Location{location}, []string{tagB}); err != nil {
+				t.Fatal(err)
+			}
+
+			Print("Then: last snapshot is deleted")
+			if got := listSnapshotFiles(t, h, tagB); len(got) != 0 {
+				t.Fatalf("Expected last snapshot %s to be deleted", tagA)
```
karol-kokoszka: Shouldn't it be tagB referenced in the message ?

pkg/service/backup/list.go
```diff
@@ -9,18 +9,52 @@ import (
...
 
 	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
 )
 
-// listManifestsInAllLocations returns manifests for all nodes of a in all
-// locations specified in hosts.
+type remoteManifestInfo struct {
+	*backupspec.ManifestInfo
+
+	RetentionMode  scyllaclient.RetentionLockMode
```
karol-kokoszka: Seems to be unused.

karol-kokoszka (approved):
Just two nits. 
Thanks ! 👍

pkg/service/backup/list.go
Michal-Leszczynski: Actually, I can change `Protected` to also take it into consideration (e.g. check `retainUntil` only when mode is unlocked/locked). 

The `scyllaclient.RetentionLockMode` is overloaded and perhaps should be changed - from `--retention-lock-mode` POV it describes backup retention mode (none/unlocked/locked/hold), but from cloud provider POV, it only describes retention lock mode, as hold is a different feature.

pkg/service/backup/service_retention_lock_integration_test.go
Michal-Leszczynski: done

pkg/service/backup/list.go
Michal-Leszczynski: I went ahead and decoupled backup side and scyllaclient sides to avoid further confusion.

Michal-Leszczynski:
v2:
- added correct snapshot tag in test failure - squashed into [b2b73bc](https://github.com/scylladb/scylla-manager/pull/4894/commits/b2b73bc4c84ce31df0cfb3ed70173101ec496dea) ([comment](https://github.com/scylladb/scylla-manager/pull/4894#pullrequestreview-4819864910))
- decoupled backup and scyllaclient retention modes to avoid type overloading and confusion - [b6a65be](https://github.com/scylladb/scylla-manager/pull/4894/commits/b6a65bec9642bb6ed4bc4d5fbfb605efb4d5c8c8)
- check retention mode when checking manifest protection - [208023e](https://github.com/scylladb/scylla-manager/pull/4894/commits/208023e570f4f0b1bfd81f1fd2aadd7ea0f5d206) ([comment](https://github.com/scylladb/scylla-manager/pull/4894#discussion_r3683557329))

Michal-Leszczynski:
@karol-kokoszka does the refactor look good to you?

PR #4895, opened 2026-08-03
Handle tmp manifests from event based hold backup

Michal-Leszczynski:
This PR targets the event based hold backup and temporary manifest interaction.
Since event based hold backup is executed with default bucket retention policy, it makes it impossible to delete tmp manifests shortly after they were uploaded. This means that we can't replace them with regular manifests, but instead we need to keep them both.
This creates friction with multiple places in SM codebase, as the existence of tmp manifests used to indicate partial or ongoing backup.

This PR starts by preparing all impacted places in SM codebase for the existence of shadowed tmp manifest.
Temporary manifest is "shadowed", when it has also corresponding regular manifest.
Most places in SM codebase should simply ignore shadowed tmp manifests.
In terms of places like purge and delete snapshot, which are also responsible for tmp manifest cleanup - shadowed manifests are also ignored (so that we don't need to read and parse the same manifest contents multiple times), but the cleanup mechanism will try to also remove tmp manifest anyway.

This PR introduces multiple integration tests, but I also executed its dirty version against CI where all backups were event based hold by default, and apart from failures related to wrong expected manifest count in some bakcup tests, all of them passed.

Fixes https://scylladb.atlassian.net/browse/CLOUD-3214

### DIFF
--- a/pkg/service/backup/worker_deduplicate.go
+++ b/pkg/service/backup/worker_deduplicate.go
@@ -121,13 +121,16 @@ func (w *worker) deduplicateHost(ctx context.Context, h hostInfo) error {
 		if err != nil {
 			return errors.Wrap(err, "deduplication based on .crc32 content")
 		}
-		if applyHolds && willCreateVersioned {
+		if willCreateVersioned && w.RetentionLockMode != RetentionLockDisabled {
 			// Creation of versioned sstable happens by copying old sstable version with appended version
 			// suffix, then removing the original old version, then uploading the new version.
 			// This is not compatible with event based hold backup, where default backup retention policy
-			// is expected to prevent files from being deleted.
+			// is expected to prevent files from being deleted. We could make it work for object retention
+			// lock backup, but since this is an edge-case that might not even exist with modern UUID based
+			// SSTables, we simply don't need to support it. This would add both implementation and performance
+			// implications, as it would require listing snapshot dirs in stage retention lock.
 			return errors.Errorf("host %s: table %s.%s: snapshot contains sstables with integer based IDs which upload would result in creation of versioned files. "+
-				"Upload of such snapshot is not compatible with event based hold backup. "+
+				"Upload of such snapshot is not compatible with --retention-lock-mode backup. "+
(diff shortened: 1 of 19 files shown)


pkg/service/backup/worker_deduplicate.go
```diff
@@ -121,13 +121,16 @@ func (w *worker) deduplicateHost(ctx context.Context, h hostInfo) error {
...
-		if applyHolds && willCreateVersioned {
+		if willCreateVersioned && w.RetentionLockMode != RetentionLockDisabled {
 			// Creation of versioned sstable happens by copying old sstable version with appended version
 			// suffix, then removing the original old version, then uploading the new version.
 			// This is not compatible with event based hold backup, where default backup retention policy
-			// is expected to prevent files from being deleted.
+			// is expected to prevent files from being deleted. We could make it work for object retention
+			// lock backup, but since this is an edge-case that might not even exist with modern UUID based
+			// SSTables, we simply don't need to support it. This would add both implementation and performance
+			// cimplications, as it would require listing snapshot dirs in stage retention lock.
```
VAveryanov8: nit:
```suggestion
			// implications, as it would require listing snapshot dirs in stage retention lock.
```

VAveryanov8 (approved):
Looks good to me and very easy to follow! 

I'm curios if there is a way for us to check whether there are any cloud clusters that still have integer ID sstables.

Michal-Leszczynski:
> I'm curios if there is a way for us to check whether there are any cloud clusters that still have integer ID sstables.

I don't know of any elegant way to do it right now, but we could add SM metric which would be set to 1 if during backup we encountered integer based sstable. This could help us in terms of getting some info on the frequency of encountering such sstables and possibly remove the support for their edge cases (this support is not only about messy code, but also about the fact that we do additional dir listings and other ops because of it).

@VAveryanov8 @karol-kokoszka I would say it would be interesting to add such a metric (labeled by cluster/nodeID/ks/tab to now where to look for such sstable). We could populate this metric to 1 only when we actually run into such sstable and remove it when we don't (not just set it to 0 or set it to -1 at the task start, so that we don't create useless series). Do you think I can create an issue for that?

PR #4898, opened 2026-08-05
swagger: agent, correctly capitalize eventBasedHold param

Michal-Leszczynski:
All SM agent handlers expect eventBasedHold, not EventBasedHold. This was overlooked because our hold test interceptor works on swagger models, so it was consistent in the client and test interceptor interaction. Since our gcs mock does not support holds, we didn't spot it until we executed tests against the real gcs.

### DIFF
--- a/v3/swagger/agent.json
+++ b/v3/swagger/agent.json
@@ -1253,7 +1253,7 @@
           "description": "Allows overriding 'unlocked' retention policy",
           "type": "boolean"
         },
-        "EventBasedHold": {
+        "eventBasedHold": {
           "description": "Whether the object is under event based hold",
           "type": "boolean"
         }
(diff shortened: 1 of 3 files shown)


VAveryanov8 (approved):
Do you think it's possible (worth) to adjust our integration tests somehow to catch similar issue in the future?

karol-kokoszka (approved):

Michal-Leszczynski:
> Do you think it's possible (worth) to adjust our integration tests somehow to catch similar issue in the future?

So in general such a thing would be caught in most other cases, but here we lack gcs mock support for event based holds and both SM client and SM test interceptor operate on the same swagger file, so this failure wasn't caught there. 

I can extend the interceptor to have some simple validation on the json fields returned from SM agent in the PR which will bump swagger dep in the main module.

PR #4900, opened 2026-08-12
Event based hold metrics

Michal-Leszczynski:
This PR adds 5 backup related metrics:
- filesCount
- filesSkippedCount
- versionedFilesCount
- setEventBasedHolds
- removedEventBasedHolds

The first two serve as the counterpart to the already existing size based backup metrics.
Third adds the observability over versioned files, so that we can monitor their existence on prod clusters and plan their deprecation.
Last two are metrics monitoring event based hold req count - when combined with new count based metrics and already existing retention_locked_files metric, they would show req count savings achieved thanks to event based hold backup approach. 

The last 3 metrics have slightly different lifecycle - SM actually removes them at the start of the backup execution instead of setting them to -1. I believe that's a better approach, as they use nodeID and ks/tab labels which could become stale over time and would keep on being counted towards metric cardinality increasing its cost.

Fixes https://scylladb.atlassian.net/browse/CLOUD-3256

### DIFF
--- a/pkg/metrics/backup.go
+++ b/pkg/metrics/backup.go
@@ -63,7 +63,12 @@ func NewBackupMetrics() BackupMetrics {
 
 // MustRegister shall be called to make the metrics visible by prometheus client.
 func (m BackupMetrics) MustRegister() BackupMetrics {
-	prometheus.MustRegister(m.all()...)
+	return m.MustRegisterWith(prometheus.DefaultRegisterer)
+}
+
+// MustRegisterWith registers all backup metrics with the given registerer.
+func (m BackupMetrics) MustRegisterWith(reg prometheus.Registerer) BackupMetrics {
+	reg.MustRegister(m.all()...)
 	return m
 }
 
--- a/pkg/metrics/backup.go
+++ b/pkg/metrics/backup.go
@@ -9,14 +9,19 @@ import (
 )
 
 type BackupMetrics struct {
-	snapshot             *prometheus.GaugeVec
-	filesSizeBytes       *prometheus.GaugeVec
-	filesUploadedBytes   *prometheus.GaugeVec
-	filesSkippedBytes    *prometheus.GaugeVec
-	filesFailedBytes     *prometheus.GaugeVec
-	purgeFiles           *prometheus.GaugeVec
-	purgeDeletedFiles    *prometheus.GaugeVec
-	retentionLockedFiles *prometheus.GaugeVec
+	snapshot               *prometheus.GaugeVec
+	filesSizeBytes         *prometheus.GaugeVec
+	filesUploadedBytes     *prometheus.GaugeVec
+	filesSkippedBytes      *prometheus.GaugeVec
+	filesFailedBytes       *prometheus.GaugeVec
+	purgeFiles             *prometheus.GaugeVec
+	purgeDeletedFiles      *prometheus.GaugeVec
+	retentionLockedFiles   *prometheus.GaugeVec
+	filesCount             *prometheus.GaugeVec
+	filesSkippedCount      *prometheus.GaugeVec
+	versionedFilesCount    *prometheus.GaugeVec
+	setEventBasedHolds     *prometheus.GaugeVec
+	removedEventBasedHolds *prometheus.GaugeVec
 }
 
 func NewBackupMetrics() BackupMetrics {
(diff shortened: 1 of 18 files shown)


Michal-Leszczynski:
Note that restore schema tests against scylla nightly are expected to fail because of scylla side issue: https://scylladb.atlassian.net/browse/SCYLLADB-3642.

pkg/metrics/backup.go
```diff
@@ -58,13 +77,39 @@ func (m BackupMetrics) all() []prometheus.Collector {
...
+		m.retentionLockedFiles,
+		m.filesCount,
+		m.filesSkippedCount,
+	} {
+		setGaugeVecMatching(c, unspecifiedValue, clusterMatcher(clusterID))
+	}
+	// Newer metrics are deleted instead of being set to unspecifiedValue,
+	// so that series of nodes and tables that are no longer part of
+	// the cluster aren't reported indefinitely.
+	for _, c := range []*prometheus.GaugeVec{
```
VAveryanov8, line 107: > The last 3 metrics have slightly different lifecycle - SM actually removes them at the start of the backup execution instead of setting them to -1. I believe that's a better approach, as they use nodeID and ks/tab labels which could become stale over time and would keep on being counted towards metric cardinality increasing its cost.


question: Can you please elaborate a bit more about why we need this new approach? (e.g. why it's ok to have previous metrics to report indefinitely). Does old metrics has lower carnality?

Michal-Leszczynski: So I would say that the previous approach is not ok, as the stale series are only cleaned on SM server restart.
If node is replace or if table is dropped, related series will be forever (until SM restart) reported with -1 value.

I didn't want to change the behavior for older metrics, as some dashboards might actually rely on checking if metric value is set to -1 opposed to check whether the metric is exposed.

I added this behavior because I introduced the versioned file metric, which should be set rarely and even if once set, it would most likely not be set during the next backup task execution.

And in general, I followed the approach of writing new metrics in a better way instead of propagating the problems of the old metrics at the cost of inconsistency.

pkg/service/backup/worker_deduplicate.go
VAveryanov8, line 163: question: As far as I understand RcloneDeletePathsInBatches returns number of files which were actually deleted, so maybe it's worth relying on the returned value instead of on the input (e.g. `len(deduplicated)`). Also I'm curious if it's possible, that RcloneDeletePathsInBatches returns error and non-zero deleted files count.

pkg/service/backup/worker_upload.go
VAveryanov8, line 115: question: I know, it's no directly related to this pr, but still: What are the cases when the snapshot will be fully uploaded up to this point? Should we execute onRunProgress in such cases too?

VAveryanov8 (approved):
LGTM: just a few non-blocking questions

pkg/service/backup/worker_deduplicate.go
Michal-Leszczynski: > As far as I understand RcloneDeletePathsInBatches returns number of files which were actually deleted, so maybe it's worth relying on the returned value instead of on the input (e.g. len(deduplicated)). 

Dedup works on files obtained by listing remote (explicitly) and local (during index stage) dirs.
So the fact that `RcloneDeletePathsInBatches` in theory could return lower value due to the fact that local file is missing doesn't really apply here, as this very file was listed recently by SM, so we always pass existing files to `RcloneDeletePathsInBatches` (and if the file was removed in the meantime, we are in some serious troubles). 

I wanted for the count based metrics to follow the size based metrics lifecycle for consistency.
The skipped size is calculated based on the listing output, so I wanted to follow the same approach here.
It would be more difficult to switch it to the actual size of deleted files, as `RcloneDeletePathsInBatches` does not report that.

> Also I'm curious if it's possible, that RcloneDeletePathsInBatches returns error and non-zero deleted files count.

From the code it looks like that's indeed possible - moreover, it doesn't perform any sort of a rollback on error.

pkg/service/backup/worker_upload.go
Michal-Leszczynski: It could happen on resume when the job uploading given dir (either rclone or native) finished before stage upload was interrupted by an error or pause. We don't need to call `onRunProgress`, as this would already be called earlier when re-indexing remaining snapshot files (some stages trigger re-indexing, as they rely on index info - see `allowReindexOnPrevStage` for more info).

PR #4905, opened 2026-08-14
Test against scylla 2026.3 instead of 2025.4

Michal-Leszczynski:
Scylla 2025.4 is no longer supported:
https://docs.scylladb.com/stable/versioning/version-support.html
Scylla 2026.3 has rc image. Because of that, scylla 2025.4 is
removed and scylla 2026.3 is added to SM CI test matrix.

This commit also bumps other available patch versions.


Michal-Leszczynski:
@VAveryanov8 please take a look

VAveryanov8 (approved):
👍

PR #4908, opened 2026-08-28
Tablet aware restore core impl

Michal-Leszczynski:
This PR introduces the core implementation around tablet aware restore in SM ([design doc](https://scylladb.atlassian.net/wiki/spaces/RND/pages/237928765/Tablet-aware+restore+MS1)).
This consists of 2 main parts:
- tablet aware restore workload indexing and validation
- restoration of indexed workload

For now, tablet aware restore is not exposed to the user - it's toggled with restore service config internal field.
Next PRs adding proper support for tablet aware restore method will address that.

Fixes https://scylladb.atlassian.net/browse/CLOUD-780
Fixes https://scylladb.atlassian.net/browse/CLOUD-782
Fixes https://scylladb.atlassian.net/browse/SCYLLADB-2526


pkg/util2/topology/topology.go
```diff
@@ -0,0 +1,132 @@
...
+	if hosts := session.GetHosts(); len(hosts) > 0 && !singleHostSession {
+		hostID = hosts[0].HostID()
+	}
+	var dc, rack string
+
+	si := new(IterWithErr)
+	si.Iter = func(yield func(string, string) bool) {
+		it := session.QueryWithContext(ctx, "SELECT data_center, rack FROM system.peers").SetHostID(hostID).Iter()
+		for it.Scan(&dc, &rack) {
+			yield(dc, rack)
```
VAveryanov8: since it's a separate package now, I think it's worth starting checking returning values of yield, otherwise if someone will use `break` statement with this iterator, the iterator will panic, e.g. https://go.dev/play/p/N8yqbwtB5u-

pkg/service/restore/tablet/worker.go
```diff
@@ -0,0 +1,136 @@
...
+		longPollingSeconds: longPollingSeconds,
+	}
+}
+
+// Restore given tables with tablet aware restore based on provided workload.
+func (w *RestoreWorker) Restore(ctx context.Context, tables Workload) error {
+	w.logger.Info(ctx, "Started tablet aware restore")
+	defer w.logger.Info(ctx, "Finished tablet aware restore")
+
+	eg := errgroup.Group{}
```
VAveryanov8, line 55: question: Should we introduce some limit to the concurrency here?

pkg/service/restore/index.go
```diff
@@ -72,22 +78,76 @@ type SSTable struct {
...
 		}
-		rawWorkload = append(rawWorkload, lw...)
+	}
+
+	tables := make(map[cqlTable.CQLTable]struct{})
+	for _, u := range w.run.Units {
+		for _, t := range u.Tables {
+			tables[cqlTable.CQLTable{Keyspace: u.Keyspace, Name: t.Table}] = struct{}{}
+		}
+	}
```
VAveryanov8: nit: We have 6 identical lines in indexTabletAwareWorkload

pkg/service/restore/index.go
```diff
@@ -72,22 +78,76 @@ type SSTable struct {
...
+	// Don't index tables already indexed and compatible with tablet aware restore
+	for t := range tabletWorkload {
+		delete(tables, t)
+	}
+
+	rawWorkload, err := w.indexLocationWorkload(ctx, tables, manifests)
+	if err != nil {
+		return Workload{}, errors.Wrapf(err, "index workload")
```
VAveryanov8: nit: Wrap**f** but no actual formatting is used

VAveryanov8 (approved):
I like the way you've used feature flag! 
Pre-approving with some questions/nits

pkg/service/restore/tablet/worker.go
Michal-Leszczynski: From scylla POV, all should be scheduled in parallel with no limit, so we only need to worry on SM side.
Currently, we spawn one goroutine per table and don't make other significant allocations. 
If we assume that goroutine consumes 2 kb of memory initially, we should be safe with limiting it to 1000.
TBH, we could set the limit higher, but I don't see a good reason for that.
Would it make sense to you?

VAveryanov8: Yes, let's have some big enough limit - 1000 IMHO is high enough - we should also count not only amount of memory that goroutine consumes itself, but also what this goroutine is doing, e.g. sending network requests and etc.

Michal-Leszczynski: Good point - with this amount of concurrent requests, most of the connections would need to be opened from scratch and not reused from the scyllaclient connection pool.

So I decided to limit it to cluster shard cnt and clamp it to [100, 1000].

I also added round-robin for choosing tablet aware restore coordinator node, so that connections are more evenly spread across the node and SM agents.

PR #4909, opened 2026-08-28
fix(repair): repair all replicas of a small table

karol-kokoszka:
Closes https://scylladb.atlassian.net/browse/CLOUD-3672

A table under `--small-table-threshold` (default 1 GiB) is repaired with a single `small_table_optimization` call covering the whole ring, and the remaining replica sets are skipped. SM still set the `hosts` param of that call to a **single replica set**.

Scylla resolves the participants of a small table repair as all normal token owners, but then intersects them with `hosts`, so passing one replica set silences every replica outside it. On a ring with more nodes than RF those replicas were never repaired — while SM still reported the table as fully repaired.

Reported by @asias on scylladb/scylladb#31271:

> Assume we have 4 nodes in the cluster, RF=3. Does SM send one repair request that limit the hosts to 3 nodes (say n1,n2,n3 which are the owners of the token range that SM uses)? This is problematic because we never repair n4 due to the hosts filter.

## Fix

- **No host filter requested:** send no `hosts` limit, so all token owners take part.
- **Filter requested** (`--dc` / `--host` / `--ignore-down-hosts`): pass all table replicas that passed the filtering, so the limit is still honored. Such a repair does not cover all replicas, so it does not advance the tombstone GC repair time in Scylla.

The nodes sent as `hosts` are kept in a new `job` field rather than reusing `replicaSet`, which is the unit of the intensity controller bookkeeping and of the recorded per host progress. Whether a filter was requested cannot be derived from `Target.DC` (filled with all cluster DCs when `--dc` is unset), so it is recorded while parsing the task properties.

## Test

The new integration test validates that the table is spread across all nodes and that the ring has more than one replica set, then snapshots Scylla's per node repair row hash counters, runs a real repair, and requires every node to have moved them. The counters are bumped on both sides of the row hash exchange even with nothing to reconcile, so a participating node always moves them and an excluded node never does.

**Before the fix** — `.12` and `.23` hold ~900 KB of the table each, yet do zero repair work:

```
Table size on 192.168.200.11: 891376 bytes
Table size on 192.168.200.12: 906160 bytes
Table size on 192.168.200.13: 901872 bytes
Table size on 192.168.200.21: 893650 bytes
Table size on 192.168.200.22: 887484 bytes
Table size on 192.168.200.23: 914054 bytes

Repair work on 192.168.200.11: 1446 -> 2047   (delta 601)
Repair work on 192.168.200.12: 17901 -> 17901 (delta 0)
Repair work on 192.168.200.13: 16174 -> 16433 (delta 259)
Repair work on 192.168.200.21: 34983 -> 35151 (delta 168)
Repair work on 192.168.200.22: 30907 -> 31081 (delta 174)
Repair work on 192.168.200.23: 3843 -> 3843   (delta 0)

Nodes [192.168.200.12 192.168.200.23] did no repair work, so they were not repaired
--- FAIL: TestServiceRepairSmallTableOptimizationRepairsAllNodesIntegration (8.79s)
```

**After the fix** — every node does repair work:

```
Repair work on 192.168.200.11: 2047 -> 2220   (delta 173)
Repair work on 192.168.200.12: 17901 -> 18076 (delta 175)
Repair work on 192.168.200.13: 16433 -> 16605 (delta 172)
Repair work on 192.168.200.21: 35151 -> 36107 (delta 956)
Repair work on 192.168.200.22: 31081 -> 31256 (delta 175)
Repair work on 192.168.200.23: 3843 -> 4104   (delta 261)

--- PASS: TestServiceRepairSmallTableOptimizationRepairsAllNodesIntegration (8.23s)
```

Full `./pkg/service/repair` integration suite passes (12/12).

Two golden files gain `host_filter: true` — `filter_dc` and `complex` are the only `GetTarget` inputs that set `dc`.

### DIFF
--- a/pkg/service/repair/generator.go
+++ b/pkg/service/repair/generator.go
@@ -5,6 +5,7 @@ package repair
 import (
 	"context"
 	"net/netip"
+	"slices"
 	"sync/atomic"
 
 	"github.com/pkg/errors"
@@ -73,13 +74,19 @@ func (jt jobType) fullTableRepair() bool {
 }
 
 type job struct {
-	keyspace   string
-	table      string
-	master     netip.Addr
+	keyspace string
+	table    string
+	master   netip.Addr
+	// replicaSet is the replica set of the repaired ranges. It is the unit of
+	// the controller bookkeeping and of the recorded progress.
 	replicaSet []netip.Addr
-	ranges     []scyllaclient.TokenRange
-	intensity  int
-	jobType    jobType
+	// hosts is sent as the 'hosts' param of the repair API call, limiting the
+	// nodes taking part in the repair. Empty means no limit, which lets Scylla
+	// pick the participants on its own.
+	hosts     []netip.Addr
+	ranges    []scyllaclient.TokenRange
+	intensity int
+	jobType   jobType
 }
 
 type jobResult struct {
@@ -280,6 +287,7 @@ func (tg *tableGenerator) newJob() (job, bool) {
 				table:      tg.Table,
 				master:     tg.ms.Select(filtered),
 				replicaSet: filtered,
+				hosts:      tg.repairHosts(jt, filtered),
 				ranges:     ranges,
 				intensity:  intensity,
 				jobType:    jt,
@@ -288,6 +296,49 @@ func (tg *tableGenerator) newJob() (job, bool) {
 	}
 
 	return job{}, false
+}
+
+// repairHosts returns the nodes to be sent as the 'hosts' param of the repair
+// API call for a job of given jobType repairing given replica set.
+//
+// A small table repair is performed with a single API call covering the whole
+// ring, so it spans the ranges of every replica set - not just the one the job
+// was created for. Scylla resolves the participants of such a call as all of
+// the normal token owners, but it intersects them with the 'hosts' param, so
+// passing a single replica set there would leave every replica outside of it
+// unrepaired, see CLOUD-3672.
+//
+// Because of that, no 'hosts' limit is sent for such calls, unless the user
+// narrowed down the nodes taking part in the repair - in which case the limit
+// has to be honored, and all of the table replicas that passed the filtering
+// are sent instead. Note that a filtered repair does not cover all of the
+// replicas, so it does not advance the tombstone gc repair time in Scylla.
+func (tg *tableGenerator) repairHosts(jt jobType, filtered []netip.Addr) []netip.Addr {
+	if jt != smallTableJobType {
+		return filtered
+	}
+	if !tg.target.HostFilter {
+		return nil
+	}
+	return tg.allFilteredReplicas()
+}
+
+// allFilteredReplicas returns all of the table replicas that passed the
+// --dc / --host / --ignore-down-hosts filtering.
+func (tg *tableGenerator) allFilteredReplicas() []netip.Addr {
+	var out []netip.Addr
+	for _, rt := range tg.Ring.ReplicaTokens {
+		for _, h := range filterReplicaSet(rt.ReplicaSet, tg.Ring.HostDC, tg.target) {
+			if !slices.Contains(out, h) {
+				out = append(out, h)
+			}
+		}
+	}
+	// Ensure deterministic order of nodes.
+	slices.SortFunc(out, func(a, b netip.Addr) int {
+		return a.Compare(b)
+	})
+	return out
 }
 
 func (tg *tableGenerator) getRangesToRepair(allRanges []scyllaclient.TokenRange, intensity int) []scyllaclient.TokenRange {
--- a/pkg/service/repair/service.go
+++ b/pkg/service/repair/service.go
@@ -150,6 +150,10 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
 	if err := validateIgnoreDownNodes(t, status); err != nil {
 		return t, err
 	}
+	// Record whether the user narrowed down the set of nodes taking part in the
+	// repair. It has to be taken from the raw properties, as t.DC is filled with
+	// all of the cluster DCs when --dc is not set.
+	t.HostFilter = len(props.DC) > 0 || props.Host != "" || props.IgnoreDownHosts
 
 	// Get potential units - all tables matched by keyspace flag
 	f, err := ksfilter.NewFilter(props.Keyspace)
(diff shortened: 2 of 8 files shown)


Michal-Leszczynski:
So this is indeed a bug, but let's actually check its radius.

Small table opt is vnode only, so it doesn't touch tablets.

For small vnode table with tombstone_gc mode repair, this could result in unrepaired data and tombstone accumulation. Luckily, if the tombstone acc makes the table go over small table limit (1GB by default), table would be repaired as a regular one, fixing those issues. IIUC, it wouldn't result in data loss or resurrection (at least in cases where write_cl+read_cl > replica_cnt).

The most notable problem is with small vnode table using tombstone gc mode timeout, where not repairing the data could simply lead to data resurrection.

This PR highlights that we can't set DC filter in scylla repair API req to all DCs when it's not set on SM side, as this would make it so the repair timestamp is not advanced on scylla side. In the context of this particular issue, that's not really a factor, as this issue is for vnode tables only.
In vnodes, setting DC filter would still result in advancing the timestamp and correct interaction with tombstone_gc mode repair. At the end of the day, regular vnode repair always sets the host filtering for selecting desired replica set and it always repairs the table with multiple API calls.

But it looks like we have another bug, which also looks more serious - SM always sets DCs when calling tablet repair API.
For tablets, this would mean that we never advance repair timestamp on scylla side.
This messes up the interaction with tombstone_gc mode repair and incremental repair.
This is what I get from looking at the implementation, but it would be really surprising if it were true and we didn't observe this in production (or perhaps we saw issues related to that, but we never connected the dots).

One lucky thing is that the new SM tablet repair task (separate from the general purpose repair task) does not accept any filtering params and does not set them when sending tablet repair requests to scylla. All newer cluster (also every XCloud cluster) are using it, so perhaps we just didn't have too many tablet production cases back then. Anyway, this is also something that should be addressed, if verified.

Michal-Leszczynski:
I will take a closer look at the implementation today, but just to initial mentions:
- it seems like handling of the dc/host filtering for small table opt and tablet table cases should be identical. Tablet case already relies on the list of ignored nodes established on target creation. To fix this issue, and the tablet issue, we could simply add IgnoredDCs target field and plug it into small table opt and tablet repair in the same way.
- I don't really like tests relying on scylla metrics - this is not a clear API contract, we basically guess on our side what the metric does, but we don't have any expertise in this topic. I would say it's enough to focus on what requests SM is sending and leave the exact result of interaction between given req param and metrics somewhere else - it's less e2e. but it's also not the best place for a full e2e test - that should be some DTests or SCT - which also means that we are probably lacking some coverage there. Having said that, this test performs a deep validation, which is useful when dealing with APIs which don't have much documentation, like in the case of scylla API.

pkg/service/repair/service.go
```diff
@@ -150,6 +150,10 @@ func (s *Service) GetTarget(ctx context.Context, clusterID uuid.UUID, properties
 	if err := validateIgnoreDownNodes(t, status); err != nil {
 		return t, err
 	}
+	// Record whether the user narrowed down the set of nodes taking part in the
+	// repair. It has to be taken from the raw properties, as t.DC is filled with
+	// all of the cluster DCs when --dc is not set.
+	t.HostFilter = len(props.DC) > 0 || props.Host != "" || props.IgnoreDownHosts
```
Michal-Leszczynski, line 156: The problem here is that `props.DC` is not an exact list of DCs - it's a DC pattern. We can have things like `--dc=*`, but this if would still classify it as applied host filter. We are also repairing keyspaces with different replications (e.g., replicated only in dc1, replicated in dc1 and dc2).

pkg/service/repair/generator.go
```diff
@@ -73,13 +74,19 @@ func (jt jobType) fullTableRepair() bool {
...
+	// replicaSet is the replica set of the repaired ranges. It is the unit of
+	// the controller bookkeeping and of the recorded progress.
 	replicaSet []netip.Addr
-	ranges     []scyllaclient.TokenRange
-	intensity  int
-	jobType    jobType
+	// hosts is sent as the 'hosts' param of the repair API call, limiting the
+	// nodes taking part in the repair. Empty means no limit, which lets Scylla
+	// pick the participants on its own.
+	hosts     []netip.Addr
```
Michal-Leszczynski, line 86: I believe this is the right way to go - passing host filter from generator to worker.

As mentioned below, the way in which this field is set needs to be fixed.
Moreover, the fix for another issues (always setting DC filter on tablet repair), should probably be addressed in a very similar way (e.g., by passing dc filter as the job field).

To be more robust and also respect different replication strategies (e.g., repairing ks replicated in dc1 only with --dc=dc1 does not need filtering, while repairing ks replicated in dc1 and dc2 does need it), I propose to calculate hosts filter in generator based on the described ring.

For all full table repairs, we first calculate the full replica set - union of all replica sets.
Then we apply SM dc/host filtering on it.
If any of those filters removed some replicas from the full set, we need to set it in the repair request.

One additional benefit of this solution is that we also fix one more underlying problem.
When full table repair happens, the first job we issue to repair worker is the real one performing all the work, but for the purposes of progress reporting, we also send dummy jobs which simply update DB state. The problem was that since the first job had some random first replica set attached to it, it blocked only this replica set in repair controller. This allowed for dummy jobs to run and finish while the real one was still ongoing. This resulted in having 95% progress within the first second of full table repair, while the last 5% included the real work.

Calculating the real full replica set for the real full table job blocks all replicas, so it would block dummy jobs from succeeding before the real one finishes. 

This issue somewhat a result of not adjusting repair progress pooling to full tablet repair case, but even if we were to do so in some time, change proposed in this comment should make it easier to implement. 

@karol-kokoszka could I take the PR from here and:
- calculate real replica set for full table repair job (all ring replicas)
- decouple full table repair jobs from progress recording (we are sending full replica set and no ranges to the worker - such jobs shouldn't update the progress. Instead, we should follow with skip jobs just for the purpose of updating the progress)
- fix the way in which host filter is set for repair job (apply filter on all ring replicas and check if anything was filtered out - if so, set hostFilter field on repair job)
- add dc filter alongside host filter for repair job (same approach as above)
- keep your e2e test in place
- add tests for expected API params

Those shouldn't be big changes, but I think it would be easier for me to implement them, especially that you are on bchores for this week.

karol-kokoszka:
Thanks for the deep review — agreed on all of it, please take it from here.

Michal-Leszczynski:
Superseded by https://github.com/scylladb/scylla-manager/pull/4911.

PR #4911, opened 2026-09-03
Fix repair filtering params

Michal-Leszczynski:
This PR supersedes https://github.com/scylladb/scylla-manager/pull/4909 mostly according to https://github.com/scylladb/scylla-manager/pull/4909#discussion_r3901884699.
It's goal is to fix the following:
- for small vnode table opt repair - pass dc/host filter not limited by the first encountered replica set, but based on the full table replica set (resulted in partial repair)
- for tablet repair - pass dc/host filter only when it's not a no-op (resulted in not advancing table's repair timestamp)

This PR does not tackle the problem related to including partial replica set in repair job for full table repairs (causes misleading progress and unnecessary DB writes), as fixing it turned out to be slightly more complex and would require more changes to full table repair/progress flow. Therefor it's better suited for a separate PR (probably as a part of addressing https://scylladb.atlassian.net/browse/CLOUD-2919).

The original PR didn't tackle the tablet repair problem.
Also, it tired to use the same approach for detecting filtering as on the scylla side:
If filtering flags like --dc, --ignore-down-hosts, --host are set, then filtering is set.

First of all, this is not the best behavior from UX POV (setting --dc=dc1 on a single dc cluster shouldn't have negative consequences), but it makes more sense on scylla side due to implementation difficulties, repair scope limited to a single table and exact (as opposed to SM pattern) input.

For SM, this would mean that filtering is applied even when --dc=* (pattern usage) or when --ignore-down-hosts but all are up. We should be more robust than that and actually check if given filtering configuration is a no-op or not before sending the filtering params to scylla.

This PR fixes that by calculating dc/host filtering params on the generator side.
Generator has the exact info about repaired token ring and all its replicas, so it can easily see if host or dc filtering is a no-op. Then, repair worker should simply apply filters passed alongside repair job instead of trying to set them based on replica set and and target.

This PR cherry-picks the extensive e2e test validating actually repaired ranges via scylla metrics.
It also adds/extends a few tests based on sent repair request API params. 

Fixes https://scylladb.atlassian.net/browse/CLOUD-3672

### DIFF
--- a/pkg/service/repair/helper_integration_test.go
+++ b/pkg/service/repair/helper_integration_test.go
@@ -12,6 +12,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"net/netip"
 	"os"
@@ -30,6 +31,7 @@ import (
 	. "github.com/scylladb/scylla-manager/v3/pkg/testutils/db"
 	. "github.com/scylladb/scylla-manager/v3/pkg/testutils/testconfig"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/httpx"
+	"github.com/scylladb/scylla-manager/v3/pkg/util/prom"
 	"github.com/scylladb/scylla-manager/v3/pkg/util/uuid"
 	slices2 "github.com/scylladb/scylla-manager/v3/pkg/util2/slices"
 	"github.com/scylladb/scylla-manager/v3/swagger/gen/scylla/v1/client/operations"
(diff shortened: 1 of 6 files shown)


Michal-Leszczynski:
Just realized that one small change is needed - will make the PR ready for review shortly.

Michal-Leszczynski:
PR ready for review.
After making those changes, I came to the conclusion that repair filtering tests are not organized properly - they are scattered across different subtests and it's not easy to see what is actually validated. I created https://scylladb.atlassian.net/browse/CLOUD-3815 to address that, but I didn't want to include it in this PR not to stall the fixes.

pkg/service/repair/helper_integration_test.go
```diff
@@ -738,3 +746,75 @@ func chanClosedWithin(t *testing.T, c chan struct{}, d time.Duration) {
...
+// Scylla exposes its metrics on the second test network, not on the one the API
+// is served on, so the host is translated between the two.
+func nodeRepairWork(t *testing.T, host string) float64 {
+	t.Helper()
+
+	metricsHost, err := secondNetHost(host)
+	if err != nil {
+		t.Fatalf("Translate %s to the second test network: %s", host, err)
+	}
+	u := "http://" + net.JoinHostPort(metricsHost, scyllaPrometheusPort) + "/metrics?name=repair_hashes_nr"
```
karol-kokoszka:
```suggestion
	u := "http://" + net.JoinHostPort(metricsHost, scyllaPrometheusPort) + "/metrics"
```
`repair_hashes_nr` is not a metric family — the two we read below are `repair_rx_hashes_nr` and `repair_tx_hashes_nr` (cf. `ShardCount`, which queries the family name minus the `scylla_` prefix). This is harmless today only because Seastar's exporter reads `__name__`, not `name`, so the param is ignored and we get the full dump anyway. On a build that honours `name` the filter would match nothing and we'd hit `t.Fatalf("... does not expose scylla_repair_rx_hashes_nr")` against a perfectly healthy cluster.

karol-kokoszka (approved):
Approved.

Posting the review notes in full, including the findings that did not survive scrutiny — better on record than re-raised later by someone repeating the same reasoning.

## Effort

~1h50m, of which ~25 min was automated cross-checking (five verification passes, ~741k tokens, 291 tool calls) and the rest manual analysis. Outcome: **0 blockers, 0 must-fix.** About 20 items were raised; 11 were withdrawn, ~9 survive as nits.

## The findings that turned out to be invalid

**Inferring Scylla's behaviour from the API surface instead of reading Scylla's code.** The largest category.

- *"Setting both `dcs_filter` and `hosts_filter` may be wrong"* — `locator::tablet_task_info::selected_by_filters` ANDs them, and empty means unrestricted. The wider-than-necessary host filter this PR emits is functionally correct.
- *"A range with a single surviving replica could fail the whole small-table repair"* — `small_table_optimization` replaces the neighbour set with all normal token owners *before* applying `hosts` (`repair.cc:330-334`), so the "at least two endpoints" error cannot fire on this path. The PR in fact fixes the RF=2-with-a-down-node case rather than exposing it.
- *"A stale `hosts_filter` naming a dead node breaks the repair"* — `repair_range` filters neighbours by `gossiper.is_alive`, and `mandatory` is empty for tablets, so a dead node is dropped rather than fatal. The real delta is a spurious `nr_failed_ranges++`, which a retry clears.

**Inventing a failure scenario without checking the guard that already handles it.** Argued the new e2e test would break confusingly if the keyspace were changed to RF 3/3. The `len(ring.ReplicaTokens) < 2 || allReplicas.Size() <= ring.RF` guard refuses to run in exactly that topology, with an explanatory message.

**Claiming version risk without checking the support matrix.** Flagged `?name=repair_hashes_nr` as a hard failure "on a build that honours the param". Only seastar 20.05 ever read `name`; everything from 22.11 onward reads `__name__`. No supported Scylla is affected.

**Asserting something a grep would have settled.** Claimed `test_repair` did not exist in the subtest whose `IsTabletKeyspace` branch was removed, so the branch was dead. It is created at `service_repair_integration_test.go:1023` and *is* a tablet keyspace there — the condition was always true, the opposite of the claim.

**Real mechanism, unreachable consequence.** Small-table jobs repair nodes the controller never reserves via `TryBlock`. True, but table generators run sequentially and post-first-job jobs are `skipJobType` making no API call, so nothing can contend.

**Attribution errors.** Two items were pre-existing and untouched by this PR: the `len(replicaSet) > 1` guard in `client_scylla.go` (not in the diff), and `dcs_filter` carrying DCs with no replicas of the table (pre-PR passed `w.target.DC` verbatim too).

## What is worth acting on

The description undersells the tablet fix. `dcfilter.Apply` guarantees `target.DC` is never empty, and pre-PR `fullTabletTableRepair` passed it unconditionally — so **every** tablet repair sent a `dcs_filter`, meaning incremental repair and `tombstone_gc=repair` were broken for all tablet repairs, not only filtered ones. Worth stating outright.

Optional, none blocking:

- `:2186` — "Repairing tablet table should succeed" asserts nothing if no tablet repair request arrives, unlike its siblings
- `:2313` — the expected host filter is built from `HostIDs()` (all cluster nodes) rather than the table's ring, which is what the implementation uses
- `helper_integration_test.go:770` — `?name=repair_hashes_nr` matches no metric family; inert today since seastar reads `__name__`
- the `--host` change disabling small-table optimization is user-visible and is not mentioned in any commit message

PR #4913, opened 2026-09-10
Tablet aware restore progress

Michal-Leszczynski:
This PR adds tablet aware progress representation.
Adds tools for filling, querying and resuming from it.

Here is the DB representation (more context on why we need separate representation can be found in the commit messages):
```cql
CREATE TABLE restore_run_progress_tablet (
    cluster_id           uuid,
    task_id              uuid,
    run_id               uuid,

    keyspace_name        text,
    table_name           text,

    host                 text,
    scylla_task_id       text,
    started_at           timestamp,
    completed_at         timestamp,
    restored_sstables    bigint,
    total_sstables       bigint,
    size                 bigint,
    error                text,
    PRIMARY KEY ((cluster_id, task_id, run_id), keyspace_name, table_name)
) WITH default_time_to_live = 15552000;
```
with a matching go struct:
```go
// RunProgress is tablet aware restore progress representation in SM DB.
type RunProgress struct {
	// Partition key
	ClusterID uuid.UUID
	TaskID    uuid.UUID
	RunID     uuid.UUID
	// Clustering key
	Keyspace string `db:"keyspace_name"`
	Table    string `db:"table_name"`

	// Host on which tablet aware restore task was scheduled.
	// Task status can be queried only on this host.
	Host         string
	ScyllaTaskID string `db:"scylla_task_id"`
	// StartedAt is set to the scylla task timestamp
	// with SM side fallback if it's not available.
	StartedAt time.Time
	// CompletedAt is set to the scylla task timestamp
	// with SM side fallback if it's not available.
	// CompletedAt is set only when scylla task reached terminal state (done or failed).
	CompletedAt time.Time
	// Scylla reports progress in restored and total sstables
	RestoredSSTables int64 `db:"restored_sstables"`
	TotalSSTables    int64 `db:"total_sstables"`
	// Size of the restored table taken from manifest
	Size  int64
	Error string
}
```

This PR adds the following basic progress features:
- DB progress is cloned from previous run on resume, as in general restore (63fa1395)
- DB progress is updated on each long poll call (1502e466)
- On resume, already successfully restored tables are not re-restored (63fa1395)

Original issue was also mentioning the need to re-attach to ongoing task instead of issuing new tablet aware restore call. I didn't implement that, as tablet aware restore progress is already tracked on scylla side and re-restoring table with tablet aware restore continues from the last saved progress on its own. Because of that, it didn't seem necessary.

Note that this scylla side functionality has some issue, e.g., it won't allow to re-restore the table after it was truncated or dropped and recreated: https://scylladb.atlassian.net/browse/SCYLLADB-4472, but this needs to be addressed on scylla side.

This PR also fixes some preexisting issues:
- scylla restore tasks are now aborted only on pause, not on any error (284812d7)
- since aborting restore task is async on scylla side, SM now waits up to a grace period for the aborted tasks to finish before finishing the SM task (284812d7)
- more uniform distribution of restore host coordinators (54517ca0)
- long poll calls have now SM side timeout set, so that they don't hang on scylla issue (b632cdd8)
- don't skip cloning prev general restore progress on partial row (fd9d5b79)
- don't reduce scylla user task TTL during backup/restore procedures (1034cb41)
- don't log tablet aware restore messages on non tablet aware restore run (399171c4)

Fixes https://scylladb.atlassian.net/browse/CLOUD-784


pkg/service/restore/tablet/progress.go
```diff
@@ -0,0 +1,68 @@
...
+	q := schematable.RestoreRunProgressTablet.InsertQuery(session)
+	defer q.Release()
+	for i := range prev {
+		pr := &prev[i]
+		if !pr.isSuccess() {
+			continue
+		}
+		pr.RunID = runID
+		if err := q.BindStruct(pr).Exec(); err != nil {
+			return errors.Wrapf(err, "clone progress of table %s.%s", pr.Keyspace, pr.Table)
```
VAveryanov8: question: Considering that caller of CloneProgress ignoring the error (only logs), maybe it's worth continue on error (instead of returning), so if cql request will fail for one row, we still will be able to retrieve progress of others?

VAveryanov8 (approved):
LGTM!

pkg/service/restore/tablet/progress.go
Michal-Leszczynski: good point! I changed it to join errors instead.

PR #4914, opened 2026-09-15
Take MinIo images from quay.io

Michal-Leszczynski:
This is an ad-hoc workaround for CLOUD-3979,
just so that SM test env and CI can continue
to work while we prepare to migrate to minio
alternative.

It also contains additional commit adjusting restore
schema test golden files to the new system_schema
tables recently merged to scylla master.

Refs https://scylladb.atlassian.net/browse/CLOUD-3979


VAveryanov8 (approved):
👍

PR #4916, opened 2026-09-18
Don't leak goroutine in RcloneListDirIter

Michal-Leszczynski:
This series of commits addresses multiple problems with RcloneListDirIter implementation:
- returns partial results (without an error) on the next token lookup error (e8c4eab)
- decoder goroutine can leak on listing timeout (f69a25a)
- callback might be called after RcloneListDirIter already returned (59813f5)
- might hang if initial response header doesn't arrive (2f243fe)

This PR addresses that by rewriting chunks of RcloneListDirIter with ctx cancellation oriented implementation.
This also allows to get rid of the decoder goroutine once and for all and makes it easier to validate the final implementation correctness (03eec76).

More details can be found in the commit messages.

This is the first (and most important) part of work around https://scylladb.atlassian.net/browse/CLOUD-3984.

Refs https://scylladb.atlassian.net/browse/CLOUD-3984


pkg/scyllaclient/client_rclone.go
```diff
@@ -671,66 +704,70 @@ func (c *Client) RcloneListDirIter(ctx context.Context, host, remotePath string,
...
+	if err := readListStart(dec); err != nil {
+		return err
 	}
 
-	errCh := make(chan error)
-	go func() {
+	// Cancelling ctx (timeout or parent) fails any pending body read,
+	// so the stream can be decoded inline.
+	for dec.More() && ctx.Err() == nil {
 		var v RcloneListDirItem
```
karol-kokoszka: Nit: escape analysis moves `v` to the heap every iteration now, one alloc per item. Declaring it before the loop and resetting with `v = RcloneListDirItem{}` keeps the old single allocation.

pkg/scyllaclient/client_rclone.go
```diff
@@ -629,14 +629,32 @@ func (c *Client) RcloneListDir(ctx context.Context, host, remotePath string, opt
...
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
+	// The inactivity timeout is expressed as cancellation of this ctx with
+	// ErrRcloneListDirTimeout as cause. Cancelling the request ctx makes
+	// net/http close the connection, which fails any pending body read.
+	// Note that f might still block, as it doesn't accept ctx as arg,
+	// so it's user responsibility to make sure that it doesn't happen.
+	ctx, cancel := context.WithCancelCause(ctx)
+	defer cancel(nil)
+	// Difference ErrRcloneListDirTimeout from other ctx errors.
```
karol-kokoszka: Typo: Distinguish.

karol-kokoszka (approved):
LGTM. Worth noting in the description that this covers defect 1 of CLOUD-3984 only; the dedup peak-memory part is still open.

pkg/scyllaclient/client_rclone.go
Michal-Leszczynski: good catch! It was changed because reusing the same variable meant that caller can't retain it, as it would be overwritten on the next callback. This is the previous behavior, so all current callers respect that. I will bring back the previous behavior and add func comment about it.

PR #4918, opened 2026-09-22
feat(testing): add Grafana and Scylla Manager scraping to the dev env (CLOUD-4187)

karol-kokoszka:
Closes https://scylladb.atlassian.net/browse/CLOUD-4187 

Dev-environment plumbing so that the Scylla Manager Grafana dashboard can be developed against real metrics without standing up the ScyllaDB Monitoring stack. No product code is touched, and no dashboard ships here.

Split out of #4917, which carries the metrics and the dashboard itself. The two commits are unchanged from that branch.

## Why

Two things were missing from the dev environment:

**Nothing to draw metrics with.** Prometheus collected them, but every dashboard change meant hand-editing JSON and guessing, or importing into an external Grafana by hand.

**Scylla Manager was not scraped at all.** The only job left was the consul one, which discovers the *managed Scylla nodes* and not the Manager itself, so none of the Manager's own metrics were collected.

## What

**`feat(testing): add Grafana to the dev env`**

- `http://localhost:3000`, anonymous admin, dev env only
- Prometheus datasource provisioned with a fixed uid so provisioned dashboards can reference it, resolved as `http://prometheus:9090` so it works for both the ipv4 and ipv6 environments
- file provider watching `/monitoring`, picking up any `*.json` within 10s
- UI updates and deletion both disabled, so the files stay the source of truth. With UI updates allowed, saving in the UI bumps the version in Grafana's database past the file's and every later change to the file is silently ignored; with deletion allowed, checking out a commit that predates a dashboard makes Grafana drop it and not bring it back

Dashboards are mounted from `/monitoring` rather than from under `testing/`, because they are meant for the ScyllaDB Monitoring stack - the dev environment is where they are developed, not where they belong. This PR adds that directory with a README and no dashboards.

**`feat(testing): scrape Scylla Manager and add node-exporter to the dev env`**

- `scylla_manager` static scrape job, named after the production one. Static rather than discovered, because the Manager does not advertise itself through its own consul endpoint and the server container joins the network after the cluster is up
- `node-exporter` scraped under that same job, because that is how the Manager VM is scraped in production: the Manager and the node exporter beside it share it. Matching it here is what lets one dashboard query work against both environments
- without the exporter there is no signal at all for the host Scylla Manager runs on being the thing in trouble - SM's own process metrics say nothing about memory pressure caused by something else on the same host, or about swap, which the Go process collector does not expose at all

## Testing

Both were exercised continuously while building the dashboard in #4917: the Grafana instance rendered it, and every panel query was run against this Prometheus. Metrics from the `scylla_manager` job, `node-exporter` included, returned data.

https://scylladb.atlassian.net/browse/CLOUD-4187

🤖 Generated with [Claude Code](https://claude.com/claude-code)


testing/grafana/provisioning/dashboards/dashboards.yaml
```diff
@@ -0,0 +1,24 @@
...
+    # stop showing up with nothing logged. Dashboards can still be opened and
+    # tweaked in the UI to try something out, and exported from there, they
+    # just cannot be saved back into Grafana's database - which is wiped by
+    # "make down" anyway.
+    allowUiUpdates: false
+    # Deletion is disabled because the provider watches a directory inside a
+    # git checkout: switching to a branch or commit that predates a dashboard
+    # makes the file vanish, and with deletion enabled Grafana drops the
+    # dashboard and does not bring it back when the file returns.
+    disableDeletion: true
```
Michal-Leszczynski: is this really the case? I would expect that it goes against this commit message:
```
- /monitoring is watched by a file provider, so any *.json dropped there
  shows up within 10s.
```
you say that brand new file is added, but if a file existed, was removed and then re-added, it won't show up?
I haven't verified that, but would double check in case it's a slight AI oversight.

I'm not really sure why would we need to switch branched while developing grafana (and need the dashboards to stay according to the previous branch), but any behavior that is useful from dev pov is good.

Michal-Leszczynski (commented):
In general looks good:)
Just one ask - could we make the grafana/node-exported setup in dev env opt-in?
They don't take much resources, but they are also useless in gh actions CI which runs on rather limited resources.

testing/grafana/provisioning/dashboards/dashboards.yaml
karol-kokoszka: You're right - and Karol told me to say so out loud.

Verified on Grafana 11.6.1: with deletion enabled the dashboard does go away with the file, and comes straight back when the file returns. The comment was wrong, so I dropped `disableDeletion` and the justification with it.

karol-kokoszka:
Good call, done. Grafana and node-exporter now sit behind the `monitoring` compose profile and are started by `make start-dev-env-monitoring`; `make start-dev-env` and CI bring up exactly what they did before.

The node-exporter scrape target stays in the Prometheus config either way - Prometheus just reports it down until the container is started.

